PR #1, opened 2023-11-14
lando: create initial django project with basic apps (bug 1864704)

zzzeid:
- add compose.yaml with db, proxy, and app services
- add Dockerfile for nginx and lando
- add nginx config
- add self-signed certificate generation + config
- update settings.py to support basic local setup
- add format command


nginx/nginx.conf
```diff
@@ -0,0 +1,33 @@
+upstream lando {
```
cgsheeh, line 1: This nginx config is just for running the prototype in Docker, right?

cgsheeh (changes requested):
Why is there a `src/` folder that only contains `lando/`? In Python it's standard to just name the source repo after the project at the top level.

zzzeid:
> Why is there a `src/` folder that only contains `lando/`? In Python it's standard to just name the source repo after the project at the top level.

This is a pretty common approach. See, for example: https://github.com/psf/requests, https://github.com/pallets/flask, https://github.com/mozilla-conduit/mots, and numerous others.

nginx/nginx.conf
zzzeid: Yes that is correct, it's used in `compose.yaml`.

cgsheeh:
> > Why is there a `src/` folder that only contains `lando/`? In Python it's standard to just name the source repo after the project at the top level.
> 
> This is a pretty common approach. See, for example: https://github.com/psf/requests, https://github.com/pallets/flask, https://github.com/mozilla-conduit/mots, and numerous others.
> 
> NOTE: this is a preliminary project structure, in prep for more robust packaging which will be implemented later on.

It's a useless directory, regardless of who else has made one. :sweat_smile:

cgsheeh (approved):

zzzeid:
> > > Why is there a `src/` folder that only contains `lando/`? In Python it's standard to just name the source repo after the project at the top level.
> > 
> > 
> > This is a pretty common approach. See, for example: https://github.com/psf/requests, https://github.com/pallets/flask, https://github.com/mozilla-conduit/mots, and numerous others.
> > NOTE: this is a preliminary project structure, in prep for more robust packaging which will be implemented later on.
> 
> It's a useless directory, regardless of who else has made one. 😅

https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

cgsheeh:
> > > > Why is there a `src/` folder that only contains `lando/`? In Python it's standard to just name the source repo after the project at the top level.
> > > 
> > > 
> > > This is a pretty common approach. See, for example: https://github.com/psf/requests, https://github.com/pallets/flask, https://github.com/mozilla-conduit/mots, and numerous others.
> > > NOTE: this is a preliminary project structure, in prep for more robust packaging which will be implemented later on.
> > 
> > 
> > It's a useless directory, regardless of who else has made one. 😅
> 
> https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

Which of these differences is relevant to us?

zzzeid:
> > > It's a useless directory, regardless of who else has made one. 😅
> > 
> > 
> > https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
> 
> Which of these differences is relevant to us?

The link was in response to both "It's a useless directory" and "In Python it's standard to just name the source repo after the project at the top level." Both approaches are valid, and both have differences. Therefore the directory is not useless, and the other approach is not standard.

The main differences (mainly related to clear separation between "code being packaged" and "other code") won't be apparent until this project is a little more mature.

PR #2, opened 2023-11-14
settings: fetch variables from environment (bug 1864721)

zzzeid:
- create simple method that fetches settings from the environment
- fetch settings that may change based on environment in this way


src/lando/lando/settings.py
```diff
@@ -10,24 +10,41 @@
...
 """
 
+
+import os
 from pathlib import Path
 
-# Build paths inside the project like this: BASE_DIR / 'subdir'.
-BASE_DIR = Path(__file__).resolve().parent.parent
 
+def get_setting(
```
cgsheeh, line 18: We should support type hints for these settings in some way. We could create methods with concrete return types like `get_string_setting` etc, or we could add type hints to each setting definition, like `SECRET_KEY: str = get_setting` etc below. Using the `target_type` parameter is fine if we go with type hints on the definition, but without adding hints the LSP can't infer the type.

src/lando/lando/settings.py
```diff
@@ -10,24 +10,41 @@
...
+    key: str, default: any, source: any = os.environ, target_type=None
+) -> any:
+    """Fetch a setting from a source defined by `source`.
 
-# Quick-start development settings - unsuitable for production
-# See https://docs.djangoproject.com/en/dev/howto/deployment/checklist/
+    `source` should have a `get` method defined, which will be called when fetching the
```
cgsheeh, line 23: It would be better if we passed the `get` method as the source instead (ie `os.environ.get`) and then the type hint will be `Callable` instead of `Any`.

src/lando/lando/settings.py
```diff
@@ -10,24 +10,41 @@
...
+
+import os
 from pathlib import Path
 
-# Build paths inside the project like this: BASE_DIR / 'subdir'.
-BASE_DIR = Path(__file__).resolve().parent.parent
 
+def get_setting(
+    key: str, default: any, source: any = os.environ, target_type=None
+) -> any:
```
cgsheeh, line 20: We need to use `typing.Any` here, using the `any` builtin function isn't supported.

cgsheeh (changes requested):

src/lando/lando/settings.py
zzzeid: Out of the two things you suggested, I think the latter is better. This particular implementation to get settings is very preliminary, mainly to support being able to not track secrets which I am using to connect to Auth0. I think we will have to create something more robust as a follow up regardless.

src/lando/lando/settings.py
zzzeid: I don't think we should be changing the implementation of something solely to satisfy type hinting -- rather, the other way around would make sense.

I think that passing `os.environ.get` (or, in the case when `source` is say, a dictionary, passing `some_dict.get`) is awkward. In this particular case, we could use `Mapping` as a type hint to improve this type hint.

src/lando/lando/settings.py
zzzeid: After fixing up the `.env` in `compose.yaml`, this method is actually no longer needed since Docker can pick up the environment variables from the `.env` file directly. Will remove this, but we may have to revisit this in the future.

src/lando/lando/settings.py
cgsheeh: My point was that passing `Any` when the actual constraint is "an object with a `.get`" isn't very helpful, and we should constrain it within the type system where possible. Using a `Mapping` type hint would have worked as well, but I like the simpler solution of `os.getenv`.

cgsheeh (approved):

src/lando/lando/settings.py
zzzeid: Yep agreed -- this should have been a `Mapping` type from the get go and `Any` was indeed too broad.

PR #5, opened 2023-11-22
main: add basic support for repo + worker (bug 1865907)

zzzeid:
- add Repo and Worker models
- add basic support commands
- add method to initialize repo in filesystem


src/lando/main/models.py
```diff
@@ -179,3 +182,53 @@ def add_job_with_revisions(revisions: list[Revision], **params: Any) -> LandingJ
     for revision in revisions:
         job.revisions.add(revision)
     return job
+
+
+class Repo(BaseModel):
```
cgsheeh, line 187: This needs to work with hg, adding Git support is out of scope here.

cgsheeh (changes requested):

src/lando/main/models.py
zzzeid: The proof of concept is for git, and this whole class will be significantly modified once the landing worker POC + the full scaffolding is in place. It's a lot simpler to implement the POC for git, vs. mercurial.

cgsheeh: Considering we already have a full implementation for a Mercurial based landing worker, I'm not sure why that would be the case.

cgsheeh (approved):

src/lando/main/models.py
zzzeid: It's simpler to create a simple implementation for git, than to port the entire mercurial implementation (i.e., everything in `hg.py` and supporting functionality). It was a judgment call to both speed up the POC development and to test drive git landings.

PR #8, opened 2023-11-27
main: add basic support for repo + worker (bug 1865907)

zzzeid:
- add Repo and Worker models
- add basic support commands
- add method to initialize repo in filesystem


zzzeid:
Same as #5

PR #17, opened 2023-12-13
main: update models and migration files (bug 1869032)

zzzeid:
- translate remaining models to Django models
- generate new migration file
- port various methods and properties to use new models
- store patch as string rather than bytes


zzzeid:
Bit of feature creep, but I know that we've previously discussed changing the way we store patches, and this is quite a simple change that I thought we could include here.

src/lando/main/models/configuration.py
```diff
@@ -99,7 +104,10 @@ def set(
         NOTE: This method will create the variable with the provided `key` if it does
         not exist.
         """
-        record = cls.query.filter(cls.key == key.value).one_or_none()
+        try:
+            record = cls.objects.get(cls.key == key.value)
+        except record.DoesNotExist:
```
cgsheeh, line 109: Should this be `cls.DoesNotExist`?

cgsheeh (changes requested):

cgsheeh:
> Bit of feature creep, but I know that we've previously discussed changing the way we store patches, and this is quite a simple change that I thought we could include here.

Yes, I was going to make a comment about that but I came to the same conclusion as you. With the `patch_bytes` shim I think we should be covered on that front.

I know this is probably on your mind as well but the sooner we can get the existing LandoAPI tests running in this prototype the more confidence we will have about making such changes. :)

cgsheeh (approved):

PR #20, opened 2024-01-04
settings: add jinja2 support (bug 1873071)

zzzeid:
- add jinja2 to requirements
- update settings to pick up jinja2 templates
- add basic jinja2 environment to lando

### DIFF
--- a/src/lando/jinja2.py
+++ b/src/lando/jinja2.py
@@ -1,0 +1,15 @@
+from django.templatetags.static import static
+from django.urls import reverse
+
+from jinja2 import Environment
+
+
+def environment(**options):
+    env = Environment(**options)
+    env.globals.update(
+        {
+            "static": static,
+            "url": reverse,
+        }
+    )
+    return env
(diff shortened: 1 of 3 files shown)


src/lando/jinja2.py
```diff
@@ -0,0 +1,15 @@
+from django.templatetags.static import static
```
cgsheeh, line 1: What else is going to go in this module? Could we just put this in `settings.py`?

cgsheeh (changes requested):
Seems out of place to add this without including any templates.

zzzeid:
> Seems out of place to add this without including any templates.

Templates are already in the base branch, and are moved in PR #21, and will be properly ported in a different PR, since they all depend on various functionality that is not available at this time.

src/lando/jinja2.py
zzzeid: > What else is going to go in this module? Could we just put this in `settings.py`?

This is the typical way of adding a Jinja environment. Basically all the filters and custom functionality we added in the legacy repo will eventually go in here. This doesn't belong in the settings module.

cgsheeh (approved):

PR #42, opened 2024-02-05
legacy: remove sec-approval api/model code (bug 1878085)

zzzeid:


cgsheeh (approved):
LGTM. Would be nice to have this in the LandoAPI repo also so we can run tests and verify it doesn't break anything.

zzzeid:
> LGTM. Would be nice to have this in the LandoAPI repo also so we can run tests and verify it doesn't break anything.

Good idea. See https://github.com/mozilla-conduit/lando-api/pull/377. Tests seem to be passing, but we should also test on DEV at some point before deploying.

PR #43, opened 2024-02-09
jinja2: general enhancements

zzzeid:
- move filters and globals to the correct variables (bug 1879586)
- add get_messages shortcut to be used in templates (bug 1875914)


src/lando/jinja2.py
```diff
@@ -323,15 +336,7 @@ def environment(**options):
             "graph_height": graph_height,
             "graph_width": graph_width,
             "graph_x_pos": graph_x_pos,
-            "is_user_authenticated": is_user_authenticated(env),
-            "linkify_bug_numbers": linkify_bug_numbers,
-            "linkify_faq": linkify_faq,
-            "linkify_revision_ids": linkify_revision_ids,
-            "linkify_revision_urls": linkify_revision_urls,
-            "linkify_sec_bug_docs": linkify_sec_bug_docs,
-            "linkify_transplant_details": linkify_transplant_details,
```
cgsheeh, line 332: Why were these removed? Should they have been moved as well?

cgsheeh (changes requested):

src/lando/jinja2.py
zzzeid: `is_user_authenticated` was indeed moved, though in the future it will also become obsolete since we will use Django's auth functionality.

I didn't mean to remove the `linkify_` filters in this PR, will put them back!

cgsheeh (changes requested):

src/lando/jinja2.py
```diff
@@ -312,6 +314,17 @@ def message_type_to_notification_class(flash_message_category: str) -> str:
...
     env = Environment(**options)
     env.globals.update(
+        {
+            "url": reverse,
+            "config": settings,
+            "is_user_authenticated": is_user_authenticated(env),
+            "new_settings_form": UserSettingsForm,
+            "get_messages": messages.get_messages,
+            "user_has_phabricator_token": user_has_phabricator_token(env),
+
```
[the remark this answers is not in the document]
zzzeid: Heh, fixed. `lando format` will catch these in the future, but running it now will capture too many things at the same time.

cgsheeh (approved):

PR #49, opened 2024-04-29
tests: more fixes to tests (bug 1887042)

zzzeid:
No implementation changes in this PR, however there is the addition of proxy_client, which will likely be used by any future commits from the legacy repos that contain changes to API tests, until that functionality is fully ported.

Tests that have been skipped will be readded in a future PR with fixes for those tests. At this point, CI should be passing and future PRs should continue this trend.

- add proxy client to get tests to pass with minimal changes
- reorder some test arguments (e.g., project fixtures)
- make temporary changes in conftest until test env vars are fixed
- remove suppress failure (cache) in auth module (bug 1870285)
- fix queryset/db issues (leftover from bug 1886854)
- add a couple more proxy endpoints to proxy client
- update support shims to better accommodate tests
- update test cache settings
- change docker workdir
- use correct django settings for tests (lando.test_settings)
- fix hardcoded hgext directory

### DIFF
--- a/src/lando/api/legacy/api/landing_jobs.py
+++ b/src/lando/api/legacy/api/landing_jobs.py
@@ -29,7 +29,6 @@ def put(landing_job_id: str, data: dict):
             updated (for example, when trying to cancel a job that is already in
             progress).
     """
-    # TODO: fix this based on locks/etc.
     with LandingJob.lock_table:
         landing_job = LandingJob.objects.get(pk=landing_job_id)
 
--- a/src/lando/api/legacy/api/landing_jobs.py
+++ b/src/lando/api/legacy/api/landing_jobs.py
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import logging
 
-from lando.api import auth
+from lando.api.legacy import auth
 from lando.main.models.landing_job import LandingJob, LandingJobAction, LandingJobStatus
 from lando.main.support import ProblemException, g
 
@@ -30,7 +30,8 @@ def put(landing_job_id: str, data: dict):
             progress).
     """
     # TODO: fix this based on locks/etc.
-    landing_job = LandingJob.query.with_for_update().get(landing_job_id)
+    with LandingJob.lock_table:
+        landing_job = LandingJob.objects.get(pk=landing_job_id)
 
     if not landing_job:
         raise ProblemException(
@@ -49,7 +50,8 @@ def put(landing_job_id: str, data: dict):
             type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403",
         )
 
-    if data["status"] != LandingJobStatus.CANCELLED.value:
+    # TODO: fix this. See bug 1893455.
+    if data["status"] != "CANCELLED":
         raise ProblemException(
             400,
             "Invalid status provided",
--- a/src/lando/api/tests/conftest.py
+++ b/src/lando/api/tests/conftest.py
@@ -600,7 +600,7 @@ def proxy_client(monkeypatch):
                 return self._handle__post__transplants(path, **kwargs)
 
         def put(self, path, **kwargs):
-            """Handle put endpoints."""
+            """Handle put endpoint."""
             if "headers" in kwargs:
                 mock_request = {"headers": kwargs["headers"]}
                 monkeypatch.setattr("lando.api.legacy.auth.request", mock_request)
--- a/src/lando/api/tests/conftest.py
+++ b/src/lando/api/tests/conftest.py
@@ -515,7 +515,7 @@ def proxy_client(monkeypatch):
     reimplemented to not need a response or response-like object.
     """
 
-    class Response:
+    class MockResponse:
         """Mock response class to satisfy some requirements of tests."""
 
         # NOTE: The methods tested that rely on this class should be reimplemented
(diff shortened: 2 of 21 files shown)


src/lando/api/legacy/api/landing_jobs.py
```diff
@@ -30,7 +30,8 @@ def put(landing_job_id: str, data: dict):
             progress).
     """
     # TODO: fix this based on locks/etc.
```
zzzeid, line 32:
```suggestion
```

src/lando/api/tests/conftest.py
```diff
@@ -471,3 +500,101 @@ def strptime(cls, date_string, fmt):
...
+        def __init__(self, status_code=200, json=None):
+            self.json = json or {}
+            self.status_code = status_code
+            self.content_type = (
+                "application/json"
+                if str(status_code)[0] not in ("4", "5")
+                else "application/problem+json"
+            )
+
+    class _proxy:
```
cgsheeh, line 532: `class TestProxy` would be a nicer name here. If you added a constructor you could pass `monkeypatch` as an argument, and then this could be defined outside the test fixture as well.

src/lando/api/tests/conftest.py
```diff
@@ -471,3 +500,101 @@ def strptime(cls, date_string, fmt):
...
+        def get(self, path, *args, **kwargs):
+            """Handle various get endpoints."""
+            if path.startswith("/stacks/D"):
```
cgsheeh, line 535: We should add handler functions for each of these cases, like:

```python
if path.startswith("/stacks/D"):
    return self.handle_get_stacks(path, *args, **kwargs)
```

etc. It's a bit tough to follow with how many levels of indentation are in some of these branches.

src/lando/main/models/base.py
```diff
@@ -18,28 +19,38 @@
...
-                raise ValueError(f"{lock} not valid.")
-
-        def __enter__(self):
-            cursor = connection.cursor()
-            cursor.execute(f"LOCK TABLE {self._meta.db_table} IN {self.lock} MODE")
-
-        def __exit__(self, exc_type, exc_value, traceback):
-            pass
+    @classmethod
+    @property
```
cgsheeh, line 51: Nit: we can remove `*args, **kwargs` if this is a property.

src/lando/main/models/landing_job.py
```diff
@@ -182,8 +182,13 @@ def revisions_query(cls, revisions: Iterable[str]) -> QuerySet:
         that stores revisions and diff IDs. Those records are now deprecated and will
         not be included in this query.
         """
+        from django.db.models import Q
```
cgsheeh, line 185: Move this to a top-level import.

cgsheeh (changes requested):
A few higher level comments:

- There are still code formatting changes showing up in this PR. Please re-introduce the code quality tests for black and ruff now that we have passing CI.
- This PR contains quite a few changes which could be separated into their own PRs (for example, the cache suppression changes, the `lock_table` changes, etc). Please keep each PR as one unit of work.

zzzeid:
>     * There are still code formatting changes showing up in this PR. Please re-introduce the code quality tests for black and ruff now that we have passing CI.

I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1897434) to track this.

>     * This PR contains quite a few changes which could be separated into their own PRs (for example, the cache suppression changes, the `lock_table` changes, etc). Please keep each PR as one unit of work.

Those changes surfaced as part of fixing tests, and as such I think they can stay in this PR (this PR is a general one that affects many modules). However, I can reference the bugs that missed those particular changes in the commit message. I don't think we need separate PRs or bugs for them.

src/lando/api/tests/conftest.py
```diff
@@ -471,3 +500,101 @@ def strptime(cls, date_string, fmt):
...
+    Most tests that use the API no longer need to access those endpoints through
+    the API as the data can be fetched directly within the application. This client
+    is a temporary implementation to bridge tests and minimize the number of changes
+    needed to the tests during the porting process.
+
+    This client should be removed and all the tests that depend on it should be
+    reimplemented to not need a response or response-like object.
+    """
+
+    class Response:
```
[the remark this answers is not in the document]
zzzeid: Renamed to `MockResponse` to be more inline with the description in the docstring.

src/lando/api/tests/conftest.py
cgsheeh: It's mostly just a desire to avoid nesting things where they could be defined as top-level, making it easier to read, and easier to load into a REPL if we had to debug some problem by hand. I'm not too fussed about it after the addition of the handler methods.

src/lando/api/tests/test_try.py
```diff
@@ -15,6 +15,9 @@
 from lando.api.legacy.workers.landing_worker import LandingWorker
 from lando.main.models.landing_job import LandingJob, LandingJobStatus
 
+
+pytest.skip(allow_module_level=True)
```
cgsheeh, line 19: Nit: should probably add a TODO here so this doesn't fall off the radar.

cgsheeh (approved):

src/lando/api/tests/conftest.py
```diff
@@ -471,3 +500,112 @@ def strptime(cls, date_string, fmt):
...
+                monkeypatch.setattr("lando.api.legacy.auth.request", mock_request)
+
+            if path.startswith("/transplants/dryrun"):
+                return self._handle__post__transplants__dryrun(**kwargs)
+
+            if path == "/transplants":
+                return self._handle__post__transplants(path, **kwargs)
+
+        def put(self, path, **kwargs):
+            """Handle put endpoints."""
```
zzzeid, line 603:
```suggestion
            """Handle put endpoint."""
```

src/lando/api/tests/test_try.py
zzzeid: There's a few of these right now, I'll file a bug to make sure we unskip all of them.

PR #41, opened 2024-01-26
ui: revisions page and supporting functionality (git version)

zzzeid:
WIP DO NOT MERGE

PR #55, opened 2024-05-07
landing_job: Rename 'revisions' field to 'unsorted_revisions' and re-use the 'revisions' field for sorted revisions (bug 1893453)

ahochheiden:


src/lando/main/models/landing_job.py
```diff
@@ -246,6 +246,9 @@ def sort_revisions(self, revisions: list[Revision]):
                 revision=revision, landing_job=self
             ).update(index=index)
 
+    def sorted_revisions(self):
```
zzzeid, line 249: Either `get_sorted_revision` or changing this into a property would be better here.

zzzeid (changes requested):
This looks great! I am wondering though since there is also additional code that is in the testing branch (#49) and the revisions branch (#41) that also uses this property, if it would make more sense in terms of backwards compatibility to:
- rename the `revisions` field to be `unordered_revisions`
- add a `revisions` property that returns a QuerySet of the ordered revisions

src/lando/main/models/landing_job.py
```diff
@@ -150,20 +150,20 @@ def landing_job_identifier(self) -> str:
...
         if head.revision_id:
             # Return the Phabricator identifier if the head revision has one.
             return f"D{head.revision_id}"
 
         # If there is no Phabricator identifier, return the first line of the
         # non-try-syntax commit's message for the patch.
-        if self.revisions.count() > 1:
-            head = self.revisions.order_by("-id")[1]
+        if self.unsorted_revisions.count() > 1:
+            head = self.unsorted_revisions.order_by("-id")[1]
```
zzzeid, line 166: Same comment applies here as above.

zzzeid (approved):

PR #56, opened 2024-05-07
main: Add explicit sort ordering for 'LandingJobStatus' (bug 1893455)

ahochheiden:


src/lando/main/models/landing_job.py
```diff
@@ -217,7 +218,18 @@ def job_queue_query(
...
+            When(status=LandingJobStatus.IN_PROGRESS, then=2),
+            When(status=LandingJobStatus.DEFERRED, then=3),
+            When(status=LandingJobStatus.FAILED, then=4),
+            When(status=LandingJobStatus.LANDED, then=5),
+            When(status=LandingJobStatus.CANCELLED, then=6),
+            default=0,
+            output_field=IntegerField()
+        )
+
+        q = q.annotate(custom_order=ordering).order_by("-custom_order", "-priority", "created_at")
```
zzzeid, line 232: nit
```suggestion
        q = q.annotate(status_order=ordering).order_by("-status_order", "-priority", "created_at")
```

zzzeid (changes requested):
Whenever a model is changed, `lando makemigrations` should be run to generate a migration file. Meant to add that as a make command as well.

ahochheiden:
> Whenever a model is changed, `lando makemigrations` should be run to generate a migration file. Meant to add that as a make command as well.

Done.

src/lando/main/models/landing_job.py
ahochheiden: Works for me.

zzzeid (approved):
lgtm!

one nit: if you look at previous commit message history, the first line is usually kept pretty short (on my machine, under 72 characters), and it matches the PR title before merge. You can keep it brief and put any additional detail in the rest of the message. On my machine, vim automatically breaks the line at that point. I think we have some exceptions here and there, and it's more of a convention/preference so I'll leave it up to you. We should probably have a proper styleguide for commits somewhere.

Looks like the average first line length is about 65 characters.

PR #57, opened 2024-05-13
version: Use `setuptools_scm` to generate the version (bug 1867133)

ahochheiden:
`setuptools_scm` is meant to generate the `_version.py` file when building a wheel for a project, but we can use the same code to explicitly generate it when we 'build' using the Dockerfile.

The 'version' text generated is the `git tag` with the commit hash appended. The version file is not committed to VCS.

### DIFF
--- a/generate_version_file.py
+++ b/generate_version_file.py
@@ -1,0 +1,3 @@
+from setuptools_scm import get_version
+
+get_version(root="./code/", relative_to="__file__", write_to="src/lando/version.py")
--- a/generate_version_file.py
+++ b/generate_version_file.py
@@ -1,3 +1,0 @@
-from setuptools_scm import get_version
-
-get_version(root="./code/", relative_to="__file__", write_to="src/lando/version.py")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,17 +18,21 @@ dependencies = [
 ]
 name = "lando"
 requires-python = ">=3.10"
-version = "0.1.0"
+dynamic = ["version"]
 
 [project.scripts]
 lando = "lando.manage:main"
 
 [build-system]
 requires = [
-    "setuptools>=45",
+    "setuptools>=64",
+    "setuptools_scm>=8",
     "wheel",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
+version_file = "src/lando/version.py"
 
 [tool.setuptools.packages.find]
 where = ["src"]
@@ -37,4 +41,5 @@ where = ["src"]
 DJANGO_SETTINGS_MODULE = "lando.test_settings"
 testpaths = [
     "src/lando/api",
+	"src/lando/tests",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,5 +41,5 @@ where = ["src"]
 DJANGO_SETTINGS_MODULE = "lando.test_settings"
 testpaths = [
     "src/lando/api",
-	"src/lando/tests",
+    "src/lando/tests",
 ]
--- a/src/lando/tests/test_version.py
+++ b/src/lando/tests/test_version.py
@@ -1,9 +1,7 @@
 from django.core.management import call_command
-
+from lando.version import version
 
 def test_version():
-    from lando.version import version
-
     # Explicitly generate the version.py file since we can't
     # guarantee it already exists in the testing environment.
     call_command('generate_version_file')
--- a/src/lando/tests/test_version.py
+++ b/src/lando/tests/test_version.py
@@ -1,0 +1,21 @@
+import pytest
+import subprocess
+
+
+def test_version():
+    # explicitly generate the version.py file since we can't
+    # guarantee it already exists in the testing environment
+    subprocess.call(["lando", "generate_version_file"])
+
+    try:
+        from lando.version import version
+
+        # every commit will change the exact version string, so it
+        # doesn't make sense to compare it against a known value.
+        # testing that it exists, is a string, and is at least 5
+        # characters long (eg: "1.1.1") should suffice.
+        assert version is not None, "'version' should not be None"
+        assert isinstance(version, str), "'version' should be a string"
+        assert len(version) >= 5, "'version' string should be at least 5 characters long"
+    except ImportError as e:
+        pytest.fail(f"ImportError occurred: {e}")
--- a/src/lando/tests/test_version.py
+++ b/src/lando/tests/test_version.py
@@ -1,10 +1,25 @@
+from django.conf import settings
 from django.core.management import call_command
-from lando.version import version
+
+import pytest
 
 def test_version():
-    # Explicitly generate the version.py file since we can't
-    # guarantee it already exists in the testing environment.
+    # The version file may or may not exist in the testing environment.
+    # We'll explicitly remove it so that we're in a known state, then
+    # re-generate it and test against it.
+    version_file = settings.BASE_DIR / "version.py"
+    version_file.unlink(missing_ok=True)
+
+    # We should not be able to import it after it's been removed.
+    with pytest.raises(ImportError):
+        from lando.version import version
+
     call_command('generate_version_file')
+
+    try:
+        from lando.version import version
+    except ImportError:
+        pytest.fail("ImportError: Unable to import the version file after re-generation.")
 
     # Every commit will change the exact version string, so it
     # doesn't make sense to compare it against a known value.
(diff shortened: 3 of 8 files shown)


generate_version_file.py
```diff
@@ -0,0 +1,3 @@
+from setuptools_scm import get_version
+
+get_version(root="./code/", relative_to="__file__", write_to="src/lando/version.py")
```
cgsheeh, line 3: Can we use `settings.BASE_DIR` to derive these values somehow?

pyproject.toml
```diff
@@ -18,23 +18,28 @@ dependencies = [
...
 [tool.setuptools.packages.find]
 where = ["src"]
 
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "lando.test_settings"
 testpaths = [
     "src/lando/api",
-]
+	"src/lando/tests",
+]
```
cgsheeh, line 45: Newline missing here.

src/lando/tests/test_version.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+import subprocess
+
+
+def test_version():
+    # explicitly generate the version.py file since we can't
+    # guarantee it already exists in the testing environment
```
cgsheeh, line 7: Please use formal English in your comments (ie proper punctuation, capitalization, etc).

```suggestion
    # Explicitly generate the version.py file since we can't
    # guarantee it already exists in the testing environment.
```

src/lando/tests/test_version.py
```diff
@@ -0,0 +1,21 @@
...
+    subprocess.call(["lando", "generate_version_file"])
+
+    try:
+        from lando.version import version
+
+        # every commit will change the exact version string, so it
+        # doesn't make sense to compare it against a known value.
+        # testing that it exists, is a string, and is at least 5
+        # characters long (eg: "1.1.1") should suffice.
```
cgsheeh, line 16:
```suggestion
        # Every commit will change the exact version string, so it
        # doesn't make sense to compare it against a known value.
        # testing that it exists, is a string, and is at least 5
        # characters long (eg: "1.1.1") should suffice.
```

src/lando/tests/test_version.py
```diff
@@ -0,0 +1,21 @@
...
+        assert version is not None, "'version' should not be None"
+        assert isinstance(version, str), "'version' should be a string"
+        assert len(version) >= 5, "'version' string should be at least 5 characters long"
+    except ImportError as e:
+        pytest.fail(f"ImportError occurred: {e}")
```
cgsheeh, line 21: The `try/except` should only wrap the import.

pyproject.toml
```diff
@@ -18,23 +18,28 @@ dependencies = [
...
+
 [tool.setuptools.packages.find]
 where = ["src"]
 
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "lando.test_settings"
 testpaths = [
     "src/lando/api",
-]
+	"src/lando/tests",
```
cgsheeh, line 44: Indentation looks funky, looks like you need to run this through `black`.

cgsheeh (changes requested):

pyproject.toml
zzzeid: There's a `lando format` command by the way. There are some inconsistencies in terms of formatting across the repo, mainly due to the two lando legacy repos having different formatting. I filed a separate bug to track the work of fixing that (https://bugzilla.mozilla.org/show_bug.cgi?id=1897434), until then it's OK to manually run `ruff` and `black` (within the container) on files that are changed.

generate_version_file.py
zzzeid, line 1: @ahochheiden I think all the functionality in this file should move to the management command itself. Then you will have access to `settings.BASE_DIR` as @cgsheeh mentioned. I.e., no need to have this be called by subprocess, you should be able to call it directly from there.

src/lando/tests/test_version.py
zzzeid, line 21: I think it is OK to not wrap this entire thing with a try/except. After all, we are not expecting this test to raise an exception. If it does, it will fail anyway with an error. I.e., I would suggest getting rid of this try/except block altogether.

src/lando/tests/test_version.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+import subprocess
+
+
+def test_version():
+    # explicitly generate the version.py file since we can't
+    # guarantee it already exists in the testing environment
+    subprocess.call(["lando", "generate_version_file"])
```
zzzeid, line 8: If you define the functionality that generates and stores the file in a method in `lando.utils.management.commands.generate_version_file`, then you can simply import the method and call it here instead.

zzzeid (changes requested):

pyproject.toml
ahochheiden: It's because it's a tab, I think I opened this file in notepad to add the line (instead of my usual IDE) and it didn't replace the tab with four spaces.

I did actually run `lando format`, and just tried again, and it doesn't replace the tab with four spaces. I've manually corrected it though.

zzzeid: I guess we both missed that this is not a Python file (and also, is in the root directory of the project.)

zzzeid: Looks like `ruff` would have caught this, but it is not defined in LINT_PATHS. Can be added there.

generate_version_file.py
ahochheiden: After moving the logic to be inside the management command, the `root` no longer needs to be specified at all.

src/lando/tests/test_version.py
```diff
@@ -0,0 +1,17 @@
+from django.core.management import call_command
+
+
+def test_version():
+    from lando.version import version
```
cgsheeh, line 5: Make this a top-level import.

src/lando/utils/management/commands/generate_version_file.py
```diff
@@ -0,0 +1,10 @@
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = "Explicitly generate the 'version.py' file."
+
+    def handle(self, *args, **options):
+        from setuptools_scm import get_version
```
cgsheeh, line 8: Top-level import here.

cgsheeh (changes requested):

src/lando/tests/test_version.py
ahochheiden: My thinking was that the import could fail, so it's part of the test, but I guess it doesn't matter since it's unlikely there will ever be more than one test in this file.

zzzeid: I think this is a special case, because the module that we are importing is generated by the command that is being called. I think it makes sense to try the import *after* the command is run. Unless the file is there before the command is run, in which case there's something slightly off here.

src/lando/utils/management/commands/generate_version_file.py
```diff
@@ -0,0 +1,8 @@
+from django.core.management.base import BaseCommand
+from setuptools_scm import get_version
+
+class Command(BaseCommand):
+    help = "Explicitly generate the 'version.py' file."
+
+    def handle(self, *args, **options):
+        get_version(write_to="src/lando/version.py")
```
zzzeid, line 8: nit: I think it would be slightly better to do `get_version(write_to=settings.BASE_DIR / "version.py")`. That way fewer things are hardcoded.

src/lando/tests/test_version.py
zzzeid: So if this test is passing with the import at the top, it means that the command isn't being tested properly. The test should probably delete the version file, and then run the command and do the import.

src/lando/utils/management/commands/generate_version_file.py
ahochheiden: I don't think this works since it's expecting a relative path. The error is:

```
        for step, path in enumerate([other] + list(other.parents)):
            if self.is_relative_to(path):
                break
            elif not walk_up:
>               raise ValueError(f"{str(self)!r} is not in the subpath of {str(other)!r}")
E               ValueError: '/code/src/lando/version.py' is not in the subpath of '.'
```
Since `.` is `code` already, it can't resolve `/code/`.

```
get_version(write_to=Path(*settings.BASE_DIR.parts[2:]) / "version.py")
```
Would work, but it's less clear than what it is now, so I think leaving it as-is is best.

src/lando/tests/test_version.py
ahochheiden: I've expanded the `version_test` to cover deleting the file and testing the import explicitly.

zzzeid (approved):
lgtm!

PR #58, opened 2024-05-16
dockerflow: Add dockerflow module (bug 1873243) 

ahochheiden:
- I used `decorators` instead of `middleware` since I believe the `middleware` would have applied to everything, but I only wanted the `log_request` and `disable_caching` code to run for `dockerflow`.
- I added equivalent support for the logging we had with Flask, but it's not hooked up for Django yet. We still need to add that in `settings.py` with handlers and formatters.
- The `heartbeat` health 'check' could probably be more thorough, but I'm unsure what should be checked.
- A test fixture (`generate_version_file`) has been added for ensuring that the version file exists in the testing environment, mainly to reduce code duplication.


src/lando/urls.py
```diff
@@ -18,9 +18,16 @@
...
     path("admin/", admin.site.urls),
     path("oidc/", include("mozilla_django_oidc.urls")),
     path("D<int:revision_id>/", revisions.Revision.as_view()),
 ]
+
+urlpatterns += [
+    path("__version__", DockerflowViews.VersionView.as_view(), name="version"),
+    path("__heartbeat__", DockerflowViews.HeartbeatView.as_view(), name="heartbeat"),
+    path("__lbheartbeat__", DockerflowViews.LoadBalancerHeartbeatView.as_view(), name="lbheartbeat"),
+]
```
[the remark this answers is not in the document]
zzzeid: I think it is better to separate them as they are here. In fact it is not uncommon to have each list be named after the app itself and then to add up all the lists at the end, or for some of these URLs to even be defined in a different module and imported/added here. I would suggest that we organize the original list a little better though.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,66 @@
...
+
+class HeartbeatView(View):
+    @disable_caching
+    @log_request
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
+
+    def get(self, request):
+        from django.db import connections
+        from django.db.utils import OperationalError
```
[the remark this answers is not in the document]
ahochheiden: Done.

src/lando/urls.py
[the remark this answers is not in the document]
ahochheiden: Yeah, I'll leave this as-is for now. We can definitely improve the organization even more in the future.

src/lando/dockerflow/tests/test_dockerflow.py
```diff
@@ -0,0 +1,36 @@
...
+    assert response.status_code == 200
+    assert response['Content-Type'] == "application/json"
+
+
+def test_dockerflow_version_matches_disk_contents(client):
+    # The version file may or may not exist in the testing
+    # environment yet, so we should explicitly generate it.
+    call_command('generate_version_file')
+
+    from lando.version import version
```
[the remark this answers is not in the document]
ahochheiden: I've added the test fixture.

src/lando/dockerflow/decorators.py
```diff
@@ -0,0 +1,45 @@
+import time
+import logging
```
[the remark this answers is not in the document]
ahochheiden: I think it's because I do not add `dockerflow` to [LINT_PATHS here](https://github.com/mozilla-conduit/lando/blob/c17b6829e6851381390cf5ccaa3741320c7d4415/src/lando/settings.py#L188).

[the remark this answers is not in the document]
ahochheiden: The linter caught it after I added it.

src/lando/urls.py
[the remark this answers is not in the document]
zzzeid: In the long run it might be better to define app-specific urls in the app package itself (e.g., `lando.dockerflow.urls`) and then add them here one way or another. Probably a good thing to do in a separate bug though in the future as you said!

src/lando/dockerflow/decorators.py
```diff
@@ -0,0 +1,45 @@
+import time
+import logging
+from functools import wraps
+
+# Logging is not yet implemented
+request_logger = logging.getLogger("dockerflow.request_summary.logger")
```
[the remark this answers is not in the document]
ahochheiden: I agree, I think it's out of scope. I started playing with the formatting initially, but then scrapped when I realized that logging hasn't been ported yet. I think this is fine as-is and should be able to be hooked up when logging is implemented.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,66 @@
...
+        except ImportError:
+            data = {
+                "error": "Service Unavailable"
+            }
+            status_code = 503
+
+        return JsonResponse(data, status=status_code, json_dumps_params={'indent': 2})
+
+
+class HeartbeatView(View):
```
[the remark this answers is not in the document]
ahochheiden: Done.

ahochheiden:
The linter removed the fixture include. :upside_down_face:

zzzeid:
> The linter removed the fixture include. 🙃

In pytest, you do not need to import fixtures. You just have to pass them as arguments to the tests that need them (or use `autouse`).

ahochheiden:
> > The linter removed the fixture include. 🙃
> 
> In pytest, you do not need to import fixtures. You just have to pass them as arguments to the tests that need them (or use `autouse`).

https://github.com/mozilla-conduit/lando/actions/runs/9454564593/job/26042390729#step:6:101

I think I had what you said, but the error seems like it couldn't find it if it wasn't included. Is it supposed to be in a different file structurally so that it's visible before the particular tests run?

cgsheeh:
> > > The linter removed the fixture include. 🙃
> > 
> > 
> > In pytest, you do not need to import fixtures. You just have to pass them as arguments to the tests that need them (or use `autouse`).
> 
> https://github.com/mozilla-conduit/lando/actions/runs/9454564593/job/26042390729#step:6:101
> 
> I think I had what you said, but the error seems like it couldn't find it if it wasn't included. Is it supposed to be in a different file structurally so that it's visible before the particular tests run?

It probably needs to go in `conftest.py`.

src/lando/dockerflow/decorators.py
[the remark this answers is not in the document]
cgsheeh: Makes sense, we should get a bug on file for that if we don't have one already. :)

zzzeid:
> > > > The linter removed the fixture include. 🙃
> > > 
> > > 
> > > In pytest, you do not need to import fixtures. You just have to pass them as arguments to the tests that need them (or use `autouse`).
> > 
> > 
> > https://github.com/mozilla-conduit/lando/actions/runs/9454564593/job/26042390729#step:6:101
> > I think I had what you said, but the error seems like it couldn't find it if it wasn't included. Is it supposed to be in a different file structurally so that it's visible before the particular tests run?
> 
> It probably needs to go in `conftest.py`.

Yeah, since this is a shared fixture, I would put it in `lando.tests.conftest` and then import it in `lando.dockerflow.tests.conftest`. This should fix it.

src/lando/dockerflow/decorators.py
```diff
@@ -0,0 +1,45 @@
...
+
+    return _wrapped_view
+
+
+def disable_caching(view):
+    @wraps(view)
+    def _wrapped_view(self, request, *args, **kwargs):
+        response = view(self, request, *args, **kwargs)
+
+        if request.path.startswith("__"):
```
zzzeid, line 38: What's the point of having this if statement here? Given that this decorator would have to be explicitly added to views that we want it to affect?

src/lando/dockerflow/decorators.py
```diff
@@ -0,0 +1,45 @@
+import logging
+import time
+from functools import wraps
+
+# Logging is not yet implemented
+request_logger = logging.getLogger("dockerflow.request_summary.logger")
```
zzzeid, line 6: Curious where "dockerflow.request_summary.logger" came from? I think generally we do `logging.getLogger(__name__)`.

src/lando/dockerflow/decorators.py
```diff
@@ -0,0 +1,45 @@
+import logging
+import time
+from functools import wraps
+
+# Logging is not yet implemented
```
zzzeid, line 5: "not yet implemented" is a little inaccurate, since logging should just work out of the box. However, custom logging (i.e., in the way we intend for it to be in the long term, e.g. with mozlog, and where the log output is going, etc.) is not yet implemented.

src/lando/dockerflow/tests/test_dockerflow.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+from lando.tests.test_version import generate_version_file  # noqa: F401
```
zzzeid, line 2: I think this should be removed in lieu of adding the fixtures to `conftest`. Fixture should probably go in the project-level conftest but can be imported in `lando.dockerflow.conftest`.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    This view handles the version information of Lando.
+
+    It returns a JSON response containing the version information.
+    """
+
+    @disable_caching
+    @log_request
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
```
zzzeid, line 19: This should be defined at the top in a base class, say `DockerflowView`, and each other view should just extend that base class.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,83 @@
...
+
+            data = {
+                "version": version,
+            }
+            status_code = 200
+        except ImportError:
+            data = {"error": "Service Unavailable"}
+            status_code = 503
+
+        return JsonResponse(data, status=status_code, json_dumps_params={"indent": 2})
```
zzzeid, line 33: `json_dumps_params={"indent": 2}` is repeated in all these views. The value at least, can probably go in the parent class. Could also just define a `self._json_response(data, status)` that automatically fills in the necessary values.

E.g., 

```python
return self._json_response(data, status)
```

and elsewhere

```python
@staticmethod
def _json_response(self, data, status):
    return JsonResponse(data, status=status, json_dump_params={"indent": 2})

File: src/lando/dockerflow/views.py [hunk 13]
```diff
@@ -0,0 +1,83 @@
...
+    @disable_caching
+    @log_request
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
+
+    def get(self, request):
+        db_conn = connections["default"]
+        try:
+            db_conn.cursor()
+            healthy = True
```
zzzeid, line 53: This should go outside the try/except block.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,83 @@
...
+    It returns a JSON response containing the heartbeat information.
+    """
+
+    @disable_caching
+    @log_request
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
+
+    def get(self, request):
+        db_conn = connections["default"]
```
zzzeid, line 50: You could import [`connection`](https://docs.djangoproject.com/en/5.0/topics/db/sql/#executing-custom-sql-directly) instead, which always returns the default connection.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,83 @@
...
+            healthy = False
+
+        data = {
+            "healthy": healthy,
+            "services": {
+                "lando": healthy,
+            },
+        }
+
+        status = 200 if healthy else 502
```
zzzeid, line 64: I don't think `502` is the right status code here.

src/lando/tests/test_version.py
```diff
@@ -1,25 +1,22 @@
...
-    # re-generate it and test against it.
-    version_file = settings.BASE_DIR / "version.py"
-    version_file.unlink(missing_ok=True)
-
-    # We should not be able to import it after it's been removed.
-    with pytest.raises(ImportError):
-        from lando.version import version
 
+@pytest.fixture
+def generate_version_file():
```
cgsheeh, line 7: Can we change this to `lando_version` and then `return version` from the fixture?

cgsheeh (changes requested):
Looks like there's still a CI failure, FYI.

zzzeid (changes requested):

src/lando/dockerflow/decorators.py
```diff
@@ -0,0 +1,45 @@
...
+
+        summary = {
+            "errno": 0 if response.status_code < 400 else 1,
+            "agent": request.headers.get("User-Agent", ""),
+            "lang": request.headers.get("Accept-Language", ""),
+            "method": request.method,
+            "path": request.path,
+            "code": response.status_code,
+            "t": int(1000 * (end_time - start_time)),
+        }
```
zzzeid, line 24: Technically this is an [optional recommendation](https://github.com/mozilla-services/Dockerflow?tab=readme-ov-file#optional-recommendations), I think we could probably punt this to when logging is more robust (probably as part of #12)

src/lando/dockerflow/decorators.py
ahochheiden: It's similarly named in the original file, I was just being more verbose since it only applies to requests for Dockerflow.

https://github.com/mozilla-conduit/lando/blob/c17b6829e6851381390cf5ccaa3741320c7d4415/src/lando/ui/legacy/dockerflow.py#L10C1-L11C54

I removed the original line that follows the standard because it wasn't being used.

src/lando/dockerflow/decorators.py
ahochheiden: Leftover from the original middleware implementation. I'll fix it since it's not needed with the decorators.

src/lando/tests/test_version.py
ahochheiden: Done.

src/lando/dockerflow/decorators.py
ahochheiden: I'll remove the comment.

src/lando/dockerflow/views.py
ahochheiden: Done.

src/lando/dockerflow/views.py
ahochheiden: Done.

src/lando/dockerflow/views.py
ahochheiden: That's what's in [the original](https://github.com/mozilla-conduit/lando/blob/c17b6829e6851381390cf5ccaa3741320c7d4415/src/lando/ui/legacy/dockerflow.py#L83).

I'll switch it to 503, I think that makes more sense.

src/lando/dockerflow/views.py
ahochheiden: Alright, I've switched it to `connection.ensure_connection()`.

src/lando/dockerflow/views.py
ahochheiden: Done.

src/lando/dockerflow/decorators.py
ahochheiden: Alright, I'll leave it as is.

src/lando/dockerflow/tests/test_dockerflow.py
ahochheiden: Done.

src/lando/dockerflow/tests/conftest.py
```diff
@@ -0,0 +1 @@
+from lando.tests.conftest import lando_version  # noqa: F401
```
zzzeid, line 1: nit: another (more conventional) way to prevent the lint error is to add the import to `__all__`.
```suggestion
from lando.tests.conftest import lando_version

__all__ = ["lando_version"]
```

src/lando/dockerflow/views.py
```diff
@@ -1,56 +1,62 @@
...
+    It handles common functionality needed by all Dockerflow views.
     """
 
     @disable_caching
     @log_request
     def dispatch(self, request, *args, **kwargs):
         return super().dispatch(request, *args, **kwargs)
 
+    @staticmethod
+    def _json_response(self, data, status):
```
zzzeid, line 22:
```suggestion
    def _json_response(data, status):
```

When decorated with `staticmethod`, methods no longer receive (or need) the instance in the first argument.

src/lando/dockerflow/views.py
```diff
@@ -1,56 +1,62 @@
...
             }
-            status_code = 200
+            status = 200
         except ImportError:
             data = {"error": "Service Unavailable"}
-            status_code = 503
+            status = 503
 
-        return JsonResponse(data, status=status_code, json_dumps_params={"indent": 2})
+        return self._json_response(self, data=data, status=status)
```
zzzeid, line 45:
```suggestion
        return self._json_response(data=data, status=status)
```

src/lando/dockerflow/views.py
```diff
@@ -61,23 +67,18 @@ def get(self, request):
             },
         }
 
-        status = 200 if healthy else 502
+        status = 200 if healthy else 503
 
-        return JsonResponse(data, status=status, json_dumps_params={"indent": 2})
+        return self._json_response(self, data=data, status=status)
```
zzzeid, line 72:
```suggestion
        return self._json_response(data=data, status=status)
```

src/lando/tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+def lando_version():
+    # The version file may or may not exist in the testing environment.
+    # We'll explicitly generate it to ensure it's there.
+    call_command('generate_version_file')
+
+    try:
+        from lando.version import version
+        return version
+    except ImportError:
+        pytest.fail("ImportError: Unable to import the version file after generation.")
```
zzzeid, line 16: Ideally only the import stays in the try/except block. I also don't think we necessarily need to be calling `pytest.fail` here. An import error will cause a test to fail as an error, and I think it is self-evident in this case.

```suggestion
    from lando.version import version
    return version
```

zzzeid (changes requested):
Looking good! Just a few more comments.

src/lando/dockerflow/views.py
ahochheiden: Done.

src/lando/tests/conftest.py
ahochheiden: Very well.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,84 @@
...
+        try:
+            from lando.version import version
+
+            data = {
+                "version": version,
+            }
+            status = 200
+        except ImportError:
+            data = {"error": "Service Unavailable"}
+            status = 503
```
zzzeid, line 43: The try section of try/except blocks should contain as little code as possible (see [PEP 8](https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations)).
```python
try:                                                                                                                                                                                                                  
    from lando.version import version
except ImportError:
    data = {"error": "Service Unavailable"}
    status = 503 
else:
    data = { 
        "version": version,
    }   
    status = 200 
...
```

zzzeid (changes requested):
One last thing I must have missed in my previous review.

src/lando/tests/conftest.py
```diff
@@ -0,0 +1,13 @@
+from django.core.management import call_command
+
+import pytest
+
+
+@pytest.fixture
+def lando_version():
+    # The version file may or may not exist in the testing environment.
+    # We'll explicitly generate it to ensure it's there.
+    call_command('generate_version_file')
```
cgsheeh, line 10: Single quotes here, looks like this isn't being checked in the styling tests. I wonder if we should just run black/ruff on everything and forget about maintaining a list of lint paths.

cgsheeh (changes requested):

src/lando/tests/conftest.py
zzzeid: `lando.ui` will have many lint fixes if we do that. I'm on board with doing that in a separate PR and dealing with any issues that come about.

src/lando/dockerflow/views.py
ahochheiden: Done.

src/lando/tests/conftest.py
ahochheiden: I'll add the `tests` dir to the `LINT_PATHS` for now.

src/lando/dockerflow/views.py
```diff
@@ -0,0 +1,84 @@
...
+
+class VersionView(DockerflowView):
+    """
+    This view handles the version information of Lando.
+
+    It returns a JSON response containing the version information.
+    """
+
+    def get(self, request):
+        status = 200
```
zzzeid, line 34: nit: I do think it's cleaner to not rewrite this variable if there is a way around it (i.e., treat it the same way as you are treating `data` below.)

zzzeid (approved):

src/lando/dockerflow/views.py
ahochheiden: I was trying to follow the same pattern as `healthy = True` on #57, since I think you asked to move it outside the try/except there, too.

I'm fine with either, just trying to be consistent.

zzzeid: Oh - yes definitely want it to be outside the block, but the nit was about setting it to one value then overriding that value (vs. setting it only once). Hopefully this makes sense. This also applies to line 57 btw!

ahochheiden: Done.

zzzeid (approved):

cgsheeh (approved):

PR #60, opened 2024-05-17
formatting: Fix code style and code style tests (bug 1897434)

ahochheiden:
Also resolved the following warnings:
- `error: Failed to initialize cache at /.ruff_cache: Permission denied (os error 13)` by creating the directory and taking ownership of it in the Dockerfile.
- `warning: The top-level linter settings are deprecated in favour of their counterparts in the lint section. Please update the following options in code/ruff.toml:  - 'ignore' -> lint.ignore'  - 'select' -> 'lint.select'  - 'isort' -> 'lint.isort'` by making the suggested changes in the `ruff.toml`.

`test_ruff` and `test_black` are now both enabled and passing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,10 @@ RUN adduser --system --no-create-home app
 
 RUN mkdir /code
 COPY ./ /code
+
+RUN mkdir -p /code/.ruff_cache
+RUN chown -R app /code/.ruff_cache
+
 RUN pip install --upgrade pip
 
 
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,6 +1,6 @@
 line-length = 88
-select = ["C", "E", "F", "W", "B", "B9", "I"]
-ignore = ["B006", "B904", "C901", "E203", "E501"]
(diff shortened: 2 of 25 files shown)


zzzeid (approved):
I think this is good, one nit is that there might be a better way to implement the docker volumes so that they do not affect the production image, since the tests and linting only happen in the development environments. Not a blocker though.

Also I assume this directory will be ignored by source control? Wonder if it's better to just have everything go in `/tmp` or some other place.

cgsheeh (approved):

ahochheiden:
> I think this is good, one nit is that there might be a better way to implement the docker volumes so that they do not affect the production image, since the tests and linting only happen in the development environments. Not a blocker though.

I agree, that's worth investigating.

> Also I assume this directory will be ignored by source control? Wonder if it's better to just have everything go in `/tmp` or some other place.

`ruff` creates its own `.gitignore` when it populates the directory, and I don't think `git` picks up an empty directory, so it should always be ignored.

PR #64, opened 2024-05-21
lando: Fix `STATICFILES_DIRS` warning (bug 1897550)

ahochheiden:
Added a temporary file to ensure the `src\lando\static_src` directory exists to remove the `STATICFILES_DIRS` warning.

### DIFF
--- a/src/lando/static_src/TEMPORARY_FILE
+++ b/src/lando/static_src/TEMPORARY_FILE
@@ -1,0 +1,3 @@
+This file exists just to make sure this directory exists.
+
+It can be removed once actual static files are added.


zzzeid (approved):
nit: Could name the file `TEMPORARY_FILE` and put the details in the content of the file, to keep things a little cleaner.

ahochheiden:
> nit: Could name the file `TEMPORARY_FILE` and put the details in the content of the file, to keep things a little cleaner.

Sure. I was going for it being obnoxious so it's less likely to be missed when files are actually added, but your way sounds fine.

zzzeid:
> > nit: Could name the file `TEMPORARY_FILE` and put the details in the content of the file, to keep things a little cleaner.
> 
> Sure. I was going for it being obnoxious so it's less likely to be missed when files are actually added, but your way sounds fine.

Heh, either works!

PR #65, opened 2024-05-22
email: port mail functionality (bug 1888465)

zzzeid:
- update settings to support mail sending
- update functions and test to use django.core.mail
- remove legacy SMTP subsystem
- remove custom from email in method (use settings instead)
- remove mail suppression / allow list (no longer needed)


zzzeid:
Note: this hasn't been tested on an SMTP backend (tests are running on the console backend). Filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1898324) to track that work when we deploy this to GCP.

src/lando/api/legacy/tasks.py
```diff
@@ -102,28 +89,15 @@ def send_bug_update_failure_email(
...
-    with smtp.connection() as c:
-        c.send_message(
-            make_failure_email(
-                smtp.default_from,
-                recipient_email,
-                landing_job_identifier,
-                error_msg,
-                settings.LANDO_UI_URL,
-            )
+    with mail.get_connection() as c:
```
cgsheeh, line 92: Nit: could rename `c` -> `conn` or something while we're here.

src/lando/api/legacy/tasks.py
```diff
@@ -47,28 +47,15 @@ def send_landing_failure_email(
...
```
cgsheeh, line 50: Nit: could rename `c` -> `conn` or something while we're here.

cgsheeh (approved):

PR #67, opened 2024-05-23
celery: port celery functionality (bug 1870105)

zzzeid:
- move all celery functionality to lando.utils
- create new command to start celery worker
- add celery variables to settings
- add redis and celery to compose file
- remove some obsolete legacy code


src/lando/utils/tasks.py
```diff
@@ -9,17 +9,23 @@
...
 logger = logging.getLogger(__name__)
 
 
-@celery.task(
+@celery_app.task
+def debug_task():
+    print("hello")
+    return 1
+
+
```
cgsheeh, line 27: Leftover from dev?

src/lando/api/legacy/app.py
```diff
@@ -46,7 +44,6 @@ def load_config() -> dict[str, Any]:
     """Return configuration pulled from the environment."""
     config = {
         "ALEMBIC": {"script_location": "/migrations/"},
-        "DISABLE_CELERY": bool(os.getenv("DISABLE_CELERY")),
```
cgsheeh, line 49: Do we not need to disable celery any more? I don't see the corresponding variable in the settings module.

cgsheeh (changes requested):

src/lando/api/legacy/app.py
zzzeid: We've never used this variable as far as I can tell, and it doesn't provide any functionality that we need. When things were implemented manually in the "subsystem" (i.e., did not work out of the box) I think this made more sense to have. Now that we don't have to do anything extra to use Celery in Django, I don't think it makes sense to implement this unless the need for it arises.

src/lando/utils/tasks.py
zzzeid: Yeah, it will be useful once we deploy to DEV/STAGING etc... for testing celery integration, hence why I chose to keep it in.

cgsheeh: @zzzeid can you add a docstring describing that, how to use it etc.

cgsheeh (changes requested):
See comment about adding a docstring for the debug celery task.

cgsheeh (approved):

PR #70, opened 2024-05-30
staticfiles: port static file functionality (bug 1899629)

zzzeid:
- add django compressor and dependencies
- update settings file
- copy needed static source files from legacy repo
- update templates to support new functionality


zzzeid:
Closed, replaced by #71, #72, and #73.

PR #71, opened 2024-05-30
settings: add django-compressor support (bug 1899630)

zzzeid:
- add django-compressor to requirements file
- set COMPRESS_ENABLED to True for debugging purposes
- add compress extension to jinja environment


src/lando/settings.py
```diff
@@ -138,6 +139,28 @@
...
+    'css':[
+        'compressor.filters.css_default.CssAbsoluteFilter',
+        'compressor.filters.cssmin.rCSSMinFilter',
+    ],
+    'js':[
+        'compressor.filters.jsmin.JSMinFilter',
+    ]
+}
+
+COMPRESS_ENABLED = True
```
cgsheeh, line 162: Add some comments explaining these.

cgsheeh (changes requested):
Add some details to the commit message, or the bug, about what this is and why we need it.

src/lando/settings.py
zzzeid: This replaces flask-assets, along with all the asset-building functionality in the legacy lando-ui repo. Will add that to the commit message.

cgsheeh (approved):

PR #72, opened 2024-05-30
static files: move legacy files, re-add vendored files (bug 1899632)

zzzeid:


cgsheeh (approved):
I'm surprised to see these files as new additions to this repo. My understanding was that we merged the two repos into one, so these files should exist in this repo with their full history. In this case it's not a huge deal, I think the only files that really matter are a few of the `.scss` files and the `components/*.js` files, but I'm wondering if there are other files that will also lose history? Is there a way to bring these files in while maintaining the history?

zzzeid:
> I'm surprised to see these files as new additions to this repo. My understanding was that we merged the two repos into one, so these files should exist in this repo with their full history. In this case it's not a huge deal, I think the only files that really matter are a few of the `.scss` files and the `components/*.js` files, but I'm wondering if there are other files that will also lose history? Is there a way to bring these files in while maintaining the history?

I explicitly deleted these files in one of the previous commits, as I was planning on implementing a different approach to how we handle static files, however, due to time constraints this is going to be the faster approach to having things up and running without compromising on existing functionality.

The history of these files is still in the repo, though instead of them being moved, they are deleted and re-added.

zzzeid:
See #19 for context.

cgsheeh:
> The history of these files is still in the repo, though instead of them being moved, they are deleted and re-added.

If you checkout this PR locally and run `git log src/lando/static_src/legacy/js/components/Treestatus.js`, you won't see any commits in the log except for this one. So there is no `git blame` showing changes from my previous commits, for example.

Again, not a huge deal for these static UI files, and not worth blocking this PR over. For other files it would be a bigger deal so just want to make my questions more clear.

zzzeid:
> > The history of these files is still in the repo, though instead of them being moved, they are deleted and re-added.
> 
> If you checkout this PR locally and run `git log src/lando/static_src/legacy/js/components/Treestatus.js`, you won't see any commits in the log except for this one. So there is no `git blame` showing changes from my previous commits, for example.
> 
> Again, not a huge deal for these static UI files, and not worth blocking this PR over. For other files it would be a bigger deal so just want to make my questions more clear.

This is an anomaly, because I copied the latest files from Lando UI. If you do this for any other file, you will see the blame history. I can remove this file from here since it can be re-added in a later pull (and also, this would prevent a merge conflict). I think this, along with the css counterpart are the only two that will behave this way.

zzzeid:
Upon rethinking this, I think it would be better for the commit history to actually revert #19 and then remove the remainder of the files that were part of that PR. That way, everything will match with what it was in this repo originally.

cgsheeh:
> > > The history of these files is still in the repo, though instead of them being moved, they are deleted and re-added.
> > 
> > 
> > If you checkout this PR locally and run `git log src/lando/static_src/legacy/js/components/Treestatus.js`, you won't see any commits in the log except for this one. So there is no `git blame` showing changes from my previous commits, for example.
> > Again, not a huge deal for these static UI files, and not worth blocking this PR over. For other files it would be a bigger deal so just want to make my questions more clear.
> 
> This is an anomaly, because I copied the latest files from Lando UI. If you do this for any other file, you will see the blame history. I can remove this file from here since it can be re-added in a later pull (and also, this would prevent a merge conflict). I think this, along with the css counterpart are the only two that will behave this way.

I'm seeing the same thing for `git log src/lando/static_src/legacy/js/components/Stack.js`, and `git log src/lando/static_src/legacy/css/pages/RevisionPage.scss`, etc. Looks like the file was previously at `src/lando/ui/legacy/assets_src/css/pages/RevisionPage.scss`.

zzzeid:
> I'm seeing the same thing for `git log src/lando/static_src/legacy/js/components/Stack.js`, and `git log src/lando/static_src/legacy/css/pages/RevisionPage.scss`, etc. Looks like the file was previously at `src/lando/ui/legacy/assets_src/css/pages/RevisionPage.scss`.

You would have to omit the exact path (just use a wildcard for the first part) since those files have been moved.

zzzeid:
> > I'm seeing the same thing for `git log src/lando/static_src/legacy/js/components/Stack.js`, and `git log src/lando/static_src/legacy/css/pages/RevisionPage.scss`, etc. Looks like the file was previously at `src/lando/ui/legacy/assets_src/css/pages/RevisionPage.scss`.
> 
> You would have to omit the exact path (just use a wildcard for the first part) since those files have been moved.

I realized that I had already thought about this at the time when I made #19, and actually left all the custom files in the repo (doh!). I only deleted the vendored files in #19. Thanks for bringing this to my attention. I will update the PR to reflect that (i.e., just move the files over from ui/legacy).

cgsheeh (approved):
LGTM.

PR #75, opened 2024-06-03
models: add user profile model (bug 1899393)

zzzeid:
- add profile model to save various info related to user
- add imports to detect models in the `models` package


src/lando/main/models/__init__.py
zzzeid, line 1: This is so Django can automatically detect changes in models, since we moved models out of the original `models.py`.

src/lando/main/models/profile.py
```diff
@@ -0,0 +1,16 @@
...
+from django.db import models
+
+from lando.main.models.base import BaseModel
+
+
+class Profile(BaseModel):
+    user = models.OneToOneField(User, null=True, on_delete=models.SET_NULL)
+
+    # Phabricator API token.
+    phabricator_token = models.TextField(null=True, blank=True)
```
cgsheeh, line 13: Have we done an RRA for the auth changes? We probably should, especially since this change would store many Phab API tokens in plaintext in the Lando DB.

cgsheeh (changes requested):

src/lando/main/models/profile.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
+
+from django.contrib.auth.models import User
+from django.db import models
+
+from lando.main.models.base import BaseModel
+
+
+class Profile(BaseModel):
```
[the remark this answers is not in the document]
zzzeid: This is pretty typical/conventional naming, so that you can access the profile via `user.profile`. `user.lando_user` is a little more awkward, and having "user" in there is redundant. The `User` model itself is the Lando user.

src/lando/main/models/profile.py
zzzeid: No we have not, and this PR actually doesn't implement anything related to Phabricator API tokens, so I am going to remove this field from this PR. I'll schedule an RRA for this.

Also, in the long term, it would be great to automatically fetch (and possibly rotate) phabricator tokens. I'll file a bug for this as well.

src/lando/main/models/profile.py
```diff
@@ -0,0 +1,14 @@
...
+    """A model to store additional information about users."""
```
zzzeid, line 10:
```suggestion
    """A model to store additional information about users."""

```

src/lando/main/models/profile.py
zzzeid: Removed the Phabricator API key and updated PR.

src/lando/main/models/profile.py
```diff
@@ -0,0 +1,15 @@
...
+
+    user = models.OneToOneField(User, null=True, on_delete=models.SET_NULL)
+
+    # User info fetched from SSO.
+    userinfo = models.JSONField(default=dict, blank=True)
```
cgsheeh, line 15: Do we need to store the userinfo in the database? Could we just request it when the user takes an action on the page and we actually need to inspect it? I think it's worth considering now, since we can always add DB storage later if we need it for performance or other reasons.

cgsheeh (changes requested):
Question: the [docs](https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html#validate-id-tokens-by-renewing-them) mention using the `mozilla_django_oidc.middleware.SessionRefresh` middleware to handle validating and refreshing Auth0 tokens. Do we need to add that or are we okay without it?

src/lando/main/models/profile.py
zzzeid: `userinfo` has other information that we will need and will be using (e.g., name, nickname, email, and even the avatar URL). This is a low-maintenance way of mapping this information without creating individual fields. Remember that userinfo is [stored in the session](https://flask-pyoidc.readthedocs.io/en/latest/quickstart.html#openid-connect) in the legacy Lando, but due to the limitations of having two separate applications (UI/API), it just happened that we used a cookie storage instead of server-side storage, and we did not have the concept of a User outside of that session framework, but now we do.

I'm not opposed to saving those fields individually on the profile or user model instead and to process userinfo groups directly and store them as permissions, but I think even if we go that approach, it is still useful to have the original record for both validation (see below) and for debugging purposes. I don't see an advantage of storing userinfo exclusively in the session like it is done in the legacy Lando.

But the short answer is yes, we should at least store those fields I mentioned above, and the permissions that are derived from the groups. Part of this overhaul is to leverage Django's auth system by determining user permissions -- as I've done in the PR. However we may have to _validate_ them periodically or when an action needs it (see [bug 1899399](https://bugzilla.mozilla.org/show_bug.cgi?id=1899399). I think this specific problem should be punted to further down the road though as part of that bug.

zzzeid:
> Question: the [docs](https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html#validate-id-tokens-by-renewing-them) mention using the `mozilla_django_oidc.middleware.SessionRefresh` middleware to handle validating and refreshing Auth0 tokens. Do we need to add that or are we okay without it?

Yes, I think this is out of scope for this PR but can be implemented as part of [bug 1899399](https://bugzilla.mozilla.org/show_bug.cgi?id=1899399).

src/lando/main/models/profile.py
cgsheeh: > `userinfo` has other information that we will need and will be using (e.g., name, nickname, email, and even the avatar URL). This is a low-maintenance way of mapping this information without creating individual fields. Remember that userinfo is [stored in the session](https://flask-pyoidc.readthedocs.io/en/latest/quickstart.html#openid-connect) in the legacy Lando, but due to the limitations of having two separate applications (UI/API), it just happened that we used a cookie storage instead of server-side storage, and we did not have the concept of a User outside of that session framework, but now we do.

In the current Lando setup, the access token and ID token are stored as cookies in the client app, but the userinfo is not. We [request the userinfo](https://github.com/mozilla-conduit/lando-api/blob/046015ebb7b655cefc0de790e35a9bd4cbaf34a7/landoapi/auth.py#L419) on the LandoAPI side using the access token for every authenticated request which requires the userinfo. So what I'm really asking is, can we continue doing that and avoid having to store the userinfo and handle expiration of the tokens manually?

Perhaps bug 1899399 is where we plan to add the `mozilla_django_oidc.middleware.SessionRefresh` middleware, however given we are manually syncing the userinfo SCM levels into our own permissions system, it seems pertinent that we handle token expiration in a timely fashion.

cgsheeh (approved):

src/lando/main/models/profile.py
zzzeid: > In the current Lando setup, the access token and ID token are stored as cookies in the client app, but the userinfo is not. 

This is incorrect. Although we do request userinfo (and cache it) for various requests on-the-fly, the full userinfo is still stored in the session cookie as part of flask-pyoidc. We use it, for example, to access the avatar URL when showing that in the UI.

> So what I'm really asking is, can we continue doing that and avoid having to store the userinfo and handle expiration of the tokens manually?

We won't necessarily handle the expiration of tokens manually, and as part of reimplementing the `require_auth0` decorators ([bug 1900642](https://bugzilla.mozilla.org/show_bug.cgi?id=1900642)) we can handle more sensitive or urgent validation of permission groups. The legacy implementation was limited by the fact that there is no user or permission framework in Flask, and by the fact that we store things in the session cookie (as there is no database backend for Lando UI). Now that we have a user model and an authentication framework with permission, I want to leverage that as much as possible as part of this implementation.

[Bug 1870108](https://bugzilla.mozilla.org/show_bug.cgi?id=1870108) includes implementing all improvements to the auth system that can be made by not copying over things that were implemented due to limitations (or due to technical debt).

zzzeid: > it seems pertinent that we handle token expiration in a timely fashion.

[Bug 1899399](https://bugzilla.mozilla.org/show_bug.cgi?id=1899399) blocks [bug 1870108](https://bugzilla.mozilla.org/show_bug.cgi?id=1870108), so it will be handled before that bug is closed.

PR #76, opened 2024-06-03
auth: add and use custom backend (bug 1899395)

zzzeid:
- extend OIDCAuthenticationBackend
- add scm permission to the Profile model
- sync permissions during auth cycle
- add tests
- add list of current SCM groups
- remove saving of userinfo in backend instance, use claims


src/lando/main/auth.py
```diff
@@ -0,0 +1,15 @@
+from mozilla_django_oidc.auth import OIDCAuthenticationBackend
+
+
+class LandoOIDCAuthenticationBackend(OIDCAuthenticationBackend):
```
cgsheeh, line 4: Add a docstring explaining the purpose of this class.

cgsheeh (changes requested):

cgsheeh (approved):
r+, but again, just consider if we could `get_userinfo()` whenever we need it and avoid saving to the DB and dealing with syncing/expiration etc.

zzzeid:
> r+, but again, just consider if we could `get_userinfo()` whenever we need it and avoid saving to the DB and dealing with syncing/expiration etc.



> ust consider if we could `get_userinfo()` whenever we need it and avoid saving to the DB and dealing with syncing/expiration etc.

As mentioned in the other PR, leveraging the Django auth system abstraction is a core part of porting the auth functionality in Lando. Additional follow up PRs can fine tune this functionality and handle expiration/syncing (those are common problems that can be solved), but leveraging the framework as intended is good as a first step here.

cgsheeh:
> > r+, but again, just consider if we could get_userinfo() whenever we need it and avoid saving to the DB and dealing with syncing/expiration etc.
> 
> As mentioned in the other PR, leveraging the Django auth system abstraction is a core part of porting the auth functionality in Lando. Additional follow up PRs can fine tune this functionality and handle expiration/syncing (those are common problems that can be solved), but leveraging the framework as intended is good as a first step here.

I'm not suggesting we don't leverage the Django auth system, I'm asking if we need to store the data returned by Auth0 in the DB or if we can just request it when we need it and update the permissions accordingly.

zzzeid:
> > > r+, but again, just consider if we could get_userinfo() whenever we need it and avoid saving to the DB and dealing with syncing/expiration etc.
> > 
> > 
> > As mentioned in the other PR, leveraging the Django auth system abstraction is a core part of porting the auth functionality in Lando. Additional follow up PRs can fine tune this functionality and handle expiration/syncing (those are common problems that can be solved), but leveraging the framework as intended is good as a first step here.
> 
> I'm not suggesting we don't leverage the Django auth system, I'm asking if we need to store the data returned by Auth0 in the DB or if we can just request it when we need it and update the permissions accordingly.

RE: "avoid saving to the DB and dealing with syncing/expiration": There are things we need to store/sync, as I mentioned (e.g., nickname, avatar URL, name etc.) and those are things we are already doing in Lando today, just in a different way (flask-pyoidc is handling that and storing them in the session, as mentioned before). The group memberships happen to be in the same place, and for both of these things, there is no way around having to deal with syncing/expiration. The way we dealt with it in the legacy lando is to disregard the saved groups in the session, and fetching them as-needed. Moving the (re)fetching of userinfo to "when we need it" is one way to deal with syncing/expiration, and we may very well need to do that but as mentioned that will be handled in a follow up bug. Storing userinfo "as is" more or less follows what we are doing today in Lando for user-profile related things.

PR #77, opened 2024-06-03
auth: add scm permissions to model, sync (bug 1899395)

zzzeid:
- add scm permission to the Profile model
- sync permissions during auth cycle
- add tests
- add list of current SCM groups
- remove saving of userinfo in backend instance, use claims


src/lando/main/models/profile.py
```diff
@@ -1,15 +1,44 @@
...
+    class Meta:
+        permissions = SCM_PERMISSIONS
+
     user = models.OneToOneField(User, null=True, on_delete=models.SET_NULL)
 
     # User info fetched from SSO.
     userinfo = models.JSONField(default=dict, blank=True)
+
+    def update_permissions(self):
+        """Remove permissions (currently SCM) and re-add them based on userinfo."""
```
cgsheeh, line 30:
```suggestion
        """Remove permissions and re-add them based on userinfo."""
```

src/lando/main/models/profile.py
```diff
@@ -1,15 +1,44 @@
...
 from lando.main.models.base import BaseModel
 
+SCM_PERMISSIONS = (
+    ("scm_conduit", "SCM_CONDUIT"),
+    ("scm_level_1", "SCM_LEVEL_1"),
+    ("scm_level_2", "SCM_LEVEL_2"),
+    ("scm_level_3", "SCM_LEVEL_3"),
+    ("scm_versioncontrol", "SCM_VERSIONCONTROL"),
+)
+
```
cgsheeh, line 16: This list is missing quite a few scm levels we support in Lando currently (`scm_nss`, `scm_firefoxci`, etc). See `repos.py` in LandoAPI. We should add all the SCM levels for repos we support in Lando.

cgsheeh (changes requested):

src/lando/main/models/profile.py
zzzeid: > 

The reason I state "currently SCM" is that other user permissions are not affected by this method at this time. I will reword it to say "Remove SCM permissions and re-add them based on userinfo" instead.

src/lando/main/models/profile.py
zzzeid: Makes sense, will add.

src/lando/main/models/profile.py
```diff
@@ -1,15 +1,44 @@
...
+        groups = self.userinfo.get("https://sso.mozilla.com/claim/groups", [])
+        content_type = ContentType.objects.get_for_model(self.__class__)
+
+        for codename in SCM_PERMISSIONS:
+            permission = Permission.objects.get(
+                codename=codename,
+                content_type=content_type,
+            )
+            if f"all_{permission}" in groups and f"expired_{permission}" not in groups:
```
zzzeid, line 43: @cgsheeh I think this is wrong? Should it be this instead?

```suggestion
            if f"active_{permission}" in groups and f"expired_{permission}" not in groups:
```

Won't be able to test this properly until we have DEV up and running. Though I imagine having `all_<something>` and not having `expired_<something>` should always line up with having `active_<something>`.

src/lando/main/models/profile.py
```diff
@@ -1,15 +1,44 @@
...
 from django.db import models
 
 from lando.main.models.base import BaseModel
 
+SCM_PERMISSIONS = (
+    ("scm_conduit", "SCM_CONDUIT"),
+    ("scm_level_1", "SCM_LEVEL_1"),
+    ("scm_level_2", "SCM_LEVEL_2"),
+    ("scm_level_3", "SCM_LEVEL_3"),
+    ("scm_versioncontrol", "SCM_VERSIONCONTROL"),
```
zzzeid, line 14:
```suggestion
    ("scm_allow_direct_push", "SCM_ALLOW_DIRECT_PUSH"),                                                                                                                                                               
    ("scm_conduit", "SCM_CONDUIT"),
    ("scm_firefoxci", "SCM_FIREFOXCI"),
    ("scm_l10n_infra", "SCM_L10N_INFRA"),
    ("scm_level_1", "SCM_LEVEL_1"),
    ("scm_level_2", "SCM_LEVEL_2"),
    ("scm_level_3", "SCM_LEVEL_3"),
    ("scm_nss", "SCM_NSS"),
    ("scm_versioncontrol", "SCM_VERSIONCONTROL"),

```

cgsheeh:
Can we test this somehow at this stage? If we can't, we should file a bug as a reminder to write some.

zzzeid:
> Can we test this somehow at this stage? If we can't, we should file a bug as a reminder to write some.

Yes I will add some tests.

src/lando/main/auth.py
```diff
@@ -1,14 +1,30 @@
...
+from lando.main.models.profile import Profile
+
 
 class LandoOIDCAuthenticationBackend(OIDCAuthenticationBackend):
     """An extended OIDC auth backend that manipulates the user profile."""
 
     def get_userinfo(self, *args, **kwargs):
+        # Store fetched userinfo on this instance.
         self.userinfo = super().get_userinfo(*args, **kwargs)
         return self.userinfo
```
zzzeid, line 13: I realized that this is not needed, as [`get_or_create_user`](https://github.com/mozilla/mozilla-django-oidc/blob/main/mozilla_django_oidc/auth.py#L358) passes the entirety of `userinfo` as `claims`.

src/lando/main/models/profile.py
```diff
@@ -1,15 +1,55 @@
...
+
+        self.user.user_permissions.remove(*permissions.values())
+
+        groups = self.userinfo.get(CLAIM_GROUPS_KEY, [])
+
+        for codename in permissions:
+            if (
+                set(groups).intersection((f"all_{codename}", f"active_{codename}"))
+                and f"expired_{codename}" not in groups
+            ):
```
cgsheeh, line 54: Nit: this logic could go into a separate function.

src/lando/main/tests.py
```diff
@@ -1 +1,31 @@
...
+
+
+@pytest.mark.django_db(transaction=True)
+def test_LandoOIDCAuthenticationBackend__update_user_scm_access(monkeypatch):
+    backend = LandoOIDCAuthenticationBackend()
+
+    # Test that having the active/all groups will add the correct permission.
+    groups = ["active_scm_conduit", "all_scm_conduit"]
+    claims = {CLAIM_GROUPS_KEY: groups}
+    user = User.objects.create_user(username="test_user", password="test_password")
```
zzzeid, line 15:
```suggestion
    backend = LandoOIDCAuthenticationBackend()
    user = User.objects.create_user(username="test_user", password="test_password")

    # Test that having the active/all groups will add the correct permission.
    groups = ["active_scm_conduit", "all_scm_conduit"]
    claims = {CLAIM_GROUPS_KEY: groups}
```

src/lando/main/tests.py
```diff
@@ -1 +1,31 @@
...
+    user = User.objects.get(username="test_user")
+    backend.update_user(user, claims)
+    assert not user.has_perm("main.scm_conduit")
+
+    # Test that an ambiguous situation will result in no permission.
+    groups = ["expired_scm_conduit", "all_scm_conduit", "active_scm_conduit"]
+    claims = {CLAIM_GROUPS_KEY: groups}
+    user = User.objects.get(username="test_user")
+    backend.update_user(user, claims)
+    assert not user.has_perm("main.scm_conduit")
```
cgsheeh, line 31: Seems like this should be 3 separate tests, or a parameterized test.

cgsheeh (approved):

src/lando/main/models/profile.py
zzzeid: Good idea.

cgsheeh (approved):

PR #80, opened 2024-06-20
auth: trigger user profile creation on first login (bug 1903802)

zzzeid:


cgsheeh (approved):
Seems reasonable. Can you add type hints to these functions?

cgsheeh (approved):

PR #85, opened 2024-07-15
stacks: raise Http404 where applicable (bug 1907913)

zzzeid:


src/lando/api/legacy/api/stacks.py
```diff
@@ -67,13 +60,13 @@ def get(phab: PhabricatorClient, revision_id: str):
     )
     revision = phab.single(revision, "data", none_when_empty=True)
     if revision is None:
-        return not_found_problem
+        raise Http404
```
cgsheeh, line 63: We should add some parameters so it's more obvious what is missing, ie `raise Http404("The requested revision does not exist")`

cgsheeh (changes requested):

src/lando/api/legacy/api/stacks.py
zzzeid: I am re-thinking the need to have a separate "revision not found" functionality vs. just a standard 404. I'm not sure if adds any value, and it also simplifies the implementation of the "fake" 404 we provide when a user is not authorized to view a revision. I don't see the value in having custom 404s depending on what type of page you are trying to access. If this was the API I think it could be useful to have a more specific error, but in this case Django is going to interpret this as a standard 404 no matter what string we pass here. Thoughts?

cgsheeh: According to [the docs](https://docs.djangoproject.com/en/5.0/topics/http/views/#the-http404-exception) the error message only appears in debug mode, so I think we ought to include it.

zzzeid: Fair enough, it certainly doesn't hurt.

src/lando/api/legacy/api/stacks.py
```diff
@@ -67,13 +60,13 @@ def get(phab: PhabricatorClient, revision_id: str):
...
 
     nodes, edges = build_stack_graph(revision)
     try:
         stack_data = request_extended_revision_data(phab, list(nodes))
     except ValueError:
-        return not_found_problem
+        raise Http404
```
zzzeid, line 69:
```suggestion
        raise Http404("Revision does not exist")
```

src/lando/api/legacy/api/stacks.py
zzzeid, line 63:
```suggestion
        raise Http404("Revision does not exist")
```

src/lando/api/legacy/api/stacks.py
```diff
@@ -67,13 +60,13 @@ def get(phab: PhabricatorClient, revision_id: str):
...
     if revision is None:
-        return not_found_problem
+        raise Http404("Revision does not exist")
 
     nodes, edges = build_stack_graph(revision)
     try:
         stack_data = request_extended_revision_data(phab, list(nodes))
     except ValueError:
-        return not_found_problem
+        raise Http404("Revision does not exist")
```
zzzeid, line 69: Interesting though that on this particular line, this text might be misleading. I don't think it necessarily makes sense to include a more verbose message, since it could potentially expose a vulnerability (e.g., someone ends up seeing the custom message for whatever reason) since this particular error here is usually triggered due to access restrictions.

It's also not very clear that this text is only for debug purposes, and I still am not sure if it adds any value. In any case, I think we should keep it the same as the other error, or remove the error from both.

cgsheeh (approved):

PR #86, opened 2024-07-15
api: handle auth and requests correctly

zzzeid:
- pass request to views where needed (bug 1907916)
- implement decorator to replace require_auth0 (bug 1900642)
- fetch ldap_username from correct source (request)
- reimplement require_phabricator_api_key decorator (bug 1908717)
- remove deprecated usages of `g` variable
- implement require_permission decorator
- remove legacy decorators module (no longer needed)
- replace all instances of auth0_user with lando_user
- adapt tests to new functionality


src/lando/api/legacy/api/stacks.py
```diff
@@ -54,12 +55,13 @@
 
 
 @require_phabricator_api_key(optional=True)
-def get(phab: PhabricatorClient, revision_id: str):
+def get(phab: PhabricatorClient, request: HttpRequest, revision_id: str):
     """Get the stack a revision is part of.
 
     Args:
         revision_id: (string) ID of the revision in 'D{number}' format
```
zzzeid, line 62: This should probably be updated to reflect what the URL dispatcher is actually passing here (i.e., just the number portion of the revision ID.)

src/lando/api/legacy/api/try_push.py
```diff
@@ -98,10 +98,8 @@ def parse_revisions_from_request(
         )
 
 
-@auth.require_auth0(scopes=("openid", "lando", "profile", "email"), userinfo=True)
-# Re-enable this check once our Auth0 instance returns group membership for access
-# tokens granted via the Device Authorization flow.
-# @auth.enforce_request_scm_level(SCM_LEVEL_1)
+@require_authenticated_user
+@require_permission("scm_level_1")
```
cgsheeh, line 102: Tokens granted via Device Authorization Flow don't include scm levels, so we can't enforce the permissions here. Please re-add the comment from the original.

src/lando/api/legacy/transplants.py
```diff
@@ -376,23 +376,23 @@ def warning_unresolved_comments(*, phab, revision, **kwargs):
...
-    if auth0_user.is_in_groups(landing_repo.access_group.active_group):
+    if lando_user.has_perm(
+        f"main.{landing_repo.access_group.membership_group.removeprefix('all_')}"
+    ):
         return None
 
-    if auth0_user.is_in_groups(landing_repo.access_group.membership_group):
-        return "Your {} has expired.".format(landing_repo.access_group.display_name)
-
+    # TODO: update copy to include possibility of expired access?
```
cgsheeh, line 395: If the user doesn't have the Django permission, we should still be able to check `userinfo` directly to see if they are in the `expired_scm_level` group.

src/lando/api/legacy/api/stacks.py
```diff
@@ -49,12 +49,13 @@
...
```
zzzeid, line 56:
```suggestion
        revision_id: (int) ID (number portion) of the revision in D{number} format
```

src/lando/api/legacy/api/stacks.py
zzzeid: On second look, the conversion is no longer necessary, as the only time it was used was when converting it back to an int. I will consolidate the functionality into a single revision ID variable.

src/lando/main/auth.py
```diff
@@ -41,3 +48,84 @@ def create_user(self, claims: dict) -> User:
...
+def require_authenticated_user(f):
+    """
+    Decorator which requires a user to be authenticated.
+
+    Raises a PermissionError if a request is by an unauthenticated user.
+    """
+
+    def wrapper(request, *args, **kwargs):
+        if not request.user.is_authenticated:
+            raise PermissionError
```
cgsheeh, line 62:
```suggestion
            raise PermissionError("Authentication is required.")
```

or something like that.

src/lando/main/auth.py
```diff
@@ -41,3 +48,84 @@ def create_user(self, claims: dict) -> User:
...
+    """
+    Decorator that raises a PermissionError if a user is missing the given permission.
+    """
+
+    def __init__(self, permission):
+        self.required_permission = permission
+
+    def __call__(self, f: Callable) -> Callable:
+        def wrapper(request, *args, **kwargs):
+            if not request.user.has_perm(f"main.{self.required_permission}"):
```
cgsheeh, line 78: We should check for expired permissions here and indicate them in the exception message (assuming those eventually make it to the user).

src/lando/main/auth.py
```diff
@@ -41,3 +48,84 @@ def create_user(self, claims: dict) -> User:
...
+
+
+class require_phabricator_api_key:
+    """Decorator which requires and verifies the phabricator API Key.
+
+    If a user's API key is not available and optional is False, then
+    an HTTP 401 response will be returned.
+
+    The user's API key will be verified to be valid, if it is not an
+    HTTP 403 response will be returned..
```
cgsheeh, line 90:
```suggestion
    HTTP 403 response will be returned.
```

src/lando/main/auth.py
```diff
@@ -41,3 +48,84 @@ def create_user(self, claims: dict) -> User:
...
+
+    def __call__(self, f: Callable) -> Callable:
+        def wrapper(request, *args, **kwargs):
+            if not request.user.has_perm(f"main.{self.required_permission}"):
+                raise PermissionError()
+            return f(request, *args, **kwargs)
+
+
+class require_phabricator_api_key:
+    """Decorator which requires and verifies the phabricator API Key.
```
cgsheeh, line 84:
```suggestion
    """Decorator which requires and verifies the Phabricator API Key.
```

cgsheeh (changes requested):

src/lando/api/legacy/api/try_push.py
zzzeid: > Tokens granted via Device Authorization Flow don't include scm levels, so we can't enforce the permissions here. Please re-add the comment from the original.

So as implemented in this PR, this endpoint will require a user to be authenticated via Django's auth framework, and thus we can directly check whether the permission is there or not. This endpoint will not work as it did previously (e.g., via an auth token). We should file a bug to implement this change in Mach (and Lando, since we don't currently have a way to authenticate with Lando via CLI). I will add a comment and file a bug for this.

src/lando/api/legacy/transplants.py
zzzeid: I have a separate bug filed for this: https://bugzilla.mozilla.org/show_bug.cgi?id=1908702. I think this can be discussed separately from this implementation.

src/lando/main/auth.py
zzzeid: Can we consolidate this discussion in [bug 1908702](https://bugzilla.mozilla.org/show_bug.cgi?id=1908702)?

src/lando/api/legacy/api/try_push.py
cgsheeh: Please describe the change you think we would make here in more detail. How will we implement authentication to Lando in Mach without using an access token granted from Auth0? The statement "we don't currently have a way to authenticate with Lando via CLI" is not accurate, we can authenticate to Lando via the CLI, but the granted access token doesn't include the `https://sso.mozilla.com/claim/groups` key with group membership information.

zzzeid: > The statement "we don't currently have a way to authenticate with Lando via CLI" is not accurate

My statement is referring to new Lando, not legacy Lando -- should have been less ambiguous about that.

> How will we implement authentication to Lando in Mach without using an access token granted from Auth0?

This is something yet to be determined. We could do it via Auth0 still, but ultimately we would have to authenticate with Django so that all the requests get tied to the Django user. Regardless, the commented out portion here would be irrelevant since Django can check the permissions directly.

cgsheeh: It sounds like you're suggesting we roll our own auth system for Lando. Without more details, it sounds like this is far outside the scope of the Django migration. Keep in mind that while we are trying to leverage the Django permissions system, Auth0 is still the system which provides information about the users and their permissions (including important things like removing/disabling users when they leave the company). I don't think we can or should be implementing a form of auth that doesn't use Auth0 access tokens.

zzzeid: > It sounds like you're suggesting we roll our own auth system for Lando.

Not sure what you mean here? Perhaps you can elaborate. Since the endpoint is a Django one, therefore Django will have to determine if a user is authorized or not.

The new Django-based auth system is already part of the scope of this Lando porting, so is improving/"modernizing" the auth story in Lando. The way we are initially using it is via an Auth0 backend to authenticate users, and to parse user permissions via userinfo.

After chatting with @globau we've decided that the particular auth implementation for this particular endpoint (try) can be deferred to after the initial launch. One possibility is to use the Auth0 token to authenticate the user with Django directly via the endpoint (and create the user/permissions at that point), but there are other possible approaches as well that we can discuss.

zzzeid: Filed [bug 1909723](https://bugzilla.mozilla.org/show_bug.cgi?id=1909723).

cgsheeh: > > It sounds like you're suggesting we roll our own auth system for Lando.
> 
> Not sure what you mean here? Perhaps you can elaborate. Since the endpoint is a Django one, therefore Django will have to determine if a user is authorized or not.

You previously stated the "endpoint will not work as it did previously (e.g., via an auth token)". Since you haven't given any more details ("This is something yet to be determined"), it sounds like you are intending to implement some non-Auth0 access token authentication to Lando.

zzzeid: > it sounds like you are intending to implement some non-Auth0 access token authentication to Lando.

That's still a possibility (i.e., authenticating directly with Lando via the CLI, or a web page, or in some other way). Ultimately Lando will have a user record for each user who authenticates one way or another, and those users as long as they exist in Auth0 can be tied together regardless of how they originate in the system (e.g., via Auth0 token or via a web login). In any case, one simple approach (if indeed possible) would be to use Lando to fill in the missing information after validating the token that is generated with the current approach in Mach.

zzzeid: > "endpoint will not work as it did previously (e.g., via an auth token)"

Lastly to clarify this point, this is correct. With this implementation, the endpoint does not work with the auth token (in fact, no endpoint in Django supports an auth token at this point).

src/lando/api/tests/test_transplants.py
```diff
@@ -332,19 +335,13 @@ def test_dryrun_outside_codefreeze(
...
     [
         (
-            CANNED_USERINFO["NO_CUSTOM_CLAIMS"],
-            200,
-            "You have insufficient permissions to land. Level 3 "
-            "Commit Access is required. See the FAQ for help.",
-        ),
-        (CANNED_USERINFO["EXPIRED_L3"], 200, "Your Level 3 Commit Access has expired."),
-        (
-            CANNED_USERINFO["UNVERIFIED_EMAIL"],
```
zzzeid, line 345: See https://bugzilla.mozilla.org/show_bug.cgi?id=1908701.

cgsheeh (approved):
As long as we resolve bug 1909723 and bug 1908702, this is good to go.

PR #87, opened 2024-07-16
ui: port error pages (bug 1902107)

zzzeid:
- move base templates
- implement 404 and 500 templates
- remove unused legacy code


src/lando/ui/jinja2/500.html
```diff
@@ -0,0 +1,3 @@
+{% extends "default_error.html" %}
+{% set title = "500 - Oops" %}
+{% set message = "Something went wrong. Someone will look into this shortly." %}
```
cgsheeh, line 3: We should add a hint about reporting a bug, or notifying someone on Matrix.

src/lando/ui/jinja2/errorhandlers/revisions_404.html
```diff
@@ -1,21 +0,0 @@
...
-
-  <h1>Revision/Diff Not Available</h1>
-
-  <p>
-    The revision or diff you've requested does not exist or you do not have
-    permission to view it.
-  </p>
-  {% if user_is_authenticated and not user_has_phabricator_token %}
-    <p>
-      If you were trying to view a stack which includes a secure revision, you need to
```
cgsheeh, line 15: With this patch we remove the hint on the 404 page about adding the Phab API token. Could we add a `hint` parameter to the 404 page so we can add custom messaging to some 404's where appropriate? This message was added in https://bugzilla.mozilla.org/show_bug.cgi?id=1808042 at Sylvestre's request.

cgsheeh (changes requested):

cgsheeh (approved):

PR #88, opened 2024-07-16
jinja: reimplement auth template variables (bug 1908137)

zzzeid:
- reimplement user_has_phabricator_token
- reimplement user_is_authenticated (originally is_user_authenticated)

### DIFF
--- a/src/lando/ui/jinja2/partials/layout.html
+++ b/src/lando/ui/jinja2/partials/layout.html
@@ -1,3 +1,11 @@
+{% if request %}
+    {% set user_is_authenticated = request.user.is_authenticated %}
+    {% set user_has_phabricator_token = request.user.profile and request.user.profile.phabricator_api_key %}
+{% else %}
+    {% set user_is_authenticated = False %}
+    {% set user_has_phabricator_token = False %}
+    {% set request = None %}
+{% endif %}
 <!DOCTYPE html>
 <html>
 <head>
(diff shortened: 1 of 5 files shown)


src/lando/ui/jinja2/partials/layout.html
```diff
@@ -1,3 +1,11 @@
+{% if request %}
```
cgsheeh, line 1: I'm very surprised there is no way to add our own custom variables/functions/etc to templates via Python.

cgsheeh (approved):

src/lando/ui/jinja2/partials/layout.html
zzzeid: > I'm very surprised there is no way to add our own custom variables/functions/etc to templates via Python.

There is, it's just that the request context is not readily available for those variables/functions. Technically you could implement a method that gets passed the request and call it here (e.g., `user_is_authenticated = check_if_user_is_authenticated(request)`).

In the case of `user_is_authenticated`, the variable is very simple and more so used like an alias, and doesn't really need a separate method.

zzzeid: I guess implementing that method, even if it seems kind of redundant, will remove the need to include an if statement in the template. So upon rethinking this, I think it's OK to do that here.

zzzeid: Filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1909734) to track the improvement to this.

PR #101, opened 2024-08-13
models: various fixes (bug 1912967)

zzzeid:
- fix import of Worker model
- filter repos by name
- add __str__ methods to various models to improve readability
- add enabled_repo_names shim
- fix related_name on unsorted_revisions field
- add temporary placeholder for phabricator_api_key
- update migrations


src/lando/main/management/commands/__init__.py
```diff
@@ -1,6 +1,6 @@
 from time import sleep
 
-from lando.main.models import Worker
+from lando.main.models.base import Worker
```
dklawren, line 3: This will need to be changed based on the other PR.

src/lando/main/management/commands/landing_worker.py
```diff
@@ -8,7 +8,8 @@
 from django.core.management.base import BaseCommand
 from django.db import transaction
 from lando.main.management.commands import WorkerMixin
-from lando.main.models import LandingJob, LandingJobStatus
+from lando.main.models.base import Repo
```
dklawren, line 11: Same here

dklawren (changes requested):

src/lando/main/management/commands/__init__.py
zzzeid: Ah, let me merge those other ones first and then I will request re-review after updating this one.

dklawren (approved):

PR #105, opened 2024-08-15
utils: add command to create environment repos (bug 1913173)

zzzeid:
- create command to generate repos based on legacy configuration
- limit support to lower environments for time being


dklawren (approved):
LGTM and works as expected

PR #107, opened 2024-08-20
landing_worker: implement landing worker management command (bug 1914016)

zzzeid:
- create shell command that starts legacy landing worker
- pass repos to worker instance instead of fetching from env
- set target repo when triggering landing job
- initialize treestatus client in worker instance
- update tests
- add TREESTATUS_URL to settings
- rename worker to "hg-landing-worker"
- remove unneeded treestatus subsystem initialization
- raise and catch ConnectionError when initializing legacy worker
- remove unused variable from `job_processing`


src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -47,14 +44,13 @@
 
 
 @contextmanager
-def job_processing(worker: LandingWorker, job: LandingJob):
```
cgsheeh, line 50: The signature is updated here but there are no changes to callers, guessing this belongs in a different PR.

src/lando/api/legacy/workers/base.py
```diff
@@ -33,24 +34,24 @@ def PAUSE_KEY(self) -> ConfigurationKey:
         """Return the configuration key that pauses the worker."""
         raise NotImplementedError()
 
-    def __init__(self, sleep_seconds: float = 5, with_ssh: bool = True):
+    def __init__(self, repos, sleep_seconds: float = 5, with_ssh: bool = True):
```
cgsheeh, line 37: Type hint missing.

cgsheeh (changes requested):
What is the "simple wrapper" mentioned in the title?

zzzeid:
> What is the "simple wrapper" mentioned in the title?

The management command is basically it. I'll update the title to be a little more specific about this.

src/lando/api/legacy/workers/landing_worker.py
zzzeid: I'll add this in the commit message. It's an unused variable that I just removed while I was here.

zzzeid: Also forgot to update the caller, which is now done. This was pulled from a larger PR earlier, but I think it makes more sense to include it in this one. In any case, the latest 3 PRs are closely related, the last in the chain being #108

cgsheeh (approved):

PR #108, opened 2024-08-22
landing_worker: add hg repo cloning (bug 1913489)

zzzeid:
- add hg_repo_is_initialized property
- add hg_repo_prepare method
- iterate over enabled repos in worker and prepare


src/lando/main/models/repo.py
```diff
@@ -99,6 +100,28 @@ def raise_if_not(self, repo_scm):
         if repo_scm != self.scm:
             raise self._method_not_supported_for_repo_error
 
+    @property
+    def hg_repo_is_initialized(self):
```
cgsheeh, line 104: Type hints and docstring.

src/lando/main/models/repo.py
```diff
@@ -99,6 +100,28 @@ def raise_if_not(self, repo_scm):
...
+        self.raise_if_not(self.HG)
+        try:
+            self.hg._open()
+        except hglib.error.ServerError:
+            logger.info(f"{self} appears to be not initialized.")
+            return False
+        else:
+            return True
+
+    def hg_repo_prepare(self):
```
cgsheeh, line 114: Type hints and docstring.

From reading this it seems like we will likely have a `git_repo_prepare` function and a `git_repo_is_initialized`. Instead, why don't we run a `repo.assert_scm({hg/git})` and then have a more generic `repo_prepare` that does the correct setup based on the SCM for the repo?

src/lando/main/models/repo.py
```diff
@@ -99,6 +100,28 @@ def raise_if_not(self, repo_scm):
...
+        self.raise_if_not(self.HG)
+        if not self.hg_repo_is_initialized:
+            Path(self.system_path).mkdir(parents=True, exist_ok=True)
```
cgsheeh, line 117: Seems like `self.system_path` should already be a `Path`, based on the variable name.

src/lando/main/models/repo.py
```diff
@@ -99,6 +100,28 @@ def raise_if_not(self, repo_scm):
...
+            logger.info(f"Cloning {self} from pull path.")
+            self.hg.clone(self.pull_path)
+        else:
```
cgsheeh, line 120: Nit: you can `return` in the `if` block above and de-dent this.

cgsheeh (changes requested):

src/lando/main/models/repo.py
zzzeid: Git support is out of scope for this PR. Some of the git support was implemented in the prototype worker (which is still in the code base) and will be reimplemented/revisisted/combined with this as part of [bug 1895523](https://bugzilla.mozilla.org/show_bug.cgi?id=1895523). This applies for other shared aspects between git/hg repos.

src/lando/main/models/repo.py
zzzeid: This is not correct, it is a string (see the [docs](https://docs.djangoproject.com/en/5.1/ref/models/fields/#filepathfield)). There isn't a convention of a variable ending in `_path` to be of a certain type, it is referring to what the string represents (i.e., a path and not a filename, for example.).

src/lando/main/models/repo.py
zzzeid: I don't think type hints are necessary for this method. Both self type, and `None` return value are self-evident, and are also not present anywhere else in a similar context.

src/lando/main/management/commands/hg_landing_worker.py
```diff
@@ -15,6 +15,13 @@ class Command(BaseCommand, WorkerMixin):
     name = "hg-landing-worker"
 
     def handle(self, *args, **options):
+        # Clone or update repos upon worker startup.
+        for repo in self._instance.enabled_repos:
+            # Check if any associated repos are unsupported, raise exception if so.
+            repo.raise_if_not(repo.HG)
```
cgsheeh, line 21: This needs to be updated to reflect the name change in the other PR.

cgsheeh (changes requested):

cgsheeh (approved):

PR #109, opened 2024-08-27
repo: add hg interface (bug 1915147)

zzzeid:
- implement Repo.hg property
- add RepoError exception
- add _system_path property that calculates local repo path
- add raise_if_not method that asserts type of repo
- simplify LandingWorker.run_job to not require repo params
- update all tests to reflect new implementation
- set system_path on save
- comment out uplift-repo in helper script (repo does not exist)


src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -98,23 +94,14 @@ def loop(self):
             self.throttle(self.sleep_seconds)
             return
 
-        with job_processing(self, job):
+        with job_processing(job):
```
zzzeid, line 97: This will be fixed after rebasing.

src/lando/main/models/repo.py
```diff
@@ -72,20 +77,33 @@ class Repo(BaseModel):
     product_details_url = models.CharField(blank=True, default="")
     push_bookmark = models.CharField(blank=True, default="")
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.is_hg_repo:
+            self.hg = HgRepo(self.system_path or self._system_path)
```
cgsheeh, line 83: Rename one of these `system_path` variables to something else, so it's obvious what the difference is.

cgsheeh (changes requested):

src/lando/main/models/repo.py
```diff
@@ -72,20 +77,33 @@ class Repo(BaseModel):
...
 
+    @property
+    def _system_path(self):
+        return str(Path(settings.REPO_ROOT) / self.name)
+
+    @property
+    def _method_not_supported_for_repo_error(self):
+        return RepoError(f"Method is not supported for {self}")
+
+    def raise_if_not(self, repo_scm):
```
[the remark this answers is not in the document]
zzzeid: Yeah, it is used in #108, but it's more closely related to the "hg interface" than the repo cloning so I put it in here.

src/lando/main/models/repo.py
zzzeid: This naming convention for stored vs. calculated properties is somewhat conventional. The main reason this is added here is because `__init__` could be called before a repo instance is actually saved to the database, and thus `repo.system_path` will not be populated yet).

src/lando/main/models/repo.py
```diff
@@ -72,20 +79,34 @@ class Repo(BaseModel):
...
 
+    def get_system_path(self) -> str:
+        """Calculate system path based on REPO_ROOT and repository name."""
+        return str(Path(settings.REPO_ROOT) / self.name)
+
+    @property
+    def _method_not_supported_for_repo_error(self) -> RepoError:
+        return RepoError(f"Method is not supported for {self}")
+
+    def raise_for_unsupported_repo_scm(self, supported_scm):
```
cgsheeh, line 100: Type hint on `supported_scm`.

cgsheeh (changes requested):

PR #114, opened 2024-08-28
ui: port stack page and supporting functionality (bug 1876838)

zzzeid:
- fix jinja template styling
- temporarily comment out uplift related code
- implement changes to the view to make it work with new functionality
- update urls

### DIFF
--- a/src/lando/ui/jinja2/stack/stack.html
+++ b/src/lando/ui/jinja2/stack/stack.html
@@ -130,10 +130,10 @@
       </section>
       <footer class="modal-card-foot">
         <form class="StackPage-form" action="" method="post">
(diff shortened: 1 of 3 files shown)


src/lando/ui/jinja2/stack/stack.html
```diff
@@ -142,7 +142,7 @@ <h2>Landing is blocked:</h2>
           </button>
           <button class="StackPage-landingPreview-close button">Cancel</button>
         </form>
-        {% if user_has_phabricator_token %}
+        {# {% if user_has_phabricator_token %}
```
cgsheeh, line 145: Maybe we should add a `TODO` and reference to the bug.

PR #115, opened 2024-08-28
ui: fixes to navbar template (bug 1915446)

zzzeid:
- correctly fetch picture from user profile
- fix logout button functionality
- fix login button functionality
- use correct settings form attributes (not yet in use)
- update reference to phabricator_api_token


src/lando/ui/jinja2/partials/navbar.html
```diff
@@ -31,23 +31,25 @@
             <p class="control">
               <a class="Navbar-userSettingsBtn button">
                 <span class="icon">
-                  <img src="{{ session['userinfo']['picture']|avatar_url }}" />
+                  <img src="{{request.user.profile.userinfo['picture']|avatar_url }}" />
                 </span>
-                <span>&nbsp;{{ session['userinfo']['name'] }}&nbsp;</span>
+                <span>&nbsp;{{request.user.profile.userinfo['name']}}&nbsp;</span>
```
cgsheeh, line 36: Let's maintain spacing after/before the `{{`/`}}`, like `{{ request...` and then `name' }}`

cgsheeh (changes requested):

src/lando/ui/jinja2/partials/navbar.html
zzzeid: Ah nice catch, yep agreed.

PR #116, opened 2024-08-29
utils: phabricator client changes (bug 1915633)

zzzeid:
- add phabricator check to replace legacy health check
- check unprivileged and admin keys for validity
- remove phabricator subsystem
- move relevant functionality to lando.utils.phabricator
- update imports
- add get_phabricator_client method

### DIFF
--- a/src/lando/utils/apps.py
+++ b/src/lando/utils/apps.py
@@ -8,6 +8,10 @@ from lando.utils.phabricator import PhabricatorAPIException, PhabricatorClient
 
 PHAB_API_KEY_RE = re.compile(r"^api-.{28}$")
 
+PHAB_API_KEY_FORMAT_ERROR_TEMPLATE = (
+    '{} has the wrong format it must begin with "api-" and be 32 characters long.'
+)
+
 
 class UtilsConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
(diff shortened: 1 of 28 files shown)


src/lando/utils/apps.py
```diff
@@ -1,6 +1,46 @@
...
+PHAB_API_KEY_RE = re.compile(r"^api-.{28}$")
 
 
 class UtilsConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "lando.utils"
+
+
+@register()
+def phabricator_check(**kwargs):
```
cgsheeh, line 18: Return type hint here.

src/lando/utils/apps.py
```diff
@@ -1,6 +1,46 @@
...
+        errors.append(
+            "PHABRICATOR_ADMIN_API_KEY has the wrong format, "
+            'it must begin with "api-" and be 32 characters long.'
+        )
+
+    if (priv_key or unpriv_key) and not errors:
+        try:
+            PhabricatorClient(
+                settings.PHABRICATOR_URL,
+                settings.PHABRICATOR_UNPRIVILEGED_API_KEY,
```
cgsheeh, line 41: Based on the logic above, in theory `unpriv_key` could be empty and yet we still use `settings.PHABRICATOR_UNPRIVILEGED_API_KEY` unconditionally here. It seems we should only check whichever key is non-`None`, or maybe it's worth checking both to ensure one of the keys isn't expired etc?

cgsheeh (changes requested):

src/lando/utils/apps.py
zzzeid: Good suggestion, this way it is more robust that the legacy checks, and would also help with local setup.

cgsheeh (approved):

PR #117, opened 2024-08-30
commands: improve start_landing_worker command (bug 1915894)

zzzeid:
- rename command to start_landing_worker
- add name argument to specify which worker to run
- add scm field on Worker model
- update docker things


ahochheiden (approved):

ahochheiden:
Looks good. What's the plan for splitting workers between repos in the future? Are we planning on adding the repo names for a worker to handle as args for `start_landing_worker` as well?

zzzeid:
> Looks good. What's the plan for splitting workers between repos in the future? Are we planning on adding the repo names for a worker to handle as args for `start_landing_worker` as well?

Good question. Currently, you can set the values in the admin site per worker, however with the `scm` argument we are limited to one worker per scm type. I think maybe it makes sense to move the SCM setting to the worker itself, and pass the explicit name attribute to the start_landing_worker command.

ahochheiden:
Okay, so we'll have `start_landing_worker <worker_name>` and the `Worker` is created in the admin UI and associated with repos there? That sounds reasonable.

zzzeid:
> Okay, so we'll have `start_landing_worker <worker_name>` and the `Worker` is created in the admin UI and associated with repos there? That sounds reasonable.

This is done now, I also added the `scm` field to the worker. There's currently no enforcement of which repos can be associated with which worker, but that can be done as part of [bug 1917075](https://bugzilla.mozilla.org/show_bug.cgi?id=1917075).

ahochheiden (approved):

PR #121, opened 2024-09-06
environments: add Environemnt enum (bug 1917247)

zzzeid:
- add Environment enum with definitions for different environments
- replace existing strings with this new enum


src/lando/environments.py
```diff
@@ -0,0 +1,21 @@
+from enum import StrEnum, auto
+
+
+class Environment(StrEnum):
```
ahochheiden, line 4: Docstring here would be helpful.

ahochheiden (approved):

PR #125, opened 2024-09-11
sentry: port sentry support (bug 1873221)

zzzeid:


src/lando/main/sentry.py
```diff
@@ -0,0 +1,25 @@
...
+logger = logging.getLogger(__name__)
+
+
+def init_sentry():
+    """Initialize Sentry integration for remote environments."""
+    if not settings.ENVIRONMENT.is_remote:
+        raise ValueError(
+            "Attempting to initialize Sentry in non-remote environment "
+            f"({settings.ENVIRONMENT})"
+        )
```
cgsheeh, line 18: Nit: we also check this in the caller, it can be removed.

src/lando/main/sentry.py
```diff
@@ -0,0 +1,25 @@
...
+    sentry_sdk.init(
+        dsn=settings.SENTRY_DSN,
+        environment=settings.ENVIRONMENT,
+        integrations=[DjangoIntegration()],
+        traces_sample_rate=1.0,
+        release=version,
+    )
```
cgsheeh, line 25: In the previous implementation we have a log entry to indicate the status of Sentry, should we carry that forward?

cgsheeh (approved):

src/lando/main/sentry.py
zzzeid: I don't think that's necessary here, don't think I've ever looked for those logs. Easy enough to re-add if we ever feel the need to do that.

src/lando/main/sentry.py
zzzeid: Changed up the implementation slightly to simplify this.

cgsheeh (approved):

PR #126, opened 2024-09-11
logging: port legacy logging functionality (bug 1867613)

zzzeid:
- move legacy logging config to new platform
- remove unneeded loggers (flask, werkzeug)
- update django logger to match rest of the config
- convert field values to string to ensure JSON serializability
- only use custom logging on remote environments

### DIFF
--- a/src/lando/main/logging.py
+++ b/src/lando/main/logging.py
@@ -3,8 +3,6 @@ import logging
 import socket
 import traceback
 
-from lando.api.legacy.systems import Subsystem
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


cgsheeh (approved):

src/lando/main/logging.py
```diff
@@ -68,6 +66,9 @@ def __init__(self, *args, mozlog_logger=None, **kwargs):
     def format(self, record):
         """Formats a log record and serializes to mozlog json"""
 
+        # NOTE: Django passes some fields that are not JSON serializable in the record
+        # (for example, the WSGIRequest object representing the request). Therefore
+        # those values are converted to a string to avoid any issues when serializing.
```
zzzeid, line 71: As @cgsheeh pointed out, it may be nicer/cleaner in the future to use something like https://github.com/mozilla-services/python-dockerflow to cover this and other cases. It might also have additional improvements/requirements that aren't included in this legacy code.

PR #130, opened 2024-09-20
ui, profile: support user settings functionality (bug 1920172)

zzzeid:
- update forms
- rename various variables (from using "token" to using "key")
- add functionality to controller
- update template to reflect new functionality
- update test and enable


src/lando/urls.py
```diff
@@ -28,11 +29,16 @@
 urlpatterns += [
     path("__version__", DockerflowViews.VersionView.as_view(), name="version"),
     path("__heartbeat__", DockerflowViews.HeartbeatView.as_view(), name="heartbeat"),
-    path("__lbheartbeat__", DockerflowViews.LoadBalancerHeartbeatView.as_view(), name="lbheartbeat"),
+    path(
+        "__lbheartbeat__",
+        DockerflowViews.LoadBalancerHeartbeatView.as_view(),
+        name="lbheartbeat",
+    ),
```
cgsheeh, line 36: Code formatting problems? How are these still slipping past CI...?

src/lando/ui/legacy/usersettings.py
```diff
@@ -1,37 +1,27 @@
...
-            secure=current_app.config["USE_HTTPS"],
-            httponly=True,
-        )
-
-    return response
+    if not request.method == "POST":
+        return HttpResponseNotAllowed()
+
+    form = UserSettingsForm(request.POST)
+    if form.is_valid():
```
cgsheeh, line 18: The `else` clause only returns an error state, so you should reverse the `if` logic here to de-dent the main code path:

```python
if not form.is_valid():
    return JsonResponse({"errors": form.errors}, status=400)

profile = request.user.profile
...
```

cgsheeh (changes requested):

src/lando/urls.py
zzzeid: See [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1918236) bug.

src/lando/static_src/legacy/js/components/Navbar.js
```diff
@@ -101,31 +102,31 @@ $.fn.landoNavbar = function() {
...
       }
-      $phabAPITokenInput.prop('disabled', false);
-      if (!isSetPhabAPIToken) {
-        $phabAPITokenInput.prop('placeholder', 'not set');
+      $phabricatorAPIKeyInput.prop('disabled', false);
+      if (!isSetPhabricatorAPIKey) {
+        $phabricatorAPIKeyInput.prop('placeholder', 'not set');
       } else {
-        $phabAPITokenInput.prop('placeholder', 'api-############################');
+        $phabricatorAPIKeyInput.prop('placeholder', 'phabricator api key is set');
```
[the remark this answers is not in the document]
zzzeid: Sure. I think we should actually improve this UI in general, I don't particularly like that we are using the palceholder to communicate this message with the user. I'll file a bug for that.

src/lando/static_src/legacy/js/components/Navbar.js
```diff
@@ -48,22 +48,23 @@ $.fn.landoNavbar = function() {
...
         data: $(this).serialize(),
         dataType: 'json',
         success: data => {
           modalSubmitBtnOn();
+            console.log(data);
           if (!data.success) {
             return handlePhabAPITokenErrors(data.errors);
           }
-          isSetPhabAPIToken = data.phab_api_token_set;
+          isSetPhabricatorAPIKey = data.phab_api_token_set;
```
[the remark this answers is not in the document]
zzzeid: Looks like this is not needed, just a remnant from previous implementation.

zzzeid:
image
image
image

pyproject.toml
```diff
@@ -44,4 +44,5 @@ testpaths = [
     "src/lando/dockerflow",
     "src/lando/main/tests.py",
     "src/lando/tests",
+    "src/lando/ui/tests/test_forms.py",
```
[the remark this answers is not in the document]
zzzeid: This is the only test module in that directory that should be enabled as part of this PR. There are other obsolete or irrelevant tests that are still in that directory, or perhaps ones that should be ported that are not yet.

I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1922318) to track this work.

src/lando/ui/legacy/forms.py
```diff
@@ -19,5 +19,10 @@ class UpliftRequestForm(forms.Form):
...
     """Form used to provide the Phabricator API Token."""
 
-    phab_api_token = forms.RegexField(required=False, regex="^api-[a-z0-9]{28}$")
-    reset_phab_api_token = forms.BooleanField(required=False)
+    phabricator_api_key = forms.RegexField(
+        required=False,
+        regex="^api-[a-z0-9]{28}$",
+        label="Phabricator API Key",
+    )
+    phabricator_api_key.widget.attrs.update({"class": "input"})
```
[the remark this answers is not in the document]
zzzeid: This is because the form element is actually generated by Django. This is the conventional way to modify form elements in Django if using the builtin forms/widgets. I think in the long run we will likely need to move away from this, but this works well at this time since we don't have much customization.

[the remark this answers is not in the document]
cgsheeh: In the Flask world we can do things like [this](https://github.com/mozilla-conduit/lando-ui/blob/main/landoui/templates/treestatus/log.html#L43) to specify attributes in the HTML. If we can't do that in Django that's fine, just a little surprising. :sweat_smile:

pyproject.toml
cgsheeh, line 47: Please add a `TODO` comment linking to the bug here.

cgsheeh (changes requested):

pyproject.toml
zzzeid: What value will that add? We are already tracking this bug on bugzilla and it doesn't affect functionality. On line 45 we also have a single module as a test file, so this is not something unusual either.

cgsheeh: Leaving a comment in the code that points to the bug provides immediate context to future developers or reviewers, making it clear why certain tests are not enabled. It helps maintain transparency, ensures the issue isn’t forgotten, and avoids confusion when someone revisits the code later.

zzzeid: I don't agree that this is necessary in this case, but I agree that it can be, in some cases.

I think in most cases we should not be referencing bugs in the code. Bugs are mutable after all, and you can't know if a bug is referenced in the code base before you change it.

zzzeid: Regardless I will add it to this PR since this seems important to you at this stage.

cgsheeh (approved):

PR #131, opened 2024-09-23
auth: use SessionRefresh middleware (bug 1899399)

zzzeid:
- add SessionRefresh middleware
- implement decorator to force expiry of auth token
- add decorator to landing job endpoint


src/lando/main/auth.py
```diff
@@ -73,6 +77,30 @@ def wrapper(request, *args, **kwargs):
...
+def force_auth_refresh(f: Callable) -> Callable:
+    """
+    Decorator which forces authenticated session to be refreshed.
+    """
+
+    def wrapper(*args, **kwargs):
+        """Set oidc_id_token_expiration to 0, forcing session refresh."""
+        # First check that SessionRefresh is indeed enabled.
+        if "mozilla_django_oidc.middleware.SessionRefresh" not in settings.MIDDLEWARE:
+            raise RuntimeError("SessionRefresh middleware required but is not enabled.")
```
cgsheeh, line 89: Question: could we check this outside the wrapper, either in app setup or outside of `wrapper` at the start of `force_auth_refresh`? Seems this check will run on every call, when it's really a detail that we could assert once and assume elsewhere in the app.

src/lando/main/auth.py
```diff
@@ -73,6 +77,30 @@ def wrapper(request, *args, **kwargs):
...
+
+        # Find the request in the arguments. This is needed in case this decorator
+        # is used in class-based views. The request would be the first argument that
+        # is a WSGIRequest instance.
+        request = [arg for arg in args if isinstance(arg, WSGIRequest)][0]
```
cgsheeh, line 94: Is the arg not in the same position for different views?

cgsheeh (approved):

src/lando/main/auth.py
zzzeid: Django has both class based views, and method based views. This ensures things work for both types of views. For the former, the request is in the second argument.

src/lando/main/auth.py
zzzeid: It might make sense to add a separate (app) check, to ensure that SessionRefresh middleware is enabled. However this is meant here as an integrated fail-safe, since there is a security component to this decorator and functionally it would result in insecure behaviour if it is used without SessionRefresh middleware being enabled.

I think it is cleaner to keep all functionality in the wrapper, which is pretty conventional for wrappers like this.

PR #134, opened 2024-09-24
sentry: only load version information for remote interaction (bug 1920808)

shtrom:
This fixes a race where the version file generator cannot run because the version file is missing.


cgsheeh (approved):

zzzeid (approved):
Would be better to shorten the commit message so it is not truncated.

shtrom:
Ah, that's the PR title. Yes.

PR #135, opened 2024-09-27
settings: update lint paths to be more comprehensive (bug 1918236)

zzzeid:
- update LINT_PATHS
- run `lando format`

### DIFF
--- a/src/lando/settings.py
+++ b/src/lando/settings.py
@@ -12,9 +12,9 @@ https://docs.djangoproject.com/en/dev/ref/settings/
 
 import os
 from pathlib import Path
+
 from lando.environments import Environment
 from lando.main.logging import MozLogFormatter
-
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = Path(__file__).resolve().parent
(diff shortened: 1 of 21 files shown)


src/lando/settings.py
```diff
@@ -198,7 +196,10 @@
     "lando.main.auth.LandoOIDCAuthenticationBackend",
 ]
 
-LINT_PATHS = tuple(f"{BASE_DIR}/{path}" for path in ("api", "dockerflow", "main", "utils", "tests"))
+LINT_PATHS = tuple(
+    f"{BASE_DIR}/{path}"
+    for path in (".", "ui", "api", "dockerflow", "main", "utils", "tests")
+)
```
cgsheeh, line 202: Any reason we couldn't just run `black` on all files by default? Maybe it would make more sense to maintain a list of exclude paths in the `tool.black` section of `pyproject.toml` instead of maintaining a list of include paths?

src/lando/ui/legacy/api.py
```diff
@@ -40,15 +39,16 @@ def post(self, request):
...
+            )
 
         if not uplift_request_form.validate():
             errors = list(chain(*uplift_request_form.errors.values()))
             return JsonResponse({"errors": errors}, status_code=400)
 
-        revision_id = uplift_request_form.revision_id.data
-        repository = uplift_request_form.repository.data
+        # revision_id = uplift_request_form.revision_id.data
+        # repository = uplift_request_form.repository.data
```
cgsheeh, line 51: Unrelated?

cgsheeh (changes requested):
r- because of the commented uplift lines, otherwise this is good to go.

src/lando/ui/legacy/api.py
zzzeid: They are unused variables that popped up in the lint check. There's additional commented out code below that is connected to these variables as well.

src/lando/settings.py
zzzeid: These changes affect both `black` and `ruff` by the way.

I guess this is a remnant of the legacy implementation (we've always explicitly set the paths this way) though this is less relevant (or completely irrelevant perhaps) now since all relevant files are under `src/lando`.

In the future, we might want to add additional paths to this that are outside `src/lando`, so maybe this should just be changed to 
```suggestion
LINT_PATHS = (f"{BASE_DIR}/{path}", )
```

cgsheeh (approved):

PR #136, opened 2024-10-01
lando polishes (bug 1898562) (bug 1922274)

shtrom:


shtrom:
@zzzeid

zzzeid (changes requested):
Looks good, but the commit message / bug numbers could be modified to align better with existing conventions in this repo. Also the branch name is `wip`, confirming this was intended to be out of WIP state at this point?

E.g.,

```
docker: various fixes

- create /files/repos directory, update permissions (bug 1922274)
- update readme (bug 1898562)

shtrom:
Yeah, I should have thought a bit more when naming that branch, but I wasn't sure what was going into it, if anything.

I've rebased and squashed (though I think it would have been feasible in the UI), but I don't think it's worth renaming the branch. Happy to do so if you think it is.

Otherwise, this should now be good to go.

PR #139, opened 2024-10-24
auth: filter userinfo before storing (bug 1926950)

zzzeid:
- remove info that is not used by Lando
- add test

### DIFF
--- a/src/lando/main/models/profile.py
+++ b/src/lando/main/models/profile.py
@@ -35,6 +35,23 @@ SCM_NSS = SCM_PERMISSIONS_MAP["SCM_NSS"]
 SCM_VERSIONCONTROL = SCM_PERMISSIONS_MAP["SCM_VERSIONCONTROL"]
 
 
+def filter_claims(claims: dict) -> dict:
+    """Return only necessary info in the provided dict."""
+    keep_keys = ("email", "picture", "email_verified", CLAIM_GROUPS_KEY)
+
+    # Remove keys that are not present in keep_keys.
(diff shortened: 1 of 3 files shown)


src/lando/main/models/profile.py
```diff
@@ -35,6 +35,23 @@
...
+
+    # Remove keys that are not present in keep_keys.
+    claims = {key: value for key, value in claims.items() if key in keep_keys}
+
+    # Remove reference to any groups not currently used in Lando.
+    # NOTE: currently these are SCM group, however in the future other
+    # groups will need to be added here, for example "treestatus users",
+    # and other lando permissions.
+    claims[CLAIM_GROUPS_KEY] = [
+        group for group in claims[CLAIM_GROUPS_KEY] if "scm" in group.lower()
```
cgsheeh, line 50: Nit: could we check if the claims group matches something in `SCM_PERMISSIONS`? That way when we add the new permissions we will get support for them in this filter automatically.

cgsheeh (approved):

src/lando/main/models/profile.py
zzzeid: The strings in `SCM_PERMISSIONS` form the base for the various groups, so we would have to check that those strings are part of any of the returned groups, not so much directly checking membership.

I think this could be improved, but leaning towards leaving it as it is for now. Depending on how we decide to handle [bug 1908702](https://bugzilla.mozilla.org/show_bug.cgi?id=1908702), we may be able to discard the groups altogether.

PR #140, opened 2024-10-25
profile: userinfo filtering fixes

zzzeid:
- check for existence of CLAIM_GROUPS_KEY (bug 1927153)
- exclude "name" from filtered fields (bug 1927152)

### DIFF
--- a/src/lando/main/models/profile.py
+++ b/src/lando/main/models/profile.py
@@ -37,7 +37,13 @@ SCM_VERSIONCONTROL = SCM_PERMISSIONS_MAP["SCM_VERSIONCONTROL"]
 
 def filter_claims(claims: dict) -> dict:
     """Return only necessary info in the provided dict."""
-    keep_keys = ("email", "picture", "email_verified", "name", CLAIM_GROUPS_KEY)
+    keep_keys = (
+        "email",
+        "email_verified",
+        "name",
+        "picture",
+        CLAIM_GROUPS_KEY,
+    )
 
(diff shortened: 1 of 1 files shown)


src/lando/main/models/profile.py
```diff
@@ -46,9 +46,11 @@ def filter_claims(claims: dict) -> dict:
     # NOTE: currently these are SCM group, however in the future other
     # groups will need to be added here, for example "treestatus users",
     # and other lando permissions.
-    claims[CLAIM_GROUPS_KEY] = [
-        group for group in claims[CLAIM_GROUPS_KEY] if "scm" in group.lower()
-    ]
+    # This filter is not applicable to local development.
+    if CLAIM_GROUPS_KEY in claims:
```
cgsheeh, line 50: Maybe we should check for `settings.ENVIRONMENT == Environment.local` here instead, since the comment suggests that's the actual condition we want to check.

src/lando/main/models/profile.py
```diff
@@ -37,7 +37,7 @@
 
 def filter_claims(claims: dict) -> dict:
     """Return only necessary info in the provided dict."""
-    keep_keys = ("email", "picture", "email_verified", CLAIM_GROUPS_KEY)
+    keep_keys = ("email", "picture", "email_verified", "name", CLAIM_GROUPS_KEY)
```
cgsheeh, line 40: Nit: these could be sorted so it's slightly more readable.

cgsheeh (approved):

PR #141, opened 2024-10-25
middleware: add response headers middleware (bug 1927148)

zzzeid:
- port app wide headers hook (from legacy UI and API)
- re-enable previously skipped test (bug 1926964)
- remove some legacy code
- add failing test to account for future functionality

### DIFF
--- a/src/lando/middleware.py
+++ b/src/lando/middleware.py
@@ -1,3 +1,5 @@
+from collections.abc import Callable
+
 from django.conf import settings
 from django.http import HttpRequest, HttpResponse
 
@@ -5,7 +7,7 @@ from django.http import HttpRequest, HttpResponse
 class ResponseHeadersMiddleware:
     """Add custom response headers for each request."""
 
-    def __init__(self, get_response: callable):
+    def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]):
         self.get_response = get_response
 
     def __call__(self, request: HttpRequest) -> HttpResponse:
--- a/src/lando/middleware.py
+++ b/src/lando/middleware.py
@@ -21,16 +21,16 @@ class ResponseHeadersMiddleware:
 
         csp = [
             "default-src 'self'",
-            "font-src 'self' https://code.cdn.mozilla.net",
-            "style-src 'self' https://code.cdn.mozilla.net",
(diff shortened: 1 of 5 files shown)


src/lando/middleware.py
```diff
@@ -0,0 +1,41 @@
+from django.conf import settings
+from django.http import HttpRequest, HttpResponse
+
+
+class ResponseHeadersMiddleware:
+    """Add custom response headers for each request."""
+
+    def __init__(self, get_response: callable):
```
cgsheeh, line 8: Type hints for callable functions don't support the `callable` builtin function, since it is just a function and not an actual type like `list`/`dict`/`int` etc. You need to use `collections.abc.Callable` instead.

Nit: you could also add the parameter and return types since the `__call__` function below implies they are known. 
```suggestion
    def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]):
```

src/lando/middleware.py
```diff
@@ -0,0 +1,41 @@
...
+            "font-src 'self' https://code.cdn.mozilla.net",
+            "style-src 'self' https://code.cdn.mozilla.net",
+            "img-src 'self' *.cloudfront.net *.gravatar.com *.googleusercontent.com",
+            "object-src 'none'",
+            "frame-ancestors 'none'",
+            "manifest-src 'none'",
+            "worker-src 'none'",
+            "media-src 'none'",
+            "frame-src 'none'",
+            "base-uri 'none'",
```
cgsheeh, line 31: Nit: we could sort these for readability.

cgsheeh (changes requested):
Looks great, we just need to fix `callable` -> `Callable` and then this is good to go.

src/lando/middleware.py
zzzeid: I considered that, however, there is an implication in the specs that the order is preserved (although it is not respected), however that is not explicitly stated in the specs, so I decided to keep those as they appeared in the [original code](https://github.com/mozilla-conduit/lando-ui/blob/main/landoui/app.py#L63).

cgsheeh (approved):

src/lando/middleware.py
cgsheeh: Works for me. :)

PR #143, opened 2024-10-29
landing_jobs: fix job cancelling (bug 1915712)

shtrom:


src/lando/static_src/legacy/js/components/Timeline.js
```diff
@@ -1,50 +1,38 @@
-'use strict';
+'e strict';
```
zzzeid, line 1: Was this clipped by accident?

src/lando/static_src/legacy/js/components/Timeline.js
```diff
@@ -1,50 +1,38 @@
...
-     * any of the parent's siblings. For example:
-     * <div>
-     *  <div>Some content <a href="#" class="toggle-content">toggle</a></div>
-     *  <div>Other content <a href="#" class="toggle-content">toggle</a></div>
-     * </div>
-    */
-    var link = $(this);
-    link.parent().hide();
-    link.parent().siblings().show()
-});
```
zzzeid, line 50: I think this was deleted, possibly by error? (may have been from the original WIP patch). This handles the merge conflict collapse buttons, and the functionality is still needed. You can trigger this behaviour by creating a situation with a merge conflict.

zzzeid (changes requested):

src/lando/static_src/legacy/js/components/Timeline.js
shtrom: Looks like it... And it still works. Oh Javascript...

src/lando/static_src/legacy/js/components/Timeline.js
shtrom: Ah, yep, that deletion was in the patch. I'll restore it and test.

src/lando/static_src/legacy/js/components/Timeline.js
```diff
@@ -1,50 +1,51 @@
...
+    return this.each(function() {
```
zzzeid, line 45: I just noticed that in other JavaScript files we put all the code after this line, but this was moved to be outside of this block. Might have been a relic of the legacy implementation of these methods, but probably wouldn't hurt to continue following the general conventions as they are elsewhere.

src/lando/utils/management/commands/tests.py
```diff
@@ -22,9 +22,17 @@ def add_arguments(self, parser):
...
+            "-k", type=str, help="Only run tests which match the given substring."
+        )
+
     def handle(self, *args, **options):
         command = ["pytest"]
 
+        if options["k"]:
+            command.append("-k")
+            command.append(options["k"])
+
```
zzzeid, line 35: This also is out of scope for this PR, and ideally would be moved to its own PR.

zzzeid (changes requested):

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -277,7 +277,7 @@ def run_job(self, job: LandingJob) -> bool:
                     repo.hg.apply_patch(patch_buf)
                 except PatchConflict as exc:
                     breakdown = self.process_merge_conflict(
-                        exc, repo, repo.hg, revision.revision_id
+                        exc, repo, revision.revision_id
```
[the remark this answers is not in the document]
shtrom: #151

src/lando/utils/management/commands/tests.py
shtrom: #150

src/lando/static_src/legacy/js/components/Timeline.js
shtrom: Yeah, you're right. This could have been a bit racy otherwise, but the way the main.js works is that it load and executes all the code when the document is ready https://github.com/shtrom/lando/blob/d94b2fcea886ce9387a30d8f68009902533d477a/src/lando/static_src/legacy/js/main.js#L9-L17, so it didn't cause a noticeable issue.

shtrom:
Ok, this should be good to go.

Manual tests:
1. `Preview Landing`, `Land to first-repo`, then `Cancel landing request` -> cancelled ✅ 
2. Ditto, but wait 10+s (so the landing succeed) before hitting `Cancel landing request` -> Orange `Could not cancel landing request` ✅ 
3. Create a conflicting landing as per [bug 1928565)[https://bugzilla.mozilla.org/show_bug.cgi?id=1928565] (using #151), then view the stack in Lando
3.1 `expand diff` -> expands diff ✅ 
3.2  `Show raw error output` -> show ✅

cgsheeh:
@shtrom looks like we have tests for this that are marked as skipped in `src/lando/api/tests/test_landing_job.py`. Could you fix those so they are no longer skipped?

zzzeid:
> @shtrom looks like we have tests for this that are marked as skipped in `src/lando/api/tests/test_landing_job.py`. Could you fix those so they are no longer skipped?

+1. The tests should require minimal changes (outside of adding an authenticated user and any missing attributes to the user or landing job). These "ported" tests would use the actual client rather than the proxy client. I would suggest adding an "authenticated user client" fixture so that you don't have to repeat the code, you could also pass it the permissions needed upon creation instead of adding all permissions. Something like this should do it, e.g., for `test_cancel_landing_job_cancels_when_submitted`:

```diff
-@pytest.mark.skip                                                                                                                                                                                                    
 def test_cancel_landing_job_cancels_when_submitted(                                                                                                                                                                  
     db, client, landing_job, mock_permissions                                                                                                                                                                        
 ):                                                                                                                                                                                                                   
     """Test happy path; cancelling a job that has not started yet."""
-    job = landing_job(LandingJobStatus.SUBMITTED)
+    user = User.objects.create_user(username="test_user", password="test_password", email="testuser@example.org")
+    user.profile = Profile(user=user)
+    for permission in user.profile.get_all_scm_permissions():
+        user.user_permissions.add(permission)
+    client.login(username="test_user", password="test_password")
+    job = landing_job(LandingJobStatus.SUBMITTED, requester_email=user.email)
     response = client.put(
         f"/landing_jobs/{job.id}",
-        json={"status": LandingJobStatus.CANCELLED.value},
-        permissions=mock_permissions,
+        json.dumps({"status": LandingJobStatus.CANCELLED.value}),
     )
  
     assert response.status_code == 200
-    assert response.json["id"] == job.id
+    assert response.json()["id"] == job.id
+    job.refresh_from_db()
     assert job.status == LandingJobStatus.CANCELLED

```

zzzeid:
> > @shtrom looks like we have tests for this that are marked as skipped in `src/lando/api/tests/test_landing_job.py`. Could you fix those so they are no longer skipped?
> 
> +1. The tests should require minimal changes (outside of adding an authenticated user and any missing attributes to the user or landing job). These "ported" tests would use the actual client rather than the proxy client. I would suggest adding an "authenticated user client" fixture so that you don't have to repeat the code, you could also pass it the permissions needed upon creation instead of adding all permissions. Something like this should do it, e.g., for `test_cancel_landing_job_cancels_when_submitted`:
> 
> ```diff
> -@pytest.mark.skip                                                                                                                                                                                                    
>  def test_cancel_landing_job_cancels_when_submitted(                                                                                                                                                                  
>      db, client, landing_job, mock_permissions                                                                                                                                                                        
>  ):                                                                                                                                                                                                                   
>      """Test happy path; cancelling a job that has not started yet."""
> -    job = landing_job(LandingJobStatus.SUBMITTED)
> +    user = User.objects.create_user(username="test_user", password="test_password", email="testuser@example.org")
> +    user.profile = Profile(user=user)
> +    for permission in user.profile.get_all_scm_permissions():
> +        user.user_permissions.add(permission)
> +    client.login(username="test_user", password="test_password")
> +    job = landing_job(LandingJobStatus.SUBMITTED, requester_email=user.email)
>      response = client.put(
>          f"/landing_jobs/{job.id}",
> -        json={"status": LandingJobStatus.CANCELLED.value},
> -        permissions=mock_permissions,
> +        json.dumps({"status": LandingJobStatus.CANCELLED.value}),
>      )
>   
>      assert response.status_code == 200
> -    assert response.json["id"] == job.id
> +    assert response.json()["id"] == job.id
> +    job.refresh_from_db()
>      assert job.status == LandingJobStatus.CANCELLED
> ```

Another noteworthy thing here is that this is a good case study related to [bug 1897443](https://bugzilla.mozilla.org/show_bug.cgi?id=1897443). The original tests you updated in this PR use the same code path as this test, though they reach it via different means. The fake request/fake user here is replaced with a "real" user in the test client, and "real" permissions.

src/lando/api/tests/conftest.py
```diff
@@ -481,6 +481,10 @@ def __init__(self, is_authenticated=True, has_email=True, permissions=None):
 
     class FakeRequest:
         def __init__(self, *args, **kwargs):
+            self.body = "{}"
+            if "body" in kwargs:
+                self.body = kwargs["body"]
+                del kwargs["body"]
```
zzzeid, line 487: nit
```suggestion
                self.body = kwargs.pop("body")
```

src/lando/urls.py
```diff
@@ -42,3 +43,8 @@
     path("D<int:revision_id>/", revisions.Revision.as_view(), name="revisions-page"),
     path("manage_api_key/", user_settings.manage_api_key, name="user-settings"),
 ]
+
+# "API" endpoints ported from legacy API app.
+urlpatterns += [
+    path("landing_jobs/<int:landing_job_id>", landing_jobs.put, name="landing-jobs"),
```
zzzeid, line 49: As a matter of convention, there should be a trailing `/` here.

zzzeid (dismissed):
See other comments regarding un-skipping the tests for this.

shtrom:
> @shtrom looks like we have tests for this that are marked as skipped in `src/lando/api/tests/test_landing_job.py`. Could you fix those so they are no longer skipped?

I fixed some that just needed test-code updates. There are a couple that need other code changes, so I moved that to bug [1929466](https://bugzilla.mozilla.org/show_bug.cgi?id=1929466) (#152).

src/lando/api/legacy/api/landing_jobs.py
```diff
@@ -27,31 +39,46 @@ def put(request: HttpRequest, landing_job_id: str, data: dict):
...
+        return JsonResponse(data, status=400)
+
+    landing_job_id = form.cleaned_data["landing_job_id"]
+    status = form.cleaned_data["status"]
+
+    with LandingJob.lock_table:
+        try:
+            landing_job = LandingJob.objects.get(pk=landing_job_id)
+        except LandingJob.DoesNotExist:
+            raise Http404(f"A landing job with ID {landing_job_id} was not found.")
```
zzzeid, line 62: As discussed, we can either change this to a JsonResponse with a custom status code to make tests pass, or the tests can ignore the content of the response and just check the status code as a temporary solution. Though this can all be fixed as part of [bug 1929466](https://bugzilla.mozilla.org/show_bug.cgi?id=1929466).

zzzeid (approved):
lgtm!

src/lando/api/legacy/api/landing_jobs.py
shtrom: Yup, let's postpone to the other bug https://bugzilla.mozilla.org/show_bug.cgi?id=1929466#c2

PR #145, opened 2024-10-30
middleware: implement maintenance mode middleware (bug 1927149)

zzzeid:
- remove legacy hook code
- add `db` fixture to some tests
- move dockerflow urls to own namespaced module
- add ConfigurationVariable to admin UI
- return value of variables in `get` method instead of returning full object
- add maintenance page template
- add maintenance mode middleware, filter out excluded namespaces, urls
- hide login/logout buttons when in maintenance mode
- add maintenance command
- option to check for pending migrations before enabling maintenance mode


src/lando/utils/management/commands/maintenance.py
```diff
@@ -0,0 +1,63 @@
...
+                f"Action must be one of: {', '.join(action_mapping.keys())}. "
+                f'"{action}" was provided.'
+            )
+
+        unapplied_migrations = (
+            subprocess.run(["lando", "migrate", "--check", "--noinput"]).returncode != 0
+        )
+
+        if force or unapplied_migrations:
+            action_mapping[action]()
```
shtrom, line 63: I wonder if the migrations should be a separate command, that toggles the maintenance mode as needed. This way we could turn on maintenance mode for any other reason where we might not want to add the weight of migrations.

shtrom: Alternatively, we could provide a flag such as `--skip-migration` to achieve the same thing.

src/lando/api/tests/test_hooks.py
```diff
@@ -10,7 +10,7 @@
 )
 
 
-def test_app_wide_headers_set(client):
+def test_app_wide_headers_set(db, client):
```
shtrom, line 13: Where does `db` come from? Is it a fixture? I can't find its definition (not that it impacts the tests).

zzzeid: Good point. This is actually a fixture [provided](https://pytest-django.readthedocs.io/en/latest/helpers.html#db) by pytest-django, though after going over the documentation again, it seems it is intended to be used in other fixtures, and not as a shortcut/alternative to marking the test with `django_db`.

Though it works as expected, I think it's better to explicitly mark those tests as needing the db. I'll file a bug to fix other occurrences in the future, but will change these in this PR.

src/lando/utils/management/commands/maintenance.py
zzzeid: That was kind of the intention of the `force` parameter, though I do see your point. Maybe `maintenance` should modify the flags regardless, and instead we can create `pre-deploy` and `post-deploy` commands which can do the migration check instead.

src/lando/urls.py
```diff
@@ -19,25 +19,14 @@
...
 
 from lando.api.legacy.api import landing_jobs
-from lando.dockerflow import views as DockerflowViews
 from lando.ui.legacy import pages, revisions, user_settings
 
 urlpatterns = [
-    path("admin/", admin.site.urls),
+    path("", include("lando.dockerflow.urls", "dockerflow")),
     path("oidc/", include("mozilla_django_oidc.urls")),
+    path("admin/", admin.site.urls),
```
shtrom, line 27: Might be worth keeping those sorted.

shtrom (approved):
LGTM.

PR #147, opened 2024-10-31
landing_worker: add and use git scm support (bug 1895523)

shtrom:
Builds on top of the refactor for #154.


src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -129,6 +129,7 @@ def notify_user_of_landing_failure(job: LandingJob):
             job.requester_email, job.landing_job_identifier, job.error, job.id
         )
 
+    # XXX Not covered by tests
```
zzzeid, line 132: Should probably file a bug for this.

src/lando/api/tests/conftest.py
```diff
@@ -432,6 +432,28 @@ def hg_clone(hg_server, tmpdir):
     return clone_dir
 
 
+@pytest.fixture
+def git_repo(tmpdir):
```
zzzeid, line 436: Would be nice to add a docstring here. It might be also more useful to add a path parameter to this fixture, so that you can pass it elsewhere as an optional.

src/lando/api/tests/conftest.py
```diff
@@ -432,6 +432,28 @@ def hg_clone(hg_server, tmpdir):
...
+    subprocess.run(["git", "init", repo_dir.strpath], check=True)
+    file = repo_dir.join("first")
+    file.write("first file!")
+    git_setup_user(repo_dir)
+    subprocess.run(["git", "add", file.basename], check=True, cwd=repo_dir)
+    subprocess.run(["git", "commit", "-m", "first commit"], check=True, cwd=repo_dir)
+    return repo_dir
+
+
+def git_setup_user(repo_dir):
```
zzzeid, line 447: This should be a fixture. At least one that returns this method in its current form.

src/lando/api/tests/test_git.py
```diff
@@ -0,0 +1,81 @@
+import subprocess
+
+import pytest
+
+from lando.api.tests.conftest import git_setup_user
+from lando.main.scm.git import GitSCM
+
+
+# We can't use fixtures directly in parametrize [0],
+# so we jumps through some hoops.
```
zzzeid, line 10:
```suggestion
# so we jump through some hoops.
```

src/lando/api/tests/test_git.py
```diff
@@ -0,0 +1,81 @@
...
+        ("/non-existent-path", None, False),
+        ("/tmp", None, False),
+        (None, "git_repo", True),
+    ),
+)
+def test_is_initialised(path, repo_fixture_name, expected, request):
+    if not path:
+        path = request.getfixturevalue(repo_fixture_name).strpath
+    scm = GitSCM(path)
+    assert scm.repo_is_initialized == expected
```
zzzeid, line 24: This test is hard to read. I wrote an alternate suggestion here...

```suggestion
@pytest.mark.parametrize(
    "path,repo_fixture_name,expected",
    (
        ("/non-existent-path", False),
        ("/tmp", False),
        (None, True),  # generate a default path 
    ),
)
def test_is_initialised(path, expected, request, git_repo):
    if not path:
        path = git_repo.strpath
    scm = GitSCM(path)
    assert scm.repo_is_initialized == expected
```

Though ideally, this test would be rewritten to more explicitly show a git repo (along with a path) being initialized before it is checked (instead of just defaulting to a `None` path.)

src/lando/api/tests/test_git.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_clone(tmpdir, git_repo):
```
zzzeid, line 27: This test could be a bit more robust. At least checking whether the correct `git` command was called, even with a mock.

src/lando/api/tests/test_git.py
```diff
@@ -0,0 +1,81 @@
...
+        [
+            "git",
+            "commit",
+            "-m",
+            "adding new_file",
+            "--author",
+            "Lando <Lando@example.com>",
+        ],
+        cwd=clone_path.strpath,
+    )
```
zzzeid, line 64: Couldn't we use `GitSCM._git_call` here?

src/lando/main/models/repo.py
```diff
@@ -120,6 +118,8 @@ def get_scm(self) -> AbstractSCM:
         if not self._scm:
             if self.scm == SCM_HG:
                 self._scm = HgSCM(self.path)
+            elif self.scm == SCM_GIT:
+                self._scm = GitSCM(self.path, self.default_branch)
```
zzzeid, line 122: Left some comments in the other PR (#154) that should impact this implementation

src/lando/main/models/repo.py
```diff
@@ -172,24 +172,8 @@ def _find_supporting_scm(self, pull_path) -> str:
         for scm, impl in SCM_IMPLEMENTATIONS.items():
             if impl.repo_is_supported(pull_path):
                 return scm
-        if self._is_git_repo:
-            return SCM_GIT
```
zzzeid, line 176: Was this here in error originally?

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        return Path(self.path).exists() and self._git_call("status", cwd=self.path)
+
+    @classmethod
+    def repo_is_supported(self, path: str) -> bool:
+        """Determine wether the target repository is supported by this concrete implementation"""
+        # This only tests a remote target without any local filesystem interaction, the CWD doesn't matter.
+        return self._git_call("ls-remote", path, cwd="/tmp")
+
+    def _git_initialize(self):
+        self.refresh_from_db()
```
zzzeid, line 192: This is weird here... Was this a copy/paste error? `refresh_from_db` generally applies to a model object (or if used as a mixin).

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if self.is_initialized:
+            raise
+
+        result = self._git_run("clone", self.pull_path, self.name)
+        if result.returncode == 0:
+            self.is_initialized = True
+            self.save()
```
zzzeid, line 200: Same problem here. This is not a model as far as I can tell?

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+            A success indicator
+
+        """
+        command = ["git"] + list(args)
+        logger.debug("Calling " + " ".join(command) + " in " + cwd)
+        returncode = subprocess.call(
+            command,
+            cwd=cwd,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
```
zzzeid, line 289: Why suppress the output? Might be useful to keep some of this in the logs, at least for debugging.

zzzeid (changes requested):
First pass. I do plan on doing a second pass on this as well.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        yield self
+
+    @contextmanager
+    def for_push(self, requester_email: str) -> ContextManager:
+        """Context manager to prepare the repo with the correct environment variables set for pushing.
+
+        Args:
+            requester_email (str)
+        """
+        # We set the committer name to the requester's _email_ as this is the only piece of information about the user that we are comfortable making public. Names in the User objects are coming from LDAP, and may not be accepblate to use publicly.
```
cgsheeh, line 113: Split up this comment.

Also, `accepblate` -> `acceptable`.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        Returns:
+            str: The target changeset
+        """
+        branch = target_cset or self.default_branch
+        self.clean_repo()
+        self._git_run("pull", "--prune", pull_path)
+        self._git_run("checkout", "--force", "-B", branch)
+
+    def clean_repo(self, *, strip_non_public_commits=True):
+        """Reset the local repository to the origin"""
```
cgsheeh, line 157:
```suggestion

    def clean_repo(self, *, strip_non_public_commits: bool = True):
        """Reset the local repository to the origin."""
```

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        if strip_non_public_commits:
+            self._git_run("reset", "--hard", "origin/HEAD")
```
cgsheeh, line 159: Is using `origin/HEAD` correct here? My thought is we would probably want to use `origin/<branch>` for whichever branch we're working on (`origin/beta`, `origin/central` etc). Is there any guarantee that `origin/HEAD` is the correct commit to reset to?

As an alternative, did you consider creating a temp branch to work with locally while we apply patches? For example we create a `worker_D12345` branch and apply patches on it, then to rollback we just `git branch -D worker_D12345`, or `git branch autoland HEAD -f` to move the branch to the final destination on push.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+                "--author",
+                commit_author,
+                "--file",
+                f_msg.name,
+            )
+
+    def _git_last_commit_id(self) -> str:
+        return self._git_run("rev-parse", "HEAD", cwd=self.path)
+
+    def _git_run(self, *args, cwd: Optional[str] = None, must_succeed: bool = True):
```
cgsheeh, line 234: Return type is wrong in the docstring and missing from the function signature.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        if must_succeed and result.returncode:
+            raise SCMException(
+                f"Error running git command; {command=}, {path=}, {result.stderr}",
+                result.stdout,
+                result.stderr,
+            )
+        return result.stdout.strip()
+
+    @staticmethod
+    def _git_call(*args, cwd: str) -> bool:
```
cgsheeh, line 265: IMO we should only implement `git_run` and remove `git_call`, and always return the full underyling `CompletedProcess`. We can check the return code as needed in the callers if we need to assert success/failure. The `subprocess.run` API is just a better version of `subprocess.call`, and always having access to the `CompletedProcess` would give us the most flexibility in handling command output. Generally I dislike less-flexible wrappers around APIs that are already well designed.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+    ):
+        """Apply the given patch to the current repository
+
+        Args:
+            patch_buf (StringIO): The patch to apply
+
+        Returns:
+            None
+        """
+        self._git_apply_patch(diff, commit_description, commit_author, commit_date)
```
cgsheeh, line 99: Since `apply_patch` just calls `_git_apply_patch` with the same function signature, we can remove `_git_apply_patch`.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        """Return the commit_id of the tip of the current branch"""
+        return self._git_run("rev-parse", "HEAD").stdout.strip()
+
+    @property
+    def repo_is_initialized(self) -> bool:
+        """Determine whether the target repository is initialised"""
+        return Path(self.path).exists() and self._git_call("status", cwd=self.path)
+
+    @classmethod
+    def repo_is_supported(self, path: str) -> bool:
```
cgsheeh, line 186: `path` is really more of a `remote_url`.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        # This only tests a remote target without any local filesystem interaction, the CWD doesn't matter.
+        return self._git_call("ls-remote", path, cwd="/tmp")
+
+    def _git_initialize(self):
+        self.refresh_from_db()
+
+        if self.is_initialized:
+            raise
+
+        result = self._git_run("clone", self.pull_path, self.name)
```
cgsheeh, line 197: I don't think this has been tested, because `_git_run` is returning a `str` at the moment.

cgsheeh (dismissed):

src/lando/api/tests/conftest.py
shtrom: > Would be nice to add a docstring here. 

👍🏼 

I'll also move this to src.main.tests, to follow the move of  `test_git.py`.

I took the opportunity to clean up _those_ tests to use the `tmp_path: pathlib.Path` rather than `tmpdir: py.path` as we previously discussed. As the new test code is better separated,` it's easier to use the new pattern without impacting intra-file consistency of the legacy tests.

---

>  It might be also more useful to add a path parameter to this fixture, so that you can pass it elsewhere as an optional.

I'm not sure how to do it. I took a punt on something like this for the `conftest`

```
@pytest.fixture                                                                                                                                                                                                                               
def git_repo(tmp_path: pathlib.Path, repo_path: Optional[pathlib.Path] = None):
    repo_dir = tmp_path / "git_repo"
    if repo_path:
        repo_dir = repo_path       
  ....
```

and this in the test

``` 
@pytest.fixture
def repo_path():
    return Path("/non-existent-path")

def test_GitSCM_clone(tmp_path: Path, git_repo: Path, repo_path):
    ...
```

But that didn't seem to make any difference. Do you have an example of what it should look like here?

src/lando/api/tests/test_git.py
shtrom: Cool, I took that.

Ideally, we should be able to use `parametrize` with a fixture, but it's still an open isse for pytest https://github.com/pytest-dev/pytest/issues/349

src/lando/api/tests/test_git.py
shtrom: I did it at first, but I didn't like having the test rely on an internal method of the object under test, so I ended up this instead, which I think is simpler and more straightforward, at the cost of being a tad less DRY.

src/lando/api/tests/test_git.py
shtrom: Ah, yep. I added a `_monkeypatch_scm` function that allows to mock `_git_run` but to retain the side effect, so we can test that
* The correct `git` commands are called
* Their side effects are actually as desired

(And this allowed me to spot a bug 😅)

src/lando/main/models/repo.py
shtrom: It was there to keep the HgSCM-only code from #154 work. It's no longer needed.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+ENV_COMMITTER_NAME = "GIT_COMMITTER_NAME"
+ENV_COMMITTER_EMAIL = "GIT_COMMITTER_EMAIL"
+
+
+class GitSCM(AbstractSCM):
+    DEFAULT_ENV = {
+        "GIT_SSH_COMMAND": (
+            'ssh -o "StrictHostKeyChecking no" -o "PasswordAuthentication no"'
+        ),
+    }
```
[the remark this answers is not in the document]
shtrom: Agreed, SSH support works for now, and GitHub auth will be added with https://bugzilla.mozilla.org/show_bug.cgi?id=1934475. The suite PR https://github.com/mozilla-conduit/suite/pull/62 will be updated to use CGit instead, with HTTP-push support, as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1864902.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,298 @@
...
+        if force_push:
+            command += ["--force"]
+        command += [push_path]
+        if target:
+            command += [f"HEAD:{target}"]
+        self._git_run(*command)
+
+    @property
+    def REJECTS_PATH(self) -> Path:
+        return Path(self.path)
```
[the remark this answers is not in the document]
shtrom: Ah, yep. It's in the interface of the AbstractSCM, but it doesn't need to be anymore. I'll make a note to remove it as part of #154, too.

Makefile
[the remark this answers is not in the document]
shtrom: Yes, I cherry-picked the commit as I needed it, but it will disappear when the #163 is merged.

src/lando/api/legacy/workers/landing_worker.py
shtrom: https://bugzilla.mozilla.org/show_bug.cgi?id=1935549

src/lando/main/models/repo.py
shtrom: For reference: https://github.com/mozilla-conduit/lando/pull/154#discussion_r1868275696

Let's talk about the best way to do this when we catch up.

src/lando/main/scm/git.py
shtrom: Yeah definitely copy/past issues!

src/lando/main/scm/git.py
shtrom: This method is only used to get the return code for early `git`-based checks (it's also a `@staticmethod` so it can be called when we don't have an instance yet).

We can capture the output, but I'm not sure how we would provide it back. We don't throw exception for non-0 return code, so we don't have something to transport the output to the caller.

src/lando/api/legacy/workers/landing_worker.py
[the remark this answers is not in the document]
shtrom: Both. The landing_worker is now SCM-agnostic, so it's independent.

https://bugzilla.mozilla.org/show_bug.cgi?id=1935549

src/lando/api/tests/test_git.py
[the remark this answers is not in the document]
shtrom: @zzzeid had the exact same suggestion 😅

✅

src/lando/main/scm/git.py
shtrom: The problem is that `_git_run` relies on the instance existing, while `_git_call` doesn't (e.g., doesn't rely on `self.path`). It's used for early tests by the class (to see if a give `pull_path` is supported by git), and is supposed to be a lightweight call.

Now that I think more about it, maybe I can just remove the dependency on `self.path` from _git_run, and use that 🤔

src/lando/main/scm/git.py
shtrom: Ok, I found a better way! https://github.com/mozilla-conduit/lando/pull/147#discussion_r1872351301

src/lando/main/scm/git.py
shtrom: Yes and no: it can actually be a local git repo as well (that's how we test it (:)

src/lando/main/scm/git.py
shtrom: Yep. I found a few of those. Fortunately, that was dead code 😅

src/lando/api/tests/test_git.py
```diff
@@ -0,0 +1,81 @@
...
+        not new_untracked_file.check()
+    ), f"{new_untracked_file.basename} still present after clean"
+    if strip_non_public_commits:
+        assert (
+            not new_file.check()
+        ), f"Locally commited {new_file.basename} still present after stripping clean"
+    else:
+        assert (
+            new_file.check()
+        ), f"Locally commited {new_file.basename} missing after non-stripping clean"
```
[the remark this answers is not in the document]
shtrom: Ah! Neat!

src/lando/main/scm/git.py
shtrom: Yeah, `origin/HEAD` may not always be correct. It's probably safer to use `origin/{self.default_branch}`

I didn't consider creating temporary branches, but I think this will add overhead in garbage-collecting them, which may not be worth the effort.

shtrom:
@zzzeid @cgsheeh I addressed (most) of the comments, and resolved most of them. I left those that require more discussion open.

shtrom:
... and I just realised I rebased this to move a few fixup commits down the stack... Sorry about that.

I think you can just review the lot of fixup commits, which should be all the changes since your last reviews.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,258 @@
...
+        try:
+            cls._git_run("ls-remote", path)
+        except SCMException:
+            return False
+
+        return True
+
+    @classmethod
+    def _git_run(
+        cls, *args, cwd: Optional[str] = None, must_succeed: bool = True
```
cgsheeh, line 224: I don't think we use `must_succeed` anywhere, it can be removed.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,258 @@
...
+
+        cwd: str
+            Optional path to work in, default to '/'
+
+        Returns:
+            str: the standard output of the command
+        """
+        path = cwd or "/"
+        command = ["git"] + list(args)
+        logger.debug("Running " + " ".join(command) + " in " + path)
```
cgsheeh, line 241: The command logging here should include similar `extras` to the [`run_hg` method](https://github.com/mozilla-conduit/lando-api/blob/main/landoapi/hg.py#L251). These are useful for searching within the GCP log explorer.

src/lando/main/tests/test_git.py
```diff
@@ -0,0 +1,107 @@
...
+
+    mock_git_run.assert_called_with("clean", "-fdx", cwd=str(clone_path))
+    if strip_non_public_commits:
+        mock_git_run.assert_any_call(
+            "reset", "--hard", f"origin/{scm.default_branch}", cwd=str(clone_path)
+        )
+
+    assert (
+        strip_non_public_commits != new_file.exists()
+    ), f"strip_non_public_commits not honoured for {new_file}"
```
cgsheeh, line 92: This test checks that the newly introduced file was removed, but perhaps we should also check that the non-public commits were actually removed from the branch/repo? We could capture `scm.head_ref()`, clean the repo, then assert we aren't on that commit any more and that it was truly stripped.

src/lando/settings.py
```diff
@@ -214,13 +214,19 @@
 CELERY_ACCEPT_CONTENT = ["json"]
 CELERY_TASK_SERIALIZER = "json"
 
-DEFAULT_FROM_EMAIL = "Lando <lando@lando.test>"
+LANDO_USER_NAME = os.getenv("LANDO_USER_NAME", "Lando")
+LANDO_USER_EMAIL = os.getenv("LANDO_USER_EMAIL", "lando@lando.test")
+
+DEFAULT_FROM_EMAIL = f"{LANDO_USER_NAME} <{LANDO_USER_EMAIL}>"
```
cgsheeh, line 220: Bikeshed: since these new settings are all Git-specific, maybe we should reflect that in the names, like `LANDO_GIT_USER_NAME`, `LANDO_GIT_COMMITTER_NAME` etc?

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,258 @@
...
+        ),
+    }
+
+    default_branch: str
+
+    def __init__(self, path: str, default_branch: str = "main"):
+        self.default_branch = default_branch
+        super().__init__(path)
+
+    def clone(self, source):
```
cgsheeh, line 34: Type hint on `source`.

cgsheeh (changes requested):
Looking good, had a few more comments. The main blockers here are to add the logging extras, and the same meta-issues to the PR as in #154 (content from #163 in this PR, and the title).

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Returns:
+            None
+        """
+        command = ["push"]
+        if force_push:
+            command += ["--force"]
+        command += [push_path]
+        if target:
+            command += [f"HEAD:{target}"]
```
zzzeid, line 62: nit: I think it could be a tad bit easier to read if we sacrifice some repetition for readability, but not really sure.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,258 @@
...
+        os.environ[ENV_COMMITTER_NAME] = requester_email
+        os.environ[ENV_COMMITTER_EMAIL] = requester_email
+        logger.debug(
+            f"{ENV_COMMITTER_EMAIL} and {ENV_COMMITTER_NAME} set to {requester_email}"
+        )
+        try:
+            yield self
+        finally:
+            del os.environ[ENV_COMMITTER_NAME]
+            del os.environ[ENV_COMMITTER_EMAIL]
```
zzzeid, line 139: I think it may be a little better to forego env vars here and instead use command line arguments in the respective commands? (e.g., `--author`)

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,258 @@
...
+        Returns:
+            str: The target changeset
+        """
+        branch = target_cset or self.default_branch
+        self.clean_repo()
+        self._git_run("pull", "--prune", pull_path)
+        self._git_run("checkout", "--force", "-B", branch)
+        return self.head_ref()
+
+    def clean_repo(self, *, strip_non_public_commits: bool = True):
```
zzzeid, line 173: nit: I don't think the "public" commit reference is fully applicable in git. Maybe we can choose a new name that applies to both repos but doesn't reference a legacy term. Maybe we should have separate `clean` and `reset` methods.

src/lando/main/scm/git.py
```diff
@@ -0,0 +1,258 @@
...
+        """Determine whether the target repository is initialised"""
+        if not Path(self.path).exists():
+            return False
+
+        try:
+            self._git_run("status", cwd=self.path)
+        except SCMException:
+            return False
+
+        return True
```
zzzeid, line 210: I think a more positive check here would be good too, for example, checking the status code of the `git` command.

src/lando/main/scm/git.py
zzzeid, line 241: nit: also this may be more readable as an f-string.

zzzeid (dismissed):

src/lando/main/scm/git.py
shtrom: Ah! Nice!

src/lando/settings.py
shtrom: The `DEFAULT_FROM_EMAIL` constant pre-existed the git implementation, so I split it so I could reuse the components. [Looking through the code now, though, I don't see any use of it, except in a test].

I ain't painting it green! (unless you insist)

src/lando/main/scm/git.py
shtrom: We could have a more hard-coded `command`, but with two `if`s, we'd end up with 3 or 4 branches, so something like

```python

if force_push && target:
  command = [ "push", "--force", f"HEAD:{target}" ]
elif force_push:
  command = [ "push", "--force" ]
elif target:
  command = [ "push", f"HEAD:{target}" ]
else:
  command = [ "push" ]  # we could keep this one as a default, and just concatenate based on the three branches above
```

I'm not a fan, but happy to change it to that, or similar, if you think it's worthwhile.

src/lando/main/scm/git.py
shtrom: Looking through the manpage, it doesn't look like there is a `--committer` equivalent to `--author`. 

We may be able to do this via the `git config`. That would a bit too persistent for my taste I'd rather it be explicit in our code, rather than magically set in the persistent config of a dependency.

What do you think?

src/lando/main/scm/git.py
shtrom: Yep. I took it to mean “unpushed / local” commits. (The term seems to be “unpushed” for git https://duckduckgo.com/?t=ffab&q=git+unpublished+commits&ia=web)

Maybe `unpublished` commits? But then that's very similar to what we already have 😅

src/lando/main/scm/git.py
shtrom: `_git_run` will throw an SCMException on any non-0 status.

https://github.com/mozilla-conduit/lando/pull/147/files/635c25d299264d79adb887f6794b8e21c1d3f212#diff-2337f63ffc234b7c7cbd13e6ab6f4e1122832768b9ea5250508f5830a5bf71e5R247-R251

src/lando/main/scm/git.py
shtrom: I took a different approach, and used the `extra` parameter, to more closely match the logging from hg.

shtrom:
@zzzeid @cgsheeh I addressed most of your comments, and left a few more questions on the remaining ones.

src/lando/main/scm/__init__.py
```diff
@@ -5,7 +5,10 @@
...
 from lando.main.scm.hg import *
 
+# These can only be determined when all the subclasses of the AbstractSCM have been defined.
+SCM_CHOICES = {
+    klass.scm_type(): klass.scm_name() for klass in AbstractSCM.__subclasses__()
+}
 SCM_IMPLEMENTATIONS = {
-    SCM_GIT: GitSCM,
-    SCM_HG: HgSCM,
+    klass.scm_type(): klass for klass in AbstractSCM.__subclasses__()
```
shtrom: @cgsheeh Here's my suggestion to address your comment in https://github.com/mozilla-conduit/lando/pull/154#discussion_r1866870640

src/lando/main/scm/__init__.py
```diff
@@ -1,10 +1,61 @@
...
 
+# These can only be determined when all the subclasses of the AbstractSCM have been defined.
+SCM_CHOICES = {
+    klass.scm_type(): klass.scm_name() for klass in AbstractSCM.__subclasses__()
+}
 SCM_IMPLEMENTATIONS = {
-    # SCM_GIT: GitSCM,
-    SCM_HG: HgSCM,
+    klass.scm_type(): klass for klass in AbstractSCM.__subclasses__()
 }
```
shtrom, line 33: @cgsheeh Here's my suggestion to address your comment in https://github.com/mozilla-conduit/lando/pull/154#discussion_r1866870640

cgsheeh (approved):
LGTM!

src/lando/utils/management/commands/create_environment_repos.py
```diff
@@ -25,11 +25,15 @@
             "name": "git-repo",
             "url": "https://github.com/zzzeid/test-repo.git",
             "required_permission": SCM_LEVEL_1,
+            "scm_type": "git",
```
cgsheeh, line 28: Nit:
```suggestion
            "scm_type": GitScm.scm_type(),
```

cgsheeh (approved):

src/lando/api/tests/conftest.py
zzzeid: A simple example of this maybe is the `normal_patch` fixture.

shtrom: Yup, I just realised that when looking at the test for #182 (:

PR #148, opened 2024-11-01
commands: add command to generate requirements (bug 1898564)

zzzeid:
- add command that compiles requirements.txt file
- move various dependencies to pyproject.toml file
- run command to generate latest requirements.txt file
- add make commands


shtrom (approved):
LGTM!

PR #152, opened 2024-11-06
landing_jobs: return JSON responses for exceptions (bug 1929466)

shtrom:


shtrom:
I think so. I need to get back around to it, but it's not high priority.

PR #154, opened 2024-11-13
git landing worker vcs agility (Bug1895523)

shtrom:
This PR is the first part of Bug 1895523. It only moves existing code around, in preparation for a new SCM implementation for Git:
* move all Hg-only logic into the HgRepo
* introduce an AbstractScm class
* move the HgRepo into an HgSCM compatible with that class
* move non Hg-only logic (such as autoformatting [but not committing the changes]) out of the HgRepo, and into the worker
* update the worker to use the HgSCM

The second part will continue in #147.


src/lando/main/models/repo.py
```diff
@@ -38,6 +39,47 @@ class RepoError(Exception):
...
+        return self.vcs.head_ref()
+
+    def changeset_descriptions(self) -> list[str]:
+        return self.vcs.changeset_descriptions()
+
+    def update_repo(self, pull_path, target_cset):
+        return self.vcs.update_repo(pull_path, target_cset)
+
+    def format_stack(self, stack_size: int, bug_ids: list[str]):
+        return self.vcs.format_stack(stack_size, bug_ids)
```
cgsheeh, line 81: We should try and avoid this re-definition of `self.vcs` methods directly on the `Repo` class. As mentioned in our chat, we could do this with a method like `get_vcs` on `Repo` which returns a concrete object implementing `AbstractVcs`.

src/lando/main/models/repo.py
```diff
@@ -96,10 +138,12 @@ def get_mapping(cls) -> dict[str, "Repo"]:
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.path = self.system_path or self.get_system_path()
+
         if self.is_hg_repo:
-            self.hg = HgRepo(self.system_path or self.get_system_path())
+            self.vcs = HgVcs(self.path)
         else:
-            self.hg = None
+            self.vgs = None
```
cgsheeh, line 146:
```suggestion
            self.vcs = None
```

src/lando/main/models/repo.py
```diff
@@ -271,5 +302,6 @@ def apply_patch(self, patch_buffer: str):
     def last_commit_id(self) -> str:
         return self._run("rev-parse", "HEAD").stdout.strip()
 
-    def push(self):
+    def _native_push(self):
+        """XXX: unused?"""
```
cgsheeh, line 306: Let's remove this if unused.

src/lando/main/vcs/abstract_vcs.py
```diff
@@ -0,0 +1,138 @@
...
+        Args:
+            requester_email (str)
+        """
+        pass
+
+    # @property
+    # @abstractmethod
+    # def revision_id(self):
+    #     """Get the current revision_id. XXX: likely unused and broken"""
+    #     pass
```
cgsheeh, line 72: Let's remove this if unused.

src/lando/main/vcs/hg_vcs.py
```diff
@@ -0,0 +1,98 @@
...
+    def for_push(self, requester_email: str) -> ContextManager:
+        return self.hg.for_push(requester_email)
+
+    def for_pull(self) -> ContextManager:
+        return self.hg.for_pull()
+
+    # @property
+    # def revision_id(self):
+    #     """XXX: invalid?"""
+    #     return self.hg.revision.revision_id
```
cgsheeh, line 54: We should remove this if unused.

cgsheeh (changes requested):

src/lando/main/models/repo.py
shtrom: Actually, this may be something that's used, alongside the `_native_apply_patch` for push_to_try. It used to just be “repo.push()”, but I had to move it out of the way when I refactored the interface. The `_run` method it uses is git-only.

zzzeid (changes requested):
First pass!

Looks pretty good, mostly small changes. I think a couple of sections of this PR may be better separated into their own PRs. E.g., exception refactoring, parts of the code that were moved and/or modified (e.g., `hg_vcs`) which might make it easier to review.

src/lando/main/models/repo.py
```diff
@@ -38,6 +39,48 @@ class RepoError(Exception):
...
+        return self.vcs.head_ref()
+
+    def changeset_descriptions(self) -> list[str]:
+        return self.vcs.changeset_descriptions()
+
+    def update_repo(self, pull_path, target_cset):
+        return self.vcs.update_repo(pull_path, target_cset)
+
+    def format_stack(self, stack_size: int, bug_ids: list[str]):
+        return self.vcs.format_stack(stack_size, bug_ids)
```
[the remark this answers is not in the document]
shtrom: Yep, after discussing with @cgsheeh, we worked out a way to get the correct VCS directly into the worker.

src/lando/main/models/repo.py
shtrom: Cool. I think this means that this code will be (re)moved for the moment, as I'm taking a different approach where this will be implemented in a `GitScm` object. I'll leave this here for now, with a clearer name, and I'll just move it when I get to the `GitScm` implementation.

src/lando/main/models/repo.py
```diff
@@ -96,10 +139,12 @@ def get_mapping(cls) -> dict[str, "Repo"]:
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.path = self.system_path or self.get_system_path()
+
         if self.is_hg_repo:
-            self.hg = HgRepo(self.system_path or self.get_system_path())
+            self.vcs = HgVcs(self.path)
         else:
-            self.hg = None
+            self.vcs = None
```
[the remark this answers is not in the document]
shtrom: I think this'll change to a getter instead: this will make the Repo object a pure(r) ORM interface, which can provide an concrete Scm object to the landing_worker. This will allow us to get rid of the proxy function on the Repo object altogether, and let the worker interact directly with the Scm interface.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -236,20 +220,20 @@ def run_job(self, job: LandingJob) -> bool:
             True: The job finished processing and is in a permanent state.
             False: The job encountered a temporary failure and should be tried again.
         """
-        repo = job.target_repo
+        repo: Repo = job.target_repo
```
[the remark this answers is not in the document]
shtrom: I'm very influenced by my LSP here. I find it makes it easier to navigate the code, particularly as I'm not completely familiar with it yet.

src/lando/main/models/repo.py
```diff
@@ -38,6 +39,48 @@ class RepoError(Exception):
 class Repo(BaseModel):
     """Represents the configuration of a particular repo."""
 
+    # The path to the repository.
+    path: str
+    vcs: AbstractVcs
+
+    @property
+    def REJECTS_PATH(self) -> Path:
+        return self.vcs.REJECTS_PATH
```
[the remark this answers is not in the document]
shtrom: At this stage, the landing_worker still relies on this information, but I agree it may not be needed in the long run. For the moment, I prefer to limit the changes to scoping, and do heavier refactoring / deletions later, when I get a better sense of how Git support will influence the interface.

src/lando/main/models/repo.py
```diff
@@ -38,6 +39,48 @@ class RepoError(Exception):
...
+    ):
+        return self.vcs.push(push_path, target, force_push)
+
+    def last_commit_for_path(self, repo_path: str, path: str) -> bytes:
+        return self.last_commit_for_path(repo_path, path)
+
+    @property
+    def push_target(self):
+        """The target branch or bookmark to push to."""
+        return self.push_bookmark or None
```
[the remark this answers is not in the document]
shtrom: Yes, I had a wonder about this before. I thing I renamed bookmark to target in some places, under the assumption that it would be a target _branch_ in git, but I'm not 100% certain that's correct. Is it? What's a bookmark?

src/lando/main/vcs/exceptions.py
```diff
@@ -0,0 +1,74 @@
...
+
+from lando.api.legacy.hg import (
+    AutoformattingException,
+    HgmoInternalServerError,
+    LostPushRace,
+    NoDiffStartLine,
+    PatchConflict,
+    PushTimeoutException,
+    TreeApprovalRequired,
+    TreeClosed,
```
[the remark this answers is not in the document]
shtrom: Yep. This is a temporary piece: my main intent was to remove the explicit dependency of the `landing_worker` on the `legacy.hg` code. I just did some creative renaming so the naming is more correct, but this whole thing is not meant to stay very long.

src/lando/main/models/repo.py
```diff
@@ -96,10 +139,12 @@ def get_mapping(cls) -> dict[str, "Repo"]:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.path = self.system_path or self.get_system_path()
```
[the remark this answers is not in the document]
shtrom: The fact that there is both `self.system_path` and `self.get_system_path()` is a bit odd, yeah. The getter constructs a new string, and is used in the `save()` method if `self.system_path` is not defined. I suspect this is needed to initialise new objects the first time round.

Conversely, the `path` seems to be there to hide this ambivalence, so I'm not sure we can remove it. I tried, and it upset a few tests, which were easily fixed by dependings on the system_path instead, but it doesn't seem like the right way to go. I think it would make sense to make it a property 👍🏼

src/lando/main/vcs/hg_vcs.py
```diff
@@ -0,0 +1,122 @@
...
+
+import hglib
+
+from lando.api.legacy.hg import REJECTS_PATH, HgRepo
+from lando.main.vcs import AbstractVcs, VcsException
+
+logger = logging.getLogger(__name__)
+
+
+def remap_exceptions(fun):
```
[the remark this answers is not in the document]
shtrom: How about funk?

🕺🏼 🎸

src/lando/main/scm/exceptions.py
```diff
@@ -0,0 +1,80 @@
...
+)
+from lando.api.legacy.hg import (
+    TreeApprovalRequired as LegacyTreeApprovalRequired,
+)
+from lando.api.legacy.hg import (
+    TreeClosed as LegacyTreeClosed,
+)
+
+
+class ScmException(Exception):
```
cgsheeh, line 21: I'm not sure I follow what the `ScmException` and this re-mapping functionality is achieving, as the custom exceptions should be sufficient for both Git and Hg. With this change it seems we throw our own custom exception when we hit known conditions, then re-map it to these new custom exception types later on, which are themselves subclassed on another custom exception.

Most of the existing exceptions are thrown for things like "we detected that this push failed due to a `LostPushRace`, throw this custom exception so it is caught in the landing worker and the landing job status can be set appropriately". If we need to abstract those details at the SCM level, perhaps we want to make the existing exceptions subclass our new type instead of doing the conversion on the fly.

src/lando/main/scm/hg_scm.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class HgScm(AbstractScm):
+
+    path: str
+    hg: HgRepo
+
+    def __init__(self, path: str):
+        self.path = path
+        self.hg = HgRepo(path)
```
cgsheeh, line 36: How is the `HgRepo` different from the `HgScm`? It looks like all the custom methods on this class can just be moved there, and `HgRepo` can subclass `AbstractScm`.

cgsheeh (changes requested):

src/lando/main/scm/hg_scm.py
shtrom: The HgRepo is the legacy implementation. The HgScm should ultimately absorb all the logic of the HgRepo, which will allow us to get rid of the legacy bit.

src/lando/main/scm/exceptions.py
shtrom: The ScmException should replace the Hg-only exceptions. I didn't want to change the legacy code, nor to make it depend on non-legacy additions, which is why we have this remapping dance for exceptions that are essentially exactly the same, except the new ones are in a more practical namespace to cater for multiple SCMs.

src/lando/main/models/repo.py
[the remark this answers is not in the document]
cgsheeh: A bookmark is the Mercurial equivalent of a Git branch. In the Firefox Mercurial repos we don't use bookmarks, instead the `tip` commit on the `default` branch is always the only head, and we have a hook to enforce a single head per branch. For other Mercurial repos, such as v-c-t, there is an `@` bookmark which is essentially the equivalent of the default/main branch in a Git repo. It's the default bookmark which is checked out when the repo is first cloned.

For the Git landing worker, the equivalent push target would be a Git branch. I think we built this concept into Lando specifically for v-c-t, which is the only Mercurial repo where we wanted to maintain the `@` bookmark.

shtrom:
@zzzeid @cgsheeh I think this should be good for review. This is only the parts of the changes that refactor the VCS handling for agility, and ability to swap in another one.

This is a few separate, but heavily correlated changes, so I think it would be difficult to split in separate PRs. However, for legibility, I have split the changes per commits, based on the the parts of the code that are touched by each groups. Some of the commits on their own are not a functional tree, but the PR as a whole is.

cgsheeh (dismissed):
Looking great! Most of my comments are nitpicks/style things, but the "implement me" test should be removed or implemented before we land this.

I'm not entirely sure about the move of some of the autoformatting things to the landing worker from the concrete repo object, but it's simple enough to move back so I'm not fussed about it.

src/lando/main/models/repo.py
```diff
@@ -225,34 +205,36 @@ def phab_identifier(self) -> str | None:
...
 
-    def reset(self, branch=None):
-        self._run("reset", "--hard", branch or self.default_branch)
-        self._run("clean", "--force")
+    def _git_reset(self, branch=None):
+        self._git_run("reset", "--hard", branch or self.default_branch)
+        self._git_run("clean", "--force")
 
-    def apply_patch(self, patch_buffer: str):
+    def _git_apply_patch(self, patch_buffer: str):
```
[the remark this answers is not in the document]
shtrom: Yup, this happens in #147 https://github.com/mozilla-conduit/lando/pull/147/commits/15c02932c72f1f83b07a369359f550b779051533#diff-bfc6f824464a3ad577fdf49d01e2394d6dcd18703fab66587b7ef58bbe044c0b

src/lando/main/scm/__init__.py
```diff
@@ -0,0 +1,10 @@
+# flake8: noqa
+from .abstract_scm import *
+from .consts import *
+from .exceptions import *
+from .hg import *
```
zzzeid, line 5: Although we do this in `main.models`, it is conventionally discouraged to do this elsewhere (the models implementation being largely due to legacy porting, and since it is fairly self-contained). I think it would be better to just be explicit with the imports where they happen instead.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -416,3 +429,139 @@ def run_job(self, job: LandingJob) -> bool:
             self.phab_trigger_repo_update(repo.phab_identifier)
 
         return True
+
+    def read_lando_config(
```
zzzeid, line 433: These should probably go into a different module / class / mixin (maybe `lando.scm.utils`)

src/lando/main/scm/hg.py
```diff
@@ -0,0 +1,511 @@
...
+    hg_args: list
+    out: str
+    err: str
+
+    def __init__(self, hg_args: list, out: str, err: str, msg: str):
+        self.hg_args = hg_args
+        super().__init__(msg, out, err)
+
+    @classmethod
+    def from_hglib_error(cls, exc: hglib.error.CommandError):
```
zzzeid, line 53: A docstring summarizing what this method is doing would be helpful here.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -271,13 +272,27 @@ def run_job(self, job: LandingJob) -> bool:
...
 
                 try:
-                    repo.hg.apply_patch(patch_buf)
+                    # TODO: Rather than parsing the patch details from the full HG patch
+                    # stored in the job, we should read the revision's metadata (and
+                    # move to only store the diff in the patch_string, rather than an
+                    # export).
+                    patch_helper = HgPatchHelper(StringIO(revision.patch_string))
+                    if not patch_helper.diff_start_line:
+                        raise NoDiffStartLine()
```
zzzeid, line 283: This can be cleaned up by moving everything in lines 292-304 into this if statement, and moving the if statement outside the try block.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -330,7 +341,11 @@ def run_job(self, job: LandingJob) -> bool:
             # Run automated code formatters if enabled.
             if repo.autoformat_enabled:
                 try:
-                    replacements = repo.hg.format_stack(len(changeset_titles), bug_ids)
+                    landoini_config = self.read_lando_config(scm)
+                    replacements = self.format_stack(landoini_config, repo.path)
+                    self.commit_autoformatting_changes(
+                        scm, len(changeset_titles), bug_ids
+                    )
```
zzzeid, line 348: Amount of code in try block could be reduced here.

zzzeid (dismissed):

src/lando/main/models/repo.py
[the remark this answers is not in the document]
shtrom: Ok, so the `push_bookmark` is a column in the repo model.

Now that we have Git support, perhaps renaming it to `push_target` would be more SCM-agnostic (e.g., it would map to either an Hg bookmark or a Git branch).

However, is it worth the effort renaming and adding a migration?

@cgsheeh suggested the other that we should strive to get a clean DB to prod on the first prod-prod day of use, rather than having to migrate over time. So maybe now's the right time to migrate?

src/lando/api/tests/test_hg.py
```diff
@@ -243,11 +258,23 @@ def test_hg_exceptions():
...
+# @pytest.mark.parametrize(
+#     "git_returncode,hg_returncode,scm",
+#     ((255, 0, SCM_HG), (0, 255, SCM_GIT)),
+# )
+# @patch("lando.main.scm.GitScm")
+# @patch("lando.main.scm.HgScm")
+# @patch("lando.main.scm.git.subprocess")
+@pytest.mark.skip("implement me")
+def test_repo_is_supported(hg_clone):
+    pass
```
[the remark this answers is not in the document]
shtrom: Heh, I implemented it instead 😅

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -416,3 +429,139 @@ def run_job(self, job: LandingJob) -> bool:
...
+
+    def mach_path(self, path: str) -> Optional[Path]:
+        """Return the `Path` to `mach`, if it exists."""
+        mach_path = Path(path) / "mach"
+        if mach_path.exists():
+            return mach_path
+
+    def commit_autoformatting_changes(
+        self, scm: AbstractScm, stack_size: int, bug_ids: list[str]
+    ):
```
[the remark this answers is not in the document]
shtrom: This function doesn't return anything. Do we want to annotate that with `-> None`?

src/lando/main/scm/consts.py
```diff
@@ -0,0 +1,7 @@
+SCM_GIT = "git"
+SCM_HG = "hg"
+
+SCM_CHOICES = {
+    SCM_GIT: "Git",
+    SCM_HG: "Mercurial",
+}
```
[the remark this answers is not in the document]
shtrom: I think we can. I'll circle back to this, and do it on top of #147, to avoid too many conflicts.

src/lando/main/scm/hg.py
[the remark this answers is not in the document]
shtrom: This one is tricky as HgCommandError (and HgException) aren't defined yet. I can type-hint with `typing.Self`.

src/lando/main/scm/hg.py
```diff
@@ -0,0 +1,511 @@
...
+        msg = "hg error in cmd: hg {}: {}\n{}".format(
+            " ".join(str(arg) for arg in args),
+            out,
+            err,
+        ).rstrip()
+
+        for c in cls.__subclasses__():
+            for s in c.SNIPPETS:
+                if s in err or s in out:
+                    return c(args, out, err, msg)
```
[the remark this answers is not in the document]
shtrom: Ah! I agree in general, though it doesn't bother me too much in very tight loops. But ok, fixed 👍🏼

src/lando/main/scm/hg.py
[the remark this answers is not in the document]
shtrom: Yup, commit per commit, git considers the file as having been moved, but at the whole PR level, there are too many changes, and GitHub seems to have lost track of this.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -34,10 +29,28 @@
...
 from lando.utils.tasks import phab_trigger_repo_update
 
 logger = logging.getLogger(__name__)
 
+AUTOFORMAT_COMMIT_MESSAGE = """
+{bugs}: apply code formatting via Lando
+
+# ignore-this-changeset
+""".strip()
+
```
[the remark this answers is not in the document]
shtrom: This is intentional: the landing_working is now in charge of driving the autoformatting, and then requesting for the changes to be committed. As part of that, it provids the commit message to use to the SCMs.

src/lando/main/models/repo.py
```diff
@@ -104,10 +114,15 @@ def get_mapping(cls) -> dict[str, "Repo"]:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        if self.is_hg_repo:
-            self.hg = HgRepo(self.system_path or self.get_system_path())
-        else:
-            self.hg = None
+
+    def get_scm(self) -> AbstractScm:
```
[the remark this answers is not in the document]
shtrom: Making it a `@property` could be a problem, as they are not cached. This would result in returning a fresh instance everytime, which could be confusing if we expect state to be retained between uses (and we'd potentially be leaking memory).

That said, there is some awkward logic between the `scm` and the `_scm` property, that I think we can make cleaner. The main issue is that the `scm` property maps to the text column in the DB, so we can't readily swap it out.

Thinking more about it, we could have an `scm_interface` `@property`, but we'd need to do some memoisation ourselves, so we return the pre-existing instance, if any, e.g.,

```python
@property
def scm_interface(self):
    if not self._scm:
      if self.scm == SCM_HG:
          self._scm = HgSCM(self.path)
    return self._scm
```

... which is pretty much exactly the what the `get_scm` already does, but it's not a `@property`. So, yeah, could be made one, but we'd need to retain the `_scm` attribute. I'm not sure we'd benefit much here.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -236,20 +235,22 @@ def run_job(self, job: LandingJob) -> bool:
             True: The job finished processing and is in a permanent state.
             False: The job encountered a temporary failure and should be tried again.
         """
-        repo = job.target_repo
+        repo: Repo = job.target_repo
+        scm = repo.get_scm()
```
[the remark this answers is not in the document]
shtrom: So we'd no longer use the `scm` on its own, but always rely on the `repo.scm_interface` to access the SCM?

This would be more similar to the original state of the landing / repo objects, but I think it make more sense for the landing_worker to own the scm outright, rather than mediate all its interactions via the repo. This way all the configuration is held by the `repo`, and all the actions are performed via the `scm`, without any overlap save for getting the `scm` in the first place.

As a matter of fact, looking at the rest of this file, it doesn't need the `repo` at all. So we could just extract the `scm` once from the `repo`, and not even store it!

src/lando/main/scm/hg.py
```diff
@@ -0,0 +1,511 @@
...
+                "{node}",
+                "-l",
+                "1",
+                path,
+            ]
+        ).decode()
+
+    def apply_patch(
+        self, diff: str, commit_description: str, commit_author: str, commit_date: str
+    ):
```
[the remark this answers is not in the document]
shtrom: In theory, yes, but given the number of other changes around it, I think extracting it would create more conflicts that would need to be solved on the way out, and back on the way in when on the following commits, so I'm not sure it's worth it, as we'd end up in the same state.

src/lando/main/scm/__init__.py
shtrom: Ah, that's a pity. I find dot-imports to be convenient _within the same module_ (no `..`) to allow to keep the structure itself modular without hardcoding the full namespace, which makes it somewhat easier to move the module if need be.

I'll fix.

shtrom:
@zzzeid @cgsheeh I addressed (most) of the comments, and resolved most of them. I left those that require more discussion open.

src/lando/main/scm/consts.py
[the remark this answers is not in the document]
shtrom: (I still haven't done this one)

Makefile
```diff
@@ -3,10 +3,10 @@ DOCKER := $(shell which docker)
 DOCKER_COMPOSE := ${DOCKER} compose
```
cgsheeh, line 3: This includes changes from a different PR, they need to be removed before this lands.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -120,29 +133,15 @@ def process_merge_conflict(
...
-                        "--template",
-                        "{node}",
-                        "-l",
-                        "1",
-                        path,
-                    ]
-                ),
-            )
-            for path in failed_paths
+            (path, scm.last_commit_for_path(repo.path, path)) for path in failed_paths
```
cgsheeh, line 144: Isn't `repo.path` going to be the same as `scm.path` (internally as `self.path`) here? If that's the case we could simplify the signature.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -271,13 +272,38 @@ def run_job(self, job: LandingJob) -> bool:
 
             # Run through the patches one by one and try to apply them.
             for revision in job.revisions.all():
-                patch_buf = StringIO(revision.patch_string)
+                # TODO: Rather than parsing the patch details from the full HG patch
+                # stored in the job, we should read the revision's metadata (and
+                # move to only store the diff in the patch_string, rather than an
+                # export).
```
cgsheeh, line 278: Could you file a bug for this and reference it here?

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -416,3 +436,146 @@ def run_job(self, job: LandingJob) -> bool:
...
 
         return True
+
+    def apply_autoformatting(
+        self,
+        job: LandingJob,
+        scm,
+        landoini_config,
+        bug_ids,
+        changeset_titles,
```
cgsheeh, line 446: Type hints are missing here.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -416,3 +436,146 @@ def run_job(self, job: LandingJob) -> bool:
...
+        except subprocess.CalledProcessError as exc:
+            logger.warning("Failed to run automated code formatters.")
+            logger.exception(exc)
+
+            raise AutoformattingException(
+                "Failed to run automated code formatters.",
+                details=exc.stdout,
+            )
+
+    def run_code_formatters(self, path) -> str:
```
cgsheeh, line 496: Type hint.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -416,3 +436,146 @@ def run_job(self, job: LandingJob) -> bool:
...
+
+        If the `stack_size` is 1, the tip commit will get amended. Otherwise, a new
+        commit will be created on top of the stack (referencing all bugs involved in the
+        stack).
+        """
+        # When the stack is just a single commit, amend changes into it.
+        if stack_size == 1:
+            return scm.format_stack_amend()
+
+        else:
```
cgsheeh, line 576: Nit: the `else` is not necessary here.

src/lando/main/models/repo.py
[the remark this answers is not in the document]
cgsheeh: `push_target` sounds like a reasonable vcs-agnostic choice to me. :+1: 

> @cgsheeh suggested the other that we should strive to get a clean DB to prod on the first prod-prod day of use, rather than having to migrate over time. So maybe now's the right time to migrate?

This comment was more that we don't have any data in the production DB yet, so we shouldn't be afraid to make changes as we need to as we will not have to deal with actual data/behaviour changes. We will definitely do migrations over time (in fact easier migrations was one of the motivating factors in migrating to Django), but considering the whole DB is about to be wiped, if we need to change fields, this is a reasonable time to do it.

src/lando/main/models/repo.py
```diff
@@ -167,23 +158,27 @@ def save(self, *args, **kwargs):
...
-                raise ValueError("Could not determine repo type.")
+        if not self.scm:
+            self.scm = self._find_supporting_scm(self.pull_path)
 
         super().save(*args, **kwargs)
 
-    @property
-    def is_git_repo(self):
-        return self.scm is not None and self.scm == self.GIT
+    def _find_supporting_scm(self, pull_path) -> str:
```
cgsheeh, line 166: Type hint on `pull_path`.

src/lando/main/models/repo.py
```diff
@@ -167,23 +158,27 @@ def save(self, *args, **kwargs):
...
+            if impl.repo_is_supported(pull_path):
+                return scm
+        if self._is_git_repo:
+            return SCM_GIT
+        raise ValueError(f"Could not determine repo type for {pull_path}")
 
     @property
-    def is_hg_repo(self):
-        return self.scm is not None and self.scm == self.HG
+    def is_git_repo(self):
```
cgsheeh, line 180:
```suggestion
    def is_git_repo(self) -> bool:
```

cgsheeh (changes requested):
Looking good to me, just noticed a few missing type hints and had a few other small comments.

Also, please keep in mind that PRs will be squashed, so the title needs to be updated to match our commit message guidelines (one-word topic to start the title, imperative mood, space between "bug" and the number) ie `scm: refactor landing worker to be vcs-agnostic (Bug 1895523)` or something along those lines.

Makefile
shtrom: Yep, that will happen at the end, when everything else is sorted.

zzzeid (changes requested):
Couple more minor comments

src/lando/api/legacy/workers/landing_worker.py
shtrom: https://bugzilla.mozilla.org/show_bug.cgi?id=1936171

src/lando/main/models/repo.py
shtrom: I won't fix this one, as this bit of code gets killed in #147.

src/lando/api/tests/test_hg.py
```diff
@@ -268,13 +269,17 @@ def test_hgrepo_request_user(hg_clone):
...
+    "scm,repo_fixture_name,expected",
+    (
+        (HgSCM, "hg_clone", True),
+        (HgSCM, "tmpdir", False),
+    ),
+)
+def test_repo_is_supported(
+    scm: AbstractSCM, repo_fixture_name: str, expected: bool, request
+):
+    repo = request.getfixturevalue(repo_fixture_name)
```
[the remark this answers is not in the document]
shtrom: Ah, yeah, I knew I had another test like that!

src/lando/main/scm/__init__.py
```diff
@@ -1,10 +1,10 @@
 # flake8: noqa
-from .abstract_scm import *
-from .consts import *
-from .exceptions import *
-from .hg import *
+from lando.main.scm.abstract_scm import *
+from lando.main.scm.consts import *
+from lando.main.scm.exceptions import *
+from lando.main.scm.hg import *
```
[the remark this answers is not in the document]
shtrom: I'm confused now 😅 

What do you think it should look like?

shtrom:
@zzzeid questions on a couple of comments, the rest has been addressed.

src/lando/main/scm/consts.py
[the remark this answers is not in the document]
shtrom: How about this? https://github.com/mozilla-conduit/lando/pull/147/commits/d4f2f83d03a635a233be8e00912cd5f67fd4927a

src/lando/main/scm/__init__.py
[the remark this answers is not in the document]
shtrom: From our f2f chat: don't import *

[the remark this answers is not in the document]
shtrom: @zzzeid how about that 😅 https://github.com/mozilla-conduit/lando/pull/154/commits/6ac8ebc691d85834a0612fab89a72c7f1bbe74f0

cgsheeh (changes requested):

zzzeid (changes requested):
Few more small comments but this is looking very close to ready.

cgsheeh:
This looks great for me, with the exception of the code from an already-landed patch. Once you remove that this will be r+. :tada:

src/lando/api/tests/test_hg.py
```diff
@@ -155,60 +141,83 @@ def test_integrated_hgrepo_can_log(hg_clone):
...
+            raise hglib.error.CommandError(
+                (),
+                1,
+                b"",
+                b"forced fail: hunk FAILED -- saving rejects to file",
+            )
+        return original_run_hg(*args)
 
-        log_output = repo.run_hg(["log"])
-        assert "こんにちは" in log_output.decode("utf-8")
```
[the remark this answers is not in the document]
shtrom: Ah! Nice catch! This got lost in the conversion to `test_integrated_hgrepo_patch_success`. I readded some logic to do it.

src/lando/api/tests/test_try.py
```diff
@@ -240,7 +240,7 @@ def test_try_api_success_hgexport(
     )
 
     worker = LandingWorker(sleep_seconds=0.01)
-    hgrepo = HgRepo(hg_clone.strpath)
```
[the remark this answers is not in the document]
shtrom: Ah, yep. Fixed, but those tests are skipped for the moment. Probably something we want to address soon, too.

src/lando/main/models/repo.py
```diff
@@ -104,10 +114,15 @@ def get_mapping(cls) -> dict[str, "Repo"]:
...
-            self.hg = None
+
+    def get_scm(self) -> AbstractSCM:
+        """Return the SCM implementation associated with this Repository"""
+        if not self._scm:
+            if self.scm == SCM_HG:
+                self._scm = HgSCM(self.path)
+            else:
+                raise Exception(f"Repository type not supported: {self.scm}")
+        return self._scm
```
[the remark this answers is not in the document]
shtrom: Yup, let's circle back on this. One (minor) issue is that the `scm` attribute is already used as part of the model (you can see it being used on line 122, actually), so we'd need to do some renaming.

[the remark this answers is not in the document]
shtrom: Oh, ignore me, yeah, `scm_type` would make sense to me.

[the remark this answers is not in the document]
shtrom: Ok, cool, I did it!

shtrom:
@zzzeid I think all your comments are now addressed (including the `scm` property!). I left a few comments unresolved for double-checking, but we should be good to go.

If you can confirm, I'll squash and rebase this into clean commits, and we can get this for final landing approval!

src/lando/main/management/commands/start_landing_worker.py
```diff
@@ -45,7 +45,7 @@ def _handle(self, worker: Worker, repo_type: str):
         for repo in worker.enabled_repos:
             # Check if any associated repos are unsupported, raise exception if so.
             repo.raise_for_unsupported_repo_scm(repo_type)
-            scm = repo.get_scm()
+            scm = repo.scm
             scm.prepare_repo(repo.pull_path)
```
zzzeid, line 49: nit
```suggestion
            repo.scm.prepare_repo(repo.pull_path)
```

src/lando/main/models/repo.py
```diff
@@ -68,7 +68,7 @@ def push_target(self):
     # TODO: help text for fields below.
     name = models.CharField(max_length=255, unique=True)
     default_branch = models.CharField(max_length=255, default="main")
-    scm = models.CharField(
+    scm_type = models.CharField(
         max_length=3,
         choices=SCM_CHOICES,
```
zzzeid, line 73: nit:
```suggestion
        choices=SCM_TYPE_CHOICES,
```

src/lando/main/scm/hg.py
```diff
@@ -356,11 +361,7 @@ def format_stack_amend(self) -> Optional[list[str]]:
             raise exc
 
     def format_stack_tip(self, commit_message: str) -> Optional[list[str]]:
-        """Add an autoformat commit to the top of the patch stack.
-
-        Returns a list containing a single string representing the ID of the newly created commit.
```
zzzeid: May be removed by accident? Either way, no biggie.

src/lando/main/scm/hg.py
```diff
@@ -331,18 +339,15 @@ def update_from_upstream(self, source, remote_rev):
...
+        """Obtain remote head. We assume there is only a single head."""
         cset = self.run_hg(["identify", source, "-r", "default", "--id"]).strip()
 
         assert len(cset) == 12, cset
         return cset
 
     def format_stack_amend(self) -> Optional[list[str]]:
-        """Amend the top commit in the patch stack with changes from formatting.
-
-        Returns a list containing a single string representing the ID of the amended commit.
```
zzzeid: May be removed by accident?

src/lando/main/tests/test_models.py
```diff
@@ -44,14 +44,14 @@ def call(*args, **kwargs):
 
     repo.save()
 
-    assert repo.scm == scm
+    assert repo.scm_type == scm
 
 
 @pytest.mark.parametrize("scm,call_count", ((SCM_HG, 0), (SCM_GIT, 0)))
```
zzzeid, line 50: nit
```suggestion
@pytest.mark.parametrize("scm_type,call_count", ((SCM_HG, 0), (SCM_GIT, 0)))
```

zzzeid (changes requested):
Looks pretty good, just a couple of very small nits/comments.

zzzeid:
> If you can confirm, I'll squash and rebase this into clean commits, and we can get this for final landing approval!

@shtrom I think the only outstanding thing is the `push_target` discussion, otherwise it looks good!

>I think given our other discussions about mercurial bookmarks (on Slack, etc.) we should probably keep this as a mercurial-only feature until we make more robust decisions about handling git branches.

In this way, I am not sure we need to have `push_target` defined here at all, since it's just an alias to `push_bookmark`, and we still have outstanding discussions in terms of the approach to handling branches.

src/lando/main/scm/hg.py
shtrom: I did it on purpose: I kept the one-line description here, and have the full doc-string in the abstract SCM, so we have a single source of truth for the details.

shtrom:
@zzzeid I looked at removing the push_target, but got into a bit of a dead end.

It's used by the landing_worker https://github.com/mozilla-conduit/lando/blob/a45894150422ea19fd0746999c4466c02bd25730/src/lando/api/legacy/workers/landing_worker.py#L407 to instruct the SCM about where to push.

I think removing it right now would break the landing for repos that use it (so, I think we're only talking about VCT, and we said that it wasn't needed). Are we ok with this?

If so, I can do it easily, but I'd rather be sure it's safe.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -416,3 +462,138 @@ def run_job(self, job: LandingJob) -> bool:
...
 
         return True
+
+    def apply_autoformatting(
+        self,
+        scm: AbstractSCM,
+        landoini_config: Optional[configparser.ConfigParser],
+        bug_ids: list[str],
+        changeset_titles: list[str],
+    ) -> Optional[list[str]]:
```
cgsheeh, line 472: Docstring missing here.

cgsheeh (approved):
LGTM. Reminder to fix the commit message title format and write a nice commit message before squashing and landing in the repo.

zzzeid (approved):
Looks good to me, with a couple of nits.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -222,6 +221,60 @@ def extract_error_data(exception: str) -> tuple[list[str], list[str]]:
...
+            message = (
+                "Lando failed to format your patch for conformity with our "
+                "formatting policy. Please see the details below.\n\n"
+                f"{exc.details()}"
+            )
+
+            logger.exception(message)
+
+            job.transition_status(LandingJobAction.FAIL, message=message)
+            self.notify_user_of_landing_failure(job)
```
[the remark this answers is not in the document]
shtrom: ok, fixed!

cgsheeh (approved):

PR #155, opened 2024-11-22
uplift: port functionality (bug 1915695)

zzzeid:
- port existing uplift functionality
- adapt form to use new fields
- update endpoints and connect them to legacy api

Note: This ports an early version of uplift functionality that was moved over as part of the initial repo migration.

### DIFF
--- a/src/lando/api/legacy/uplift.py
+++ b/src/lando/api/legacy/uplift.py
@@ -8,6 +8,7 @@ from typing import (
 )
 
 import requests
+from django.core.cache import cache
 from packaging.version import (
     InvalidVersion,
     Version,
@@ -22,7 +23,11 @@ from lando.api.legacy.stacks import (
     request_extended_revision_data,
 )
 from lando.main.models import Repo
-from lando.utils.phabricator import PhabricatorClient
+from lando.utils.phabricator import (
+    PhabricatorAPIException,
+    PhabricatorClient,
+    get_phabricator_client,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -59,17 +64,24 @@ def get_uplift_request_form(revision: dict) -> Optional[str]:
     return bug
 
 
-# @cache.cached(
-#    key_prefix="uplift-repositories"
-# )
-def get_uplift_repositories(phab: PhabricatorClient) -> list:
-    repos = phab.call_conduit(
-        "diffusion.repository.search",
-        constraints={"projects": ["uplift"]},
-    )
+def get_uplift_repositories() -> list:
+    CACHE_KEY = "uplift-repos"
 
-    repos = phab.expect(repos, "data")
-
+    repos = cache.get(CACHE_KEY)
+    if not repos:
+        phab = get_phabricator_client()
+        try:
+            repos = phab.call_conduit(
+                "diffusion.repository.search",
+                constraints={"projects": ["uplift"]},
+            )
+        except PhabricatorAPIException as e:
+            logger.exception(e)
+            repos = []
+        else:
+            repos = phab.expect(repos, "data")
+            repos = [phab.expect(repo, "fields", "shortName") for repo in repos]
+            cache.set(CACHE_KEY, repos)
     return repos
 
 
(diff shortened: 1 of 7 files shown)


src/lando/api/legacy/uplift.py
```diff
@@ -59,17 +64,24 @@ def get_uplift_request_form(revision: dict) -> Optional[str]:
...
-# )
-def get_uplift_repositories(phab: PhabricatorClient) -> list:
-    repos = phab.call_conduit(
-        "diffusion.repository.search",
-        constraints={"projects": ["uplift"]},
-    )
-
-    repos = phab.expect(repos, "data")
-
+def get_uplift_repositories() -> list:
```
shtrom, line 67: Is there a type that we could make it a `list[]` of?

src/lando/ui/legacy/forms.py
```diff
@@ -12,8 +15,32 @@ class TransplantRequestForm(forms.Form):
 class UpliftRequestForm(forms.Form):
     """Form used to request uplift of a stack."""
 
-    revision_id = forms.RegexField(regex="D[0-9]+$")
-    repository = forms.CharField()
+    revision_id = forms.RegexField(
+        regex="D[0-9]+$",
```
shtrom, line 19: We can make the RE stricter.

```suggestion
        regex="^D[0-9]+$",
```

src/lando/ui/legacy/forms.py
```diff
@@ -12,8 +15,32 @@ class TransplantRequestForm(forms.Form):
...
+    )
+    repository = forms.ChoiceField(
+        widget=forms.Select(),
+        choices=((repo, repo) for repo in get_uplift_repositories()),
+    )
+
+    def clean_repository(self) -> str:
+        repo_name = self.cleaned_data["repository"]
+        all_repos = Repo.get_mapping()
+        repository = all_repos.get(repo_name)
```
shtrom, line 31: Does this only select upliftable repos?

src/lando/ui/legacy/revisions.py
```diff
@@ -21,6 +21,49 @@
...
+    @force_auth_refresh
+    def post(self, request: HttpRequest) -> HttpResponse:
+        """Process the uplift request submission."""
+        uplift_request_form = UpliftRequestForm(request.POST)
+        errors = []
+
+        if not request.user.is_authenticated:
+            raise PermissionError()
+
+        if uplift_request_form.is_valid() and not errors:
```
shtrom, line 34: `errors` will always be empty here, as it only gets set below on line 53. I'd suggest moving the error management LL52-64 up before this, and return early, and keep the happy case as the continuation of the method.

shtrom (changes requested):
I think the error-handling is `Uplift.post` is in the wrong order.

Apart from that (and a few more nits), this looks pretty good.

shtrom:
It might be worth adding some tests for the API bit, too.

src/lando/ui/legacy/forms.py
zzzeid: No, it will just verify that the repo is in fact a repo. I will clean up this section a little, but I think in the long run we should probably sync an "uplift" flag that is on the Repo object itself, and that way we can query these repos directly without contacting phabricator.

src/lando/api/legacy/uplift.py
zzzeid: I can add that.

Side note: this is an early version of uplift functionality, and as part of updating the repo with all the changes in the lando-api and lando-ui repos, this code will be affected.

src/lando/ui/legacy/forms.py
shtrom: Yep. It might be worth having a deliberate action or command to sync from Phabricator, but otherwise not do it, or do it periodically rather than on demand. Looking in suite, I saw that lando fires a large number of requests at Phabricator, and reducing them could potentially make the whole experience faster.

zzzeid:
> It might be worth adding some tests for the API bit, too.

This is an early version of uplift, with an existing test that is marked as xfail. I don't want to add new functionality here that may be coming in from additional pulls from the legacy repos (which itself contains a few more change to the uplift functionality).

shtrom (approved):
LGTM!

PR #156, opened 2024-11-25
repo: loosen validation on repo system_path (bug 1922274)

shtrom:


src/lando/main/models/repo.py
```diff
@@ -29,6 +30,11 @@
 )
 
 
+def validate_path_in_repo_root(value):
+    if not value.startswith(settings.REPO_ROOT):
+        raise ValidationError(f"Path must be in {settings.REPO_ROOT}.")
```
zzzeid, line 35: Might be good to use pathlib here to do a bit more rigorous validation (e.g., check that the parent directory is the repo root.)

zzzeid (changes requested):

zzzeid (approved):

PR #157, opened 2024-11-26
revisions: fix commit flag parsing (bug 1927930)

zzzeid:


shtrom (approved):
I'm a bit confused about _why_ this fixes the bug, but I tested it, and it does nicely: the JS send a JSON-encoded string, now.

I have also noticed that, when nothing is ticked, it now sends `"null"` instead of `"[]"`, but this seems functionally equivalent.

zzzeid:
> I'm a bit confused about _why_ this fixes the bug, but I tested it, and it does nicely: the JS send a JSON-encoded string, now.
> 
> I have also noticed that, when nothing is ticked, it now sends `"null"` instead of `"[]"`, but this seems functionally equivalent.

`cleaned_data` contains the data that is already converted to the correct type (via the `to_python` method). This is why this works as expected.

PR #159, opened 2024-11-28
middleware: various CSP fixes (bug 1934038)

zzzeid:
- add *.googleapis.com
- fix inline styles for debug pages
- remove unneeded hosts (cloudfront.net and googleusercontent.com)

### DIFF
--- a/src/lando/middleware.py
+++ b/src/lando/middleware.py
@@ -32,13 +32,19 @@ class ResponseHeadersMiddleware:
             "font-src 'self' https://code.cdn.mozilla.net",
             "frame-ancestors 'none'",
             "frame-src 'none'",
-            "img-src 'self' *.cloudfront.net *.gravatar.com *.googleusercontent.com",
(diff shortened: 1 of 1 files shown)


src/lando/middleware.py
```diff
@@ -32,14 +32,20 @@ def __call__(self, request: HttpRequest) -> HttpResponse:
...
             "manifest-src 'none'",
             "media-src 'none'",
             "object-src 'none'",
-            "style-src 'self' https://code.cdn.mozilla.net",
+            "script-src 'self' *.googleapis.com",
             "worker-src 'none'",
         ]
 
+        if settings.DEBUG and response.status_code >= 400:
+            # This is needed for debug pages which have inline CSS.
```
cgsheeh, line 44: Nit:

```suggestion
            # `unsafe-inline` is needed for debug pages which have inline CSS.
```

cgsheeh (approved):

PR #160, opened 2024-12-02
landing_worker: set debug level according to verbosity CLI option (bug 1935044)

shtrom:


zzzeid (changes requested):
I am a little hesitant to add this, since management commands generally use `self.stdout` and `self.stderr`, along with various `self.style` additions and interpretations of `CommandError`. So I think we are mixing those concepts with logging here, which generally does not affect output defined within management commands. I think the intended way to use this is to determine where and when to produce output within a management command (e.g., https://github.com/django/django/blob/main/django/core/management/commands/migrate.py#L198).

PR #161, opened 2024-12-02
settings: add remote_settings module (bug 1909317)

zzzeid:


shtrom (commented):
LGTM, but it looks like we lost the `if ENVIRONMENT.is_remote:` conditional. Is it handled by something else automatically, or do we need to re-add this logic?

zzzeid:
> LGTM, but it looks like we lost the `if ENVIRONMENT.is_remote:` conditional. Is it handled by something else automatically, or do we need to re-add this logic?

Ah yes, this is handled in [this PR](https://github.com/mozilla-it/webservices-infra/pull/3452) and is no longer needed.

dklawren (approved):

PR #162, opened 2024-12-02
middleware: more CSP fixes (bug 1934091)

zzzeid:
- update font-src and style-src entires with correct values


shtrom (approved):
LGTM. 

We could move `*.googleapis.com https://code.cdn.mozilla.net` into a single constant rather than duplicating it, but this is good enough to be merged as is.

PR #163, opened 2024-12-04
tests: default to running in local stack, with option to do INSUITE=1…

shtrom:
… (bug 1934761)

Also, don't expose the DB port on the host

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ __pycache__
 env
 staticfiles
 src/lando/version.py
+.test-use-suite
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,10 @@ DOCKER := $(shell which docker)
 DOCKER_COMPOSE := ${DOCKER} compose
 ARGS_TEST ?=
 
-ifeq ($(STANDALONE), 1)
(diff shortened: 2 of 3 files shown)


zzzeid (approved):
Not opposed to this, but it basically just changes the default behaviour (i.e., it will make it more convenient for some people, less convenient for others, depending on preference). We should probably have a command that changes the default behaviour instead.

Also: commit message first line is too long, should be modified to not get truncated.

shtrom:
> it basically just changes the default behaviour (i.e., it will make it more convenient for some people, less convenient for others, depending on preference).

Ah, yeah, it does. I thought reducing the number of dependencies to run the tests as a was good, but I could have missed something. What are the advantages of running in suite over locally?

The main issue I saw was port collision (partially addressed in the original commit, with the counterpart in https://github.com/mozilla-conduit/suite/pull/63). After doing that, both modes seemed equivalent to me.

If there's more, we can just swap the default back.

> We should probably have a command that changes the default behaviour instead.

Good idea. I did just that in the new commit. It has a `test-use-local` and `test-use-suite` command that sets the default in a local `.gitignore`d file, and then you can forget about it.

It can also be overridden in an ad hoc fashion with `make INSUITE=[0|1]`.

PR #176, opened 2024-12-12
skip failing repos (bug 1936693)

shtrom:
Currently, workers (and setup_dev) fail when any of the repos they should handle is misconfigured. This change makes handling those cases more graceful by
* logging errors
* skipping the repo and continuing with what's supported

(This PR is on top of #154)


src/lando/main/management/commands/start_landing_worker.py
```diff
@@ -46,11 +46,23 @@ def _handle(self, worker: Worker, repo_type: str):
             # Check if any associated repos are unsupported, raise exception if so.
             repo.raise_for_unsupported_repo_scm(repo_type)
             scm = repo.get_scm()
-            scm.prepare_repo(repo.pull_path)
+            repos_to_handle = []
+            try:
+                scm.prepare_repo(repo.pull_path)
+                repos_to_handle.append(repo)
```
zzzeid, line 52: This line should be outside the `try` block.

src/lando/main/management/commands/start_landing_worker.py
```diff
@@ -46,11 +46,23 @@ def _handle(self, worker: Worker, repo_type: str):
...
+            except Exception as e:
```
zzzeid, line 53: What sorts of exceptions are we expecting here? It would be good to be more explicit about them.

src/lando/main/management/commands/start_landing_worker.py
```diff
@@ -46,11 +46,23 @@ def _handle(self, worker: Worker, repo_type: str):
...
+                        "exception": e,
+                        "origin": repo.pull_path,
+                        "scm": scm.scm_type,
+                    },
+                )
 
         # Continue with starting the worker.
         try:
-            landing_worker = LandingWorker(worker.enabled_repos)
+            landing_worker = LandingWorker(repos_to_handle)
```
zzzeid, line 65: If there's a temporary failure for whatever reason, this would mean that we have to force-restart the worker in order to pick up additional repos. I think it makes sense to consider that in the handling of these errors, for example, by detecting which repos failed and maybe retrying in the future.

src/lando/main/management/commands/start_landing_worker.py
```diff
@@ -46,11 +46,23 @@ def _handle(self, worker: Worker, repo_type: str):
             # Check if any associated repos are unsupported, raise exception if so.
             repo.raise_for_unsupported_repo_scm(repo_type)
             scm = repo.get_scm()
-            scm.prepare_repo(repo.pull_path)
+            repos_to_handle = []
```
zzzeid, line 49: This should be outside the for loop, otherwise it is being reset on every iteration.

zzzeid (changes requested):

src/lando/main/management/commands/start_landing_worker.py
shtrom: 🤦🏼 of course. How did that even work?!

src/lando/main/management/commands/start_landing_worker.py
shtrom: I saw `ValueError`s. However I don't think we want to be specific now, so we can capture all errors, expected or not, rather than die on unexpected ones.

src/lando/main/management/commands/start_landing_worker.py
shtrom: Ah, yes. We only want to execute it when the `prepare_repo` has been successful, but that can go into an `else` clause instead.

src/lando/main/management/commands/start_landing_worker.py
shtrom: Yes.

Adding a retry would be substantially more work here, but a log message on temporary failures may not be sufficient in prod... I guess we could add a config option to determine whether we want to fail gracefully or not?

shtrom:
@zzzeid fixed obvious mistakes, and a couple of questions.

shtrom:
Let's stick with the workers dying as a simple way to have a periodic retry.

PR #181, opened 2024-12-18
remote_settings: add email configuration (bug 1898324)

zzzeid:
- add settings for SMTP configuration in remote environments
- set default console backend for local environment


zzzeid:
Also see https://github.com/mozilla-it/webservices-infra/pull/3622.

src/lando/remote_settings.py
```diff
@@ -39,3 +39,12 @@
     "root": {"handlers": ["null"]},
     "disable_existing_loggers": True,
 }
+
+EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
```
cgsheeh, line 43: `EMAIL_BACKEND` is defined here and in `settings.py`. For consistency, you could remove one, or import to the required file if it should be defined in both places.

cgsheeh (approved):

src/lando/remote_settings.py
zzzeid: This is an overridden from the value in `settings`, which is set to the console backend.

cgsheeh: Make sense, I didn't notice they are different. Not great that they have the same name and almost the exact same namespace, tbh.

shtrom (approved):
LGTM.

PR #182, opened 2024-12-19
landing_worker: fix handling of merge conflicts (bug 1938162) (bug 1935549)

shtrom:
Add test coverage for good measure (and remove the never-used `job.error_breakdown["content"]`).

Much better!

Before:
2024-12-19T16:54:18,990160289+11:00
After:
2024-12-19T16:55:32,929871564+11:00

### DIFF
--- a/src/lando/api/legacy/spec/swagger.yml
+++ b/src/lando/api/legacy/spec/swagger.yml
@@ -1,4 +1,4 @@
-swagger: '2.0'
+swagger: "2.0"
 info:
   title: Lando API
   description: An API to get and land Phabricator revisions.
@@ -31,17 +31,17 @@ paths:
           required: true
           in: body
           schema:
-            $ref: '#/definitions/TransplantRequest'
+            $ref: "#/definitions/TransplantRequest"
       responses:
         200:
           description: OK
           schema:
-            $ref: '#/definitions/TransplantAssessment'
+            $ref: "#/definitions/TransplantAssessment"
         default:
           description: Unexpected error
           schema:
             allOf:
-              - $ref: '#/definitions/Error'
+              - $ref: "#/definitions/Error"
   /transplants:
     get:
       operationId: landoapi.api.transplants.get_list
@@ -59,12 +59,12 @@ paths:
           schema:
             type: array
             items:
-              $ref: '#/definitions/Transplant'
+              $ref: "#/definitions/Transplant"
         default:
           description: Unexpected error
           schema:
             allOf:
-              - $ref: '#/definitions/Error'
+              - $ref: "#/definitions/Error"
     post:
       description: |
         Sends request to transplant service. If a Phabricator API
@@ -77,7 +77,7 @@ paths:
           in: body
           schema:
             allOf:
-              - $ref: '#/definitions/TransplantRequest'
+              - $ref: "#/definitions/TransplantRequest"
               - type: object
                 properties:
                   confirmation_token:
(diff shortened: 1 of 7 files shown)


zzzeid (dismissed):
Looks good to me, just a couple of small changes.

src/lando/main/scm/abstract_scm.py
```diff
@@ -27,6 +27,11 @@ def scm_name(cls) -> str:
         """Return a _human-friendly_ string identifying the supported SCM (e.g.,
         `Mercurial`)."""
 
+    @classmethod
+    def reject_path(cls) -> Path:
```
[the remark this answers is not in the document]
shtrom: Yup. Python 3.13 is gonna deprecat abstract properties, which is a bit unfortunate in this case.

src/lando/api/tests/test_landings.py
```diff
@@ -456,6 +456,69 @@ def test_lose_push_race(
     assert job.status == LandingJobStatus.DEFERRED
 
 
+@pytest.mark.django_db
+def test_merge_conflict(
```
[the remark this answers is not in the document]
shtrom: Agreed. I won't do it here, but I meant to do some cleanup as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1934485#c3

src/lando/api/legacy/spec/swagger.yml
zzzeid, line 1: Unrelated change in this yaml file (looks like formatting). Also, wanted to take this opportunity to mention [bug 1927139](https://bugzilla.mozilla.org/show_bug.cgi?id=1927139).

src/lando/ui/jinja2/stack/partials/timeline.html
zzzeid, line 1: Also unrelated change here to this html file.

We should probably add Jinja2 files to the linters and run them on all the templates though, in a separate PR.

zzzeid (changes requested):
Looks good, minus the unrelated changes that should be removed from this PR.

src/lando/ui/jinja2/stack/partials/timeline.html
shtrom: https://bugzilla.mozilla.org/show_bug.cgi?id=1938838

src/lando/api/legacy/spec/swagger.yml
shtrom: Ok, so this file needs to go.

src/lando/ui/jinja2/stack/partials/timeline.html
shtrom: There is a change in the HTML file, though, so I thought I might as well lint it once and for all (in a separate commit), prior to making the change.

src/lando/main/scm/abstract_scm.py
```diff
@@ -27,6 +27,11 @@ def scm_name(cls) -> str:
         """Return a _human-friendly_ string identifying the supported SCM (e.g.,
         `Mercurial`)."""
 
+    @classmethod
+    def get_rejects_path(cls) -> Path:
+        """Return the path where the SCM stores reject files."""
+        return Path(".")
```
zzzeid, line 33: nit: can you add an explanation why this is set as the default? Is it ever used?

zzzeid (approved):

src/lando/main/scm/abstract_scm.py
shtrom: Will do. It's used by the landing worker to locate the REJECTS, as Git and Hg put them in different locations https://github.com/mozilla-conduit/lando/pull/182/files#diff-15a775db92cbe9bf7e735238755c515149c96ba8bebb70ca8d0bbb1e0f9657daR164...

But the local directory seems like a reasonable default.

PR #189, opened 2025-01-02
GitSCM: add GitHub token support (bug 1934475)

shtrom:


src/lando/main/scm/git.py
```diff
@@ -56,13 +78,64 @@ def push(
     ):
         """Push local code to the remote repository."""
         command = ["push"]
+
         if force_push:
             command += ["--force"]
+
+        if match := re.match(GITHUB_URL_RE, push_path):
+            # We only fetch a token if no authentication is explicitely specified in
+            # the push_url
```
cgsheeh, line 87:
```suggestion
            # We only fetch a token if no authentication is explicitly specified in
            # the push_url.
```

src/lando/main/tests/test_git.py
```diff
@@ -138,6 +139,34 @@ def test_GitSCM_clean_repo(
...
+    scm = GitSCM(str(git_repo))
+    scm._git_run = MagicMock()
+    scm._get_github_token = MagicMock()
+    scm._get_github_token.side_effect = ["ghs_yolo"]
+
+    scm.push("https://github.com/some/repo")
+
+    assert scm._git_run.call_count == 1
+    assert scm._get_github_token.call_count == 1
+    assert "git:ghs_yolo@github.com" in scm._git_run.call_args[0][1]
```
cgsheeh, line 152: Assert strings clarifying the significance of these checks would be nice.

Makefile
```diff
@@ -1,7 +1,7 @@
 SHELL := /bin/bash
 DOCKER := $(shell which docker)
 DOCKER_COMPOSE := ${DOCKER} compose
-ARGS_TEST ?=
+ARGS_TESTS ?=
```
cgsheeh, line 4: Unrelated?

compose.yaml
```diff
@@ -53,6 +53,25 @@ services:
...
+    environment:
+      DEFAULT_GRACE_SECONDS: 10
+      LANDING_WORKER_USERNAME: app
+      # See [0] and [1] to setup up and install a GitHub app
+      # [0] https://docs.github.com/en/enterprise-cloud@latest/apps/creating-github-apps/registering-a-github-app/registering-a-github-app
+      # [1] https://docs.github.com/en/enterprise-cloud@latest/apps/using-github-apps/installing-your-own-github-app
+      # GITHUB_APP_ID: 123456789
+      # GITHUB_APP_PRIVKEY: |
+      #  -----BEGIN RSA PRIVATE KEY-----
+      #  ...
```
cgsheeh, line 68: Should these be uncommented for testing the Github app auth? It's unclear to me what this comment is for.

pyproject.toml
```diff
@@ -30,6 +30,7 @@ dependencies = [
     "rs-parsepatch",
     "sentry-sdk",
     "setuptools-scm",
+    "simple-github",
```
cgsheeh, line 33: You will need to re-generate `requirements.txt` to include this in the Docker container.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -314,7 +314,7 @@ def run_job(self, job: LandingJob) -> bool:
                 # Try again, this is a temporary failure.
                 return False
             except Exception as e:
-                message = f"Unexpected error while fetching repo from {repo.pull_path}."
+                message = f"Unexpected error while fetching repo from {repo.name}."
```
cgsheeh, line 317: Why the switch to `repo.name` here? We don't overwrite the `pull_path`/`push_path` outside of `scm.push()`, right?

cgsheeh (changes requested):

Makefile
shtrom: Yup, it's a nit I ran into as I was testing. I think it's small enough to bundle with this?

compose.yaml
shtrom: It's meant to be documentation of sort. 
As this requires a dedicated app registration, and a private key, it can't be setup by default in the repo and/or I don't think we should ship a key, albeit a test one.

I left that here as a mind-jogger of how to configure it. Arguably, I should probably also update the readme.

src/lando/api/legacy/workers/landing_worker.py
shtrom: We can also include the username and password in the push and pull paths. That's how suite does for CGit. It's a corner-case, but it's still a risk of leaking those (hardcoded) creds.

src/lando/main/scm/git.py
shtrom: 😅

src/lando/main/scm/git.py
```diff
@@ -56,13 +78,64 @@ def push(
...
+                    push_path = f"https://git:{token}@github.com/{match['owner']}/{match['repo']}"
+
         command += [push_path]
+
         if push_target:
             command += [f"HEAD:{push_target}"]
+
         self._git_run(*command, cwd=self.path)
 
+    @staticmethod
```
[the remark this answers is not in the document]
shtrom: I wondered about this, but decided to have a first cut at it with the least amount of moving parts. This is it, which I think is good enough.

Looking at it now, and how small an addition it is, and knowing that we are mainly targeting GitHub, I'm tempted to leave it here, as adding this in a separate module (that currently doesn't exist) seems a bit overkill as I doubt we'll have to add any other auth system any time soon (at which point, I agree we should split it out better).

compose.yaml
shtrom: I left that here, but added doc about the most important parameters in the README, too.

shtrom: I had a change of heart 😬 I moved everything neatly into the README.

src/lando/main/scm/git.py
```diff
@@ -17,6 +22,23 @@
...
+#     pct-encoded   = "%" HEXDIG HEXDIG
+#     sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
+#                 / "*" / "+" / "," / ";" / "=
+#
+# [0] https://www.rfc-editor.org/rfc/rfc3986
+URL_USERINFO_RE = re.compile(
+    "(?P<userinfo>[-A-Za-z0-9:._~%!$&'*()*+;=]*@)", flags=re.MULTILINE
+)
+GITHUB_URL_RE = re.compile(
+    f"https://{URL_USERINFO_RE.pattern}?github.com/(?P<owner>[-A-Za-z0-9]+)/(?P<repo>[^/]+)"
```
cgsheeh, line 39: TIL about `.pattern`, very useful!

src/lando/main/scm/git.py
```diff
@@ -56,13 +78,64 @@ def push(
...
         command += [push_path]
+
         if push_target:
             command += [f"HEAD:{push_target}"]
+
         self._git_run(*command, cwd=self.path)
 
+    @staticmethod
+    def _get_github_token(repo_owner: str, repo_name: str) -> Optional[str]:
+        """Obtain a fresh GitHub token to psuh to the specified repo.
```
cgsheeh, line 111:
```suggestion
        """Obtain a fresh GitHub token to push to the specified repo.
```

src/lando/main/scm/git.py
```diff
@@ -238,6 +313,10 @@ def _git_run(cls, *args, cwd: Optional[str] = None) -> str:
 
         return out
 
+    @staticmethod
+    def _redact_url_userinfo(s: str) -> str:
+        return re.sub(URL_USERINFO_RE, "[REDACTED]@", s)
```
cgsheeh, line 318: Please don't use single letter variable names. :)

cgsheeh (approved):

src/lando/main/scm/git.py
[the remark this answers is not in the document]
shtrom: Yeah! Me too (:

src/lando/main/scm/git.py
shtrom: Yeah! Me too!

PR #190, opened 2025-01-06
test_landings: increase git landing coverage (bug 1934485)

shtrom:
* upgrade landings tests to work with both Hg and Git repos
* fix a few simple bugs in the Git implementation


src/lando/api/tests/test_landings.py
```diff
@@ -777,29 +829,32 @@ def test_format_single_success_changed(
     ), "Autoformat via amending should only land a single commit."
 
 
+def _scm_get_previous_hash(scm: AbstractSCM) -> str:
+    if scm.scm_type() == SCM_TYPE_HG:
+        return HgSCM.run_hg(scm, ["log", "-r", "tip^", "-T", "{node}"]).decode("utf-8")
+    return GitSCM._git_run("rev-parse", "HEAD^", cwd=scm.path)
```
cgsheeh, line 835: We should have an `AbstractSCM.get_previous_hash` instead of this function which checks for the specific implementation.

src/lando/api/tests/test_landings.py
```diff
@@ -855,27 +906,32 @@ def test_format_stack_success_changed(
     ), "Autoformat commit has incorrect commit message."
 
 
+def _scm_get_last_commit_message(scm: AbstractSCM) -> str:
+    if scm.scm_type() == SCM_TYPE_HG:
+        return HgSCM.run_hg(scm, ["log", "-r", "tip", "-T", "{desc}"]).decode("utf-8")
+    return GitSCM._git_run("log", "--pretty=%B", "HEAD^..", cwd=scm.path)
```
cgsheeh, line 912: Same as above, these methods should go on the `AbstractSCM` and the implementations.

src/lando/main/tests/conftest.py
```diff
@@ -31,11 +36,24 @@ def git_setup_user():
...
         cwd=repo_dir,
     )
+
+
+def _git_ignore_denyCurrentBranch(repo_dir: Path):
+    """Disable error when pushing to this non-bare repo.
+
+    This is a sane protection in general, but it gets in the way of the tests here,
+    where we just want a target, and don't care much about the final state of this
+    target after everything is done."""
```
cgsheeh, line 54:
```suggestion
    target after everything is done.
    """
```

cgsheeh (changes requested):

src/lando/api/tests/test_landings.py
```diff
@@ -283,26 +347,26 @@ def test_integrated_execute_job(
...
-        name="mozilla-central",
-        url=hg_server,
-        required_permission=SCM_LEVEL_3,
-        push_path=hg_server,
-        pull_path=hg_server,
-        force_push=True,
-        system_path=hg_clone.strpath,
-    )
+    repo = repo_mc(repo_type)
+    repo.force_push = True
```
zzzeid, line 366: I think rather than manually overriding this here, it would be better to have a repo "factory" fixture that is used by all the repo fixtures above, and then you can just pass it additional overriding parameters if needed. That way, if we ever had any hooks or triggers that rely on either the database state of the repo, or things that happen upon repo creation, this type of override won't cause any problems.

This Repo object should be saved before proceeding, since the value is not going to be reflected in line 378 (when repos are fetched from the db).

src/lando/api/tests/test_landings.py
```diff
@@ -325,30 +389,30 @@ def test_integrated_execute_job_with_force_push(
...
-        name="mozilla-central",
-        url=hg_server,
-        required_permission=SCM_LEVEL_3,
-        push_path=hg_server,
-        pull_path=hg_server,
-        push_target="@",
-        system_path=hg_clone.strpath,
-    )
+    repo = repo_mc(repo_type)
+    repo.push_target = "@"
```
zzzeid, line 412: Same comment as above RE: repo state on creation / using a repo factory.

src/lando/api/tests/test_landings.py
```diff
@@ -519,27 +583,27 @@ def test_merge_conflict(
...
-        push_path=hg_server,
-        pull_path=hg_server,
-        approval_required=True,
-        autoformat_enabled=False,
-        system_path=hg_clone.strpath,
-    )
-
+    repo = repo_mc(repo_type)
+    repo.approval_required = True
+    repo.autoformat_enabled = False
```
zzzeid, line 605: Same comment as above -- ideally these are just passed as overrides to the fixture.

src/lando/api/tests/test_landings.py
```diff
@@ -685,42 +747,38 @@ def test_format_patch_success_unchanged(
...
-        name=tree,
-        url=hg_server,
-        push_path=hg_server,
-        pull_path=hg_server,
-        required_permission=SCM_LEVEL_3,
-        autoformat_enabled=True,
-        system_path=hg_clone.strpath,
-    )
+    repo = repo_mc(repo_type)
+    repo.autoformat_enabled = True
```
zzzeid, line 767: Same here.

src/lando/api/tests/test_landings.py
zzzeid: If these are only used in tests, I would hesitate to put them directly in the main implementation, but maybe they can be added in a mixin or a test class that extends `AbstractSCM`. Again, depends if they are only used in tests.

src/lando/api/tests/test_landings.py
```diff
@@ -777,29 +829,32 @@ def test_format_single_success_changed(
...
-        push_path=hg_server,
-        pull_path=hg_server,
-        required_permission=SCM_LEVEL_3,
-        autoformat_enabled=True,
-        system_path=hg_clone.strpath,
-    )
-
-    hgrepo = HgSCM(hg_clone.strpath)
+    repo = repo_mc(repo_type)
+    repo.autoformat_enabled = (True,)
```
zzzeid, line 855: Same issue here regarding adding these as overrides to a fixture, and how they are not saved.

src/lando/api/tests/test_landings.py
zzzeid: Same comment here as in `_scm_get_previous_hash` -- i.e., maybe those should be added to a mixin or test class.

src/lando/api/tests/test_landings.py
```diff
@@ -855,27 +906,32 @@ def test_format_stack_success_changed(
...
-        name=tree,
-        required_permission=SCM_LEVEL_3,
-        url=hg_server,
-        push_path=hg_server,
-        pull_path=hg_server,
-        autoformat_enabled=True,
-        system_path=hg_clone.strpath,
-    )
+    repo = repo_mc(repo_type)
+    repo.autoformat_enabled = True
```
zzzeid, line 933: Same here as other instance of this.

src/lando/api/tests/test_landings.py
```diff
@@ -913,26 +969,26 @@ def test_format_patch_fail(
...
-        name="mozilla-central",
-        required_permission=SCM_LEVEL_3,
-        url=hg_server,
-        push_path=hg_server,
-        pull_path=hg_server,
-        autoformat_enabled=True,
-        system_path=hg_clone.strpath,
-    )
+    repo = repo_mc(repo_type)
+    repo.autoformat_enabled = True
```
zzzeid, line 990: Same here as other instance of this.

src/lando/main/scm/git.py
```diff
@@ -229,7 +247,13 @@ def format_stack_amend(self) -> Optional[list[str]]:
 
     def format_stack_tip(self, commit_message: str) -> Optional[list[str]]:
         """Add an autoformat commit to the top of the patch stack."""
-        self._git_run("commit", "--all", "--message", commit_message, cwd=self.path)
+        try:
+            self._git_run("commit", "--all", "--message", commit_message, cwd=self.path)
+        except SCMException as exc:
+            if "nothing to commit, working tree clean" in exc.out:
+                return []
```
zzzeid, line 254: Is there a more rigorous way of doing this check? Is there a specific error code we could be checking when raising `SCMException`? Looking at your implementation of `_git_run`, looks like we are raising this exception specifically based on a return code, and perhaps we shouldn't be raising this as an exception at all from the get go.

src/lando/main/tests/conftest.py
```diff
@@ -1,11 +1,16 @@
-import pathlib
 import subprocess
+from pathlib import Path
 
 import pytest
 
 
 @pytest.fixture
-def git_repo(tmp_path: pathlib.Path):
+def git_repo_seed(request) -> Path:
```
zzzeid, line 8: Can you add a docstring here to explain what this is doing? Also, it doesn't look like `request` is being used.

src/lando/api/tests/conftest.py
```diff
@@ -30,8 +30,13 @@
 from lando.main.models import SCM_LEVEL_1, SCM_LEVEL_3, Profile, Repo
 from lando.main.scm import SCM_TYPE_HG
 from lando.main.support import LegacyAPIException
+from lando.main.tests.conftest import git_repo, git_repo_seed
 from lando.utils.phabricator import PhabricatorClient
 
+# We need some local usage of those imported fixtures to satisfy the linters.
+# This is it.
+__all__ = ["git_repo", "git_repo_seed"]
+
```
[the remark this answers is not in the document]
zzzeid: I am in favour of creating different conftests, including one that would be more available globally, but I'd hesitate to move tests from a more app-specific location to a less specific one. I think in the long term (once most of the modules are moved out of a "legacy" structure) it should be easier to contain tests and fixtures within their respective apps.

zzzeid (changes requested):

src/lando/api/tests/conftest.py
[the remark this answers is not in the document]
shtrom: I toyed with the idea of moving things out of legacy, but that seems like a risky gambit. If we cleared the legacy, I think the conftests would naturally coallesce into a single one, but that felt like too many moving pieces at once.

src/lando/api/tests/test_landings.py
shtrom: I had the same thought as @zzzeid here. Same for the other one that gets the whole commit message.

Both are only used in tests at the moment.

src/lando/main/scm/git.py
shtrom: The initial implementation had a conditional to allow throwing an exception or not, but I realised at the time that we always were throwing an exception, so I removed the parameter.

We could re-add it, but I can't find documentation about `git commit`'s error codes (at the moment it return `1` for this case), so we'd still have to inspect the error message to determine the cause of the error (FWIW, this is also how we do it for Mercurial) https://github.com/mozilla-conduit/lando/blob/main/src/lando/main/scm/hg.py#L372-L377

shtrom:
@zzzeid @cgsheeh test pass with the fix to test_hg. This should be good for one last look.

cgsheeh (approved):

src/lando/main/tests/conftest.py
```diff
@@ -5,7 +5,11 @@
 
 
 @pytest.fixture
-def git_repo_seed(request) -> Path:
+def git_repo_seed() -> Path:
+    """
+    Return the path to a diff file to apply on an empty repo to create a known base for
+    application of other patches as part of the tests.
```
zzzeid, line 11: First line of docstring should be a full description / sentence, with elaboration on other lines if needed.

zzzeid (changes requested):
nit: in first line of commit message: `test -> test_landings`.

PR #191, opened 2025-01-10
AbstractSCM: move process_merge_conflict to SCM (bug 1940876)

shtrom:
Builds on top of #190.


src/lando/api/tests/test_landings.py
```diff
@@ -935,8 +886,8 @@ def test_format_patch_fail(
 
     revisions = [
         create_patch_revision(1, patch=PATCH_FORMATTING_PATTERN_FAIL),
-        create_patch_revision(2),
-        create_patch_revision(3),
+        create_patch_revision(2, patch=normal_patch(0)),
+        create_patch_revision(3, patch=normal_patch(1)),
```
shtrom, line 890: Without an explicit `patch`, `NORMAL_PATCH_1` will always get used. Hg doesn't mind reapplying an existing patch, and just skips it silently, but Git is not so lenient, and fails.

Making sure the patches are different and part of the same series makes both SCMs happy for this test.

src/lando/api/tests/test_landings.py
```diff
@@ -991,8 +941,8 @@ def test_format_patch_no_landoini(
     treestatusdouble.open_tree(repo.name)
 
     revisions = [
-        create_patch_revision(1),
-        create_patch_revision(2),
+        create_patch_revision(1, patch=None),
```
shtrom, line 944: Explicitely setting the patch to `None` will make `create_patch_revision` use the revision number to determine which NORMAL_PATCH_* to use.

src/lando/api/tests/test_landings.py
```diff
@@ -565,7 +563,17 @@ def test_merge_conflict(
...
+    # Make sure a conflict actually happened.
+    assert repo_type in [
+        SCM_TYPE_GIT,
+        SCM_TYPE_HG,
+    ], f"Unsupported repository type: {repo_type=}"
+    if repo_type == SCM_TYPE_GIT:
+        assert "Rejected hunk" in caplog.text
+    elif repo_type == SCM_TYPE_HG:
+        assert "hunks FAILED" in caplog.text
+
```
cgsheeh, line 576: You should add an extra argument to this test's parameters which is the expected string to search for when a conflict is detected. That way you can remove the "Unsupported repository type" assertion and instead do `assert scm_conflict_string in caplog.text`.

src/lando/api/tests/test_landings.py
cgsheeh: Perhaps we should add this as a comment in the code for the benefit of some future version of us. :)

src/lando/main/scm/abstract_scm.py
```diff
@@ -102,6 +94,29 @@ def apply_patch(
...
+
+            revision_id: revision_id
+            failed paths: list[dict]
+                path: str
+                url: str
+                changeset_id: str
+            rejects_paths: dict[str, dict]
+                <str>: dict[str, str] (conflicted file path)
+                    path: str (reject file path)
+                    content: str
```
cgsheeh, line 117: Nit: this looks ripe for a `dataclass` implementation. ;)

cgsheeh (approved):
r- for the cleanup around the test parameters, otherwise this looks great.

src/lando/main/scm/git.py
```diff
@@ -163,7 +166,7 @@ def apply_patch(
             f_diff.flush()
 
             cmds = [
-                ["apply", f_diff.name],
+                ["apply", "--reject", f_diff.name],
```
zzzeid, line 169: It would be good to add a comment or something in the commit message about this parameter, and how it relates to `.rej` files.

src/lando/main/scm/git.py
```diff
@@ -177,7 +180,55 @@ def apply_patch(
...
-                self._git_run(*c, cwd=self.path)
+                try:
+                    self._git_run(*c, cwd=self.path)
+                except SCMException as exc:
+                    if "error: patch" in exc.err:
+                        raise PatchConflict(exc.err) from exc
+
+    def process_merge_conflict(
+        self,
+        exception: PatchConflict,
```
zzzeid, line 191: nit: I think it would be nice to rename this argument to something else other than `exception`.

src/lando/api/tests/test_landings.py
zzzeid: Agreed with this, also we don't need the first assert here, since `repo_type` is explicitly set in the parameters.

zzzeid (changes requested):
I think it would be good to add a test for

src/lando/api/tests/test_landings.py
shtrom: It's in the docstring of the fixture https://github.com/mozilla-conduit/lando/pull/191/files/315c3d469e73b2f2e50e738f4c5d8be0537b6e79#diff-5ae86d15f2e0c12ed398cedca1ad7cb0760e212cea88edbf9687e24f99366dc1R36-R40

Is there a better location where to move/duplicate it?

src/lando/api/tests/test_landings.py
shtrom: Ah, yep, nice!

src/lando/api/tests/test_landings.py
shtrom: Found a spot 😁

src/lando/main/scm/abstract_scm.py
shtrom: It... does doesn't it? I walked away from it due to the nestedness of the structure, and it being slightly off-track, but it would be nice.

https://bugzilla.mozilla.org/show_bug.cgi?id=1941245

src/lando/main/scm/git.py
shtrom: Yep. I found the passing the exception a bit iffy anyway. Turns out I can just extract the error message from the Exception in the landing worker, and pass a clean string here instead.

shtrom:
@zzzeid @cgsheeh test pass with the fix to test_hg. This should be good for one last look (bis repetitas).

src/lando/api/tests/test_landings.py
```diff
@@ -988,6 +981,8 @@ def test_format_patch_no_landoini(
     treestatusdouble.open_tree(repo.name)
 
     revisions = [
+        # Patch=None let create_patch_revision determin the patch to use based on the
```
zzzeid, line 984:
```suggestion
        # Patch=None lets create_patch_revision determine the patch to use based on the
```

zzzeid (approved):
Couple of typos otherwise good to go.

PR #195, opened 2025-01-14
test_hg: fix newline_bug test with Hg 6.9 (bug 1941011)

shtrom:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1541181

The fix for that bug added a `\r` in the file with no newline at end of file. Removing this from both the diff and the test repo created for the test fixes the failure.

Is removing the `\r` correct? Unclear, but a quick test with a fresh HG repo and a file with no newline at the end shows that no `\r` is added.


shtrom:
I don't think this is a correct fix due to this: https://bz.mercurial-scm.org/show_bug.cgi?id=6127#c2, where the `\r` is actually the issue.

shtrom:
Addressed differently in #192

PR #196, opened 2025-01-16
PushLog: model & recording support (bug 1940610, 1940611)

shtrom:
* pushlog: add basic models (bug 1940610)
* landing_worker: add PushLog support (bug 1940611)

Builds on #200


src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -234,131 +235,135 @@ def run_job(self, job: LandingJob) -> bool:
...
-                    LandingJobAction.FAIL,
-                    message=message + f"\n{e}",
-                )
-                self.notify_user_of_landing_failure(job)
-                return True
-
-            # Run through the patches one by one and try to apply them.
-            for revision in job.revisions.all():
-
+        with PushLogForRepo(repo, job.requester_email) as pushlog:
```
shtrom, line 238: This diff is better viewed with `Hide whitespace` enabled.

2025-01-30T16:30:12,436737036+11:00

src/lando/main/tests/test_git.py
```diff
@@ -139,21 +140,35 @@ def test_GitSCM_clean_repo(
     ), f"strip_non_public_commits not honoured for {new_file}"
 
 
-def test_GitSCM_push_get_github_token(git_repo: Path):
```
shtrom, line 142: Looks like I lost a test here. Need to be re-added.

src/lando/pushlog/models/push.py
```diff
@@ -0,0 +1,73 @@
+from django.db import models
+
+from lando.main.models import Repo
+
+from .commit import Commit
+from .consts import MAX_BRANCH_LENGTH, MAX_URL_LENGTH
+
+PUSH_SCM_TYPE_GIT = "git"
```
shtrom, line 8: We want to support HG, too, so need to add it.

shtrom: ... or maybe this is just not needed anymore...

src/lando/pushlog/models/push.py
```diff
@@ -0,0 +1,73 @@
...
+        auto_now_add=True,
+        db_index=True,
+    )
+
+    # Maximum total lengths are defined in RFC-5321 [0]: 64 for the local-part, and 255
+    # for the domain.
+    # [0] https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.1
+    user = models.EmailField(max_length=64 + 1 + 255)
+
+    commits = models.ManyToManyField(Commit)
```
shtrom, line 40: Would we want to keep actual SHA in this object? Same as we keep the repo_url?

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,116 @@
...
+class PushLog:
+    repo: Repo
+    push: Push
+    user: str
+
+    commits: list
+    # ManyToMany relationships that needs to be stored separately until Commits are
+    # save(), and the association can be made.
+    files: dict[str, list[str]]
+    parents: dict[str, list[str]]
```
shtrom, line 52: No longer needed.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if not files:
+            # Same a commits above.
+            files = {}
+        self.files = files
+
+        if not parents:
+            # Same a commits above.
+            parents = {}
+        self.parents = parents
```
shtrom, line 79: Those are stored in commits now; no longer needed.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,116 @@
...
+    def add_commit(self, scm_commit: SCMCommit) -> Commit:
+        # We create a commit object in memory, but will only write it into the DB when
+        # the whole push is done, and we have a transaction open.
+        commit = Commit.from_scm_commit(self.repo, scm_commit)
+        self.commits.append(commit)
+
+        # We can only attach parents and files to the Commit when it exists in the DB,
+        # so we hold them here in the meantime.
+        self.parents[scm_commit.hash] = scm_commit.parents
+        self.files[scm_commit.hash] = scm_commit.files
```
shtrom, line 95: No longer needed.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,116 @@
...
+        Don't use directly if using a PushLogForRepo.
+        """
+        push = Push.objects.create(repo=self.repo, user=self.user)
+        for commit in self.commits:
+            # We need to save the commit before we can associate files to it.
+            commit.save()
+
+            commit.add_parents(self.parents[commit.hash])
+            commit.add_files(self.files[commit.hash])
+            push.commits.add(commit)
```
shtrom, line 112: No longer needed, each commit looks after themselves now.

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,124 @@
...
+@pytest.fixture
+def make_commit(make_hash):
+    def commit_factory(repo: Repo, seqno: int, message=None) -> Commit:
+        "Create a non-descript commit with a sequence number in the test DB."
+        if not message:
+            message = f"Commit {seqno}"
+
+        return Commit.objects.create(
+            # Create a 40-character string, of which the first 8 bytes represent the seqno
+            # in decimal representation.
```
shtrom, line 41: Move this to `make_hash`

src/lando/pushlog/tests/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+
+    files = File.objects.filter(commit=commit2)
+    assert files.count() == 2
+    assert "file-1" in repr(files)
+    assert "file-3" in repr(files)
+
+
+@pytest.mark.django_db()
+def test__pushlog__models__Commit_unique(make_repo, make_commit):
+    # Model.objects.create() creates _and saves_ the object
```
shtrom, line 110: Move this comment to make_repo.

src/lando/pushlog/tests/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+    with pytest.raises(
+        IntegrityError,
+        match=r"duplicate key.*pushlog_commit_repo_id",
+    ):
+        make_commit(repo, 1)
+
+
+@pytest.mark.django_db()
+def test__pushlog__models__File_unique(make_repo, make_file):
+    # Model.objects.create() creates _and saves_ the object
```
shtrom, line 123: Ditto.

src/lando/pushlog/tests/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+
+    assert tag1.name in repr(tag1)
+
+    assert retrieved_tags.count() == 2
+    assert rtag1.id == tag1.id
+
+
+@pytest.mark.django_db()
+def test__pushlog__models__Tag_unique(make_repo, make_commit, make_tag):
+    # Model.objects.create() creates _and saves_ the object
```
shtrom, line 153: Ditto.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,123 @@
...
+        self,
+        repo: Repo,
+        user: str,
+        commits: list = [],
+    ):
+        self.repo = repo
+        self.user = user
+
+        if not commits:
+            # We cannot us the default value of the argument as a mutable type, and will
```
zzzeid, line 63:
```suggestion
            # We cannot use the default value of the argument as a mutable type, and will
```

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,123 @@
...
+
+    confirmed: bool = False
+
+    commits: list
+
+    def __init__(
+        self,
+        repo: Repo,
+        user: str,
+        commits: list = [],
```
zzzeid, line 57: There's a Python gotcha here, should not specify a mutable argument ([]), which you hint at below.

```suggestion
        commits: list = None,
```

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,123 @@
...
+        return commit
+
+    def remove_tip_commit(self) -> Commit:
+        """Remove the tip commit from the PushLog, returning it."""
+        tip_commit = self.commits[-1]
+        self.commits.remove(tip_commit)
+        return tip_commit
+
+    def confirm(self, value: bool = True):
+        self.confirmed = value
```
zzzeid, line 89: Docstring missing. Also, the name is a little confusing. It should be something like `set_confirmed`, or `confirm` should always set `self.confirmed` to `True`.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,123 @@
...
+            # get reused on every initialisation.
+            commits = []
+        self.commits = commits
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}({self.repo!r}, {self.user}, {self.commits!r})"
+        )
+
+    def add_commit(self, scm_commit: SCMCommit) -> Commit:
```
zzzeid, line 73: Docstring missing.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+            raise e
+
+
+class PushLog:
+    repo: Repo
+    push: Push
+    user: str
+
+    confirmed: bool = False
```
zzzeid, line 49: nit

```suggestion
    is_confirmed: bool = False
```

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,123 @@
...
+        commit = Commit.from_scm_commit(self.repo, scm_commit)
+        self.commits.append(commit)
+
+        return commit
+
+    def remove_tip_commit(self) -> Commit:
+        """Remove the tip commit from the PushLog, returning it."""
+        tip_commit = self.commits[-1]
+        self.commits.remove(tip_commit)
+        return tip_commit
```
zzzeid, line 86: I think it would be cleaner to not have to remove a "tip commit" in the pushlog, and instead do that before collecting the pushlog commits.

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+from lando.main.models import Repo
+from lando.main.scm.commit import Commit as SCMCommit
+from lando.pushlog.models import Commit, File, Push, Tag
+
+
+@pytest.fixture
+def make_repo():
+    def repo_factory(seqno: int) -> Repo:
+        "Create a non-descript repository with a sequence number in the test DB."
+        # Model.objects.create() creates _and saves_ the object
```
zzzeid, line 14: nit: This is self-evident, not sure we have to add a comment about it.

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+            default_branch=f"main-{seqno}",
+        )
+
+    return repo_factory
+
+
+@pytest.fixture
+def make_hash():
+    """Create a 40-character string, of which the first 8 bytes represent the seqno
+    in decimal representation."""
```
zzzeid, line 28: First line in docstring should be self-sufficient.

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+    def hash_factory(seqno: int):
+        return str(seqno).zfill(8) + "f" + 31 * "0"
+
+    return hash_factory
+
+
+@pytest.fixture
+def make_commit(make_hash):
+    def commit_factory(repo: Repo, seqno: int, message=None) -> Commit:
+        "Create a non-descript commit with a sequence number in the test DB."
```
zzzeid, line 39:
```suggestion
        """Create a non-descript commit with a sequence number in the test DB."""
```

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+            datetime=datetime.now(),
+        )
+
+    return commit_factory
+
+
+@pytest.fixture
+def make_file():
+    def file_factory(repo: Repo, seqno: int) -> File:
+        "Create a non-descript file with a sequence number in the test DB."
```
zzzeid, line 57:
```suggestion
        """Create a non-descript file with a sequence number in the test DB."""
```

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+            name=f"file-{seqno}",
+        )
+
+    return file_factory
+
+
+@pytest.fixture
+def make_tag():
+    def tag_factory(repo: Repo, seqno: int, commit: Commit) -> Tag:
+        "Create a non-descript tag with a sequence number in the test DB."
```
zzzeid, line 69:
```suggestion
        """Create a non-descript tag with a sequence number in the test DB."""
```

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+            commit=commit,
+        )
+
+    return tag_factory
+
+
+@pytest.fixture
+def make_push():
+    def push_factory(repo: Repo, commits: list[Commit]):
+        "Create a non-descript push containing the associated commits in the test DB."
```
zzzeid, line 82:
```suggestion
        """Create a non-descript push containing the associated commits in the test DB."""
```

zzzeid (changes requested):
First pass. I think it would be easier to review this once you merge the new changes in `main` since there is some old code kicking around.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -235,139 +237,143 @@ def run_job(self, job: LandingJob) -> bool:
...
 
-                date = patch_helper.get_header("Date")
-                user = patch_helper.get_header("User")
+                # Run through the patches one by one and try to apply them.
+                logger.warning(
+                    f"About to land {job.revisions.count()} revisions: {job.revisions.all()} ..."
+                )
+                for revision in job.revisions.all():
+                    try:
+                        logger.warning(f"Landing {revision} ...")
```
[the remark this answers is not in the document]
shtrom: Ah, yep. I wanted to make sure I was getting the messages when debugging, but that's a bit overkill for prod use 😅

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -235,139 +237,143 @@ def run_job(self, job: LandingJob) -> bool:
...
                     )
                     self.notify_user_of_landing_failure(job)
                     return True
 
-                date = patch_helper.get_header("Date")
-                user = patch_helper.get_header("User")
+                # Run through the patches one by one and try to apply them.
+                logger.warning(
+                    f"About to land {job.revisions.count()} revisions: {job.revisions.all()} ..."
+                )
```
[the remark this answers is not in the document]
shtrom: Ah, I did add a `DJANGO_LOG_LEVEL=debug` in my `.env`, but that didn't seem to increase the verbosity.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -235,139 +237,143 @@ def run_job(self, job: LandingJob) -> bool:
...
+                for revision in job.revisions.all():
+                    try:
+                        logger.warning(f"Landing {revision} ...")
+                        scm.apply_patch(
+                            revision.diff,
+                            revision.commit_message,
+                            revision.author,
+                            revision.timestamp,
+                        )
+                        new_commit = scm.describe_commit()
```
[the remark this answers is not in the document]
shtrom: I'll put in in an `else`. I had misgiving in doing so before, as it was moving this single line quite further away, below all the `except`s from the one that creates the commit.

I'm now wondering if I can get away with the `else` before all the `except`. I'll try that.

...

Nope, the `else` must be after all the `except`s and `finally` ):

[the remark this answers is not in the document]
shtrom: ... Which I already had anyway 🤦🏼

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -235,139 +237,143 @@ def run_job(self, job: LandingJob) -> bool:
...
+                        push_target=repo.push_target,
+                        force_push=repo.force_push,
                     )
-                except PatchConflict as exc:
-                    breakdown = scm.process_merge_conflict(
-                        repo.pull_path, revision.revision_id, str(exc)
-                    )
-                    job.error_breakdown = breakdown
-
+                    pushlog.confirm()
```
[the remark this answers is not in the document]
shtrom: Do you mean the pushlog should have a record of failed pushes, too?

src/lando/main/scm/abstract_scm.py
```diff
@@ -3,6 +3,8 @@
 from pathlib import Path
 from typing import Any, ContextManager, Optional
 
+from .commit import Commit
```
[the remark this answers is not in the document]
shtrom: I don't think we can do this, as the `abstract_scm` is part of the `scm` module, so we'd have a circular dependency.

I can do a `from lando.main.scm.commit import Commit`

src/lando/main/scm/git.py
```diff
@@ -230,6 +233,41 @@ def process_merge_conflict(
 
         return breakdown
 
+    def describe_commit(self, revision_id: str = "HEAD") -> Commit:
+        """Return Commit metadata."""
+        separator = "".join(
+            [chr(c) for c in random.choices(range(ord("A"), ord("Z")), k=16)]
+        )
```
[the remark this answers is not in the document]
shtrom: It's used when parsing the semi-structured output of the SCM status messages, which may include multi-line strings (e.g., the commit message).

They're also uncontrolled strings, so we can't use a simple separator for fear of it actually occurring in some of the strings.

So I ended up going for a long random string which, hopefully, will never appear in any output. This is kinda similar to the way MIME does it for multipart email messages.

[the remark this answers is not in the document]
shtrom: (moved to a utility method in the abstract_scm)

src/lando/main/scm/hg.py
```diff
@@ -317,6 +324,41 @@ def process_merge_conflict(
...
+        )
+
+        output = self.run_hg(["log", "-r", revision_id, "-T", format]).decode("utf-8")
+        parts = re.split(f"{separator}", output)
+        metadata: dict[str, Any] = dict(p.split(":", 1) for p in parts)
+
+        metadata["parents"] = metadata["parents"].split()
+        # {parents} in Mercurial is empty if the commit has a single parent.
+        # We re-add it manually, but only if it is a non-null parent.
+        if not metadata["parents"] and not metadata["parent"] == 40 * "0":
```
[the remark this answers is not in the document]
shtrom: It's very literally 40 times the character "0". Python arithmetic 🤷🏼 

But... Yeah, it's essentially an empty parent.

src/lando/pushlog/management/commands/view_pushlog.py
```diff
@@ -0,0 +1,37 @@
...
+            repo = Repo.objects.get(name=repo_name)
+        except Repo.DoesNotExist:
+            raise CommandError(f"Repository not found: {repo_name}")
+
+        pushes = Push.objects.filter(repo=repo).order_by("-push_id")
+        if limit:
+            pushes = pushes[:limit]
+
+        for push in pushes:
+            print(push)
```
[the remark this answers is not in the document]
shtrom: This should be removed, and whoever wrote it punished for leaving debug code around 😉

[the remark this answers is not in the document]
shtrom: Correction: yes.

src/lando/pushlog/models/push.py
```diff
@@ -0,0 +1,81 @@
...
+        max_push_id = (
+            cls.objects.filter(repo=repo)
+            .order_by("-push_id")
+            .values_list("push_id", flat=True)
+        )
+
+        if max_push_id:
+            return max_push_id[0] + 1
+
+        return 1
```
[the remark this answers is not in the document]
shtrom: We need push_ids scoped to a specific repo. So we can't use the natural primary key in the DB as the ID, and need to maintain it separately.

This essentially grabs the last push_id for the repo and +1s it.

It's not great, but I couldn't find another way to do it. I looked at using sequences, but we'd have needed one sequence per repo, which seemed overkill. I'm open to other suggestions!

src/lando/pushlog/models/push.py
shtrom: A commit can be pushed to multiple branches independently, so I think we may find a commit in more than one push.

What I meant with the SHA question was whether we should use the SHA as a database key, rather than an additional internal column. The pro is that the push table on its own would contain useful information to reconcile commits, even with incomplete information. The con are that the column will be 160bit, and that it would diverge from standard DB models. Thinking more about it, it's probably getting too corner case to be worth it.

src/lando/pushlog/models/commit.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+class File(models.Model):
+    name = models.CharField(max_length=MAX_PATH_LENGTH)
+
+    repo = models.ForeignKey(
+        Repo,
+        # We don't want to delete the PushLog, even if we were to delete the repo
+        # object.
+        on_delete=models.DO_NOTHING,
```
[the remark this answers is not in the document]
shtrom: I did it, then reverted it.

The model doesn't allow NULL there to start with, which I don't really want to retract.

In addition, if we were to delete a repo and setting this to NULL, we'd get a large number of dangling entities which could no longer be grouped by repo. If we just DO_NOTHING, the reference to the Repo will obviously be broken, but we'd still be able to group everything together by (deleted) repo

There are also UNIQUE constraints that include the `repo` column, which means that if we delete two repos that, say, have the same, say, file and set the repo to NULL, we would start getting uniqueness constraint violations.

So, I think keeping DO_NOTHING here is the best option: we should still be able to delete a repo, but the existing pushlog entries for this repo will remain untouched.

src/lando/pushlog/pushlog.py
shtrom: Ah! Yes. I handle the problem just below, by creating a fresh new list https://github.com/mozilla-conduit/lando/pull/196/files/4a97d5c23c741a958303178ff3252d37838b744d#diff-b518a05517dedc31aceb83ab380e3bd35de03582d5fc1fa528c716e639882463R62-R65

I'm reluctant to use `None` in the prototype, as this would require to make it an `Optional[list]`, unless you insist?

src/lando/pushlog/tests/conftest.py
shtrom: I wasn't aware of it before I discovered it (:

But yes, this is pretty standard, and documented, Django.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -235,139 +237,143 @@ def run_job(self, job: LandingJob) -> bool:
...
-            # Run through the patches one by one and try to apply them.
-            for revision in job.revisions.all():
-                # TODO: Rather than parsing the patch details from the full HG patch
-                # stored in the job, we should read the revision's metadata (and
-                # move to only store the diff in the patch_string, rather than an
-                # export).
-                # https://bugzilla.mozilla.org/show_bug.cgi?id=1936171
-                patch_helper = HgPatchHelper(StringIO(revision.patch_string))
-                if not patch_helper.diff_start_line:
+        with PushLogForRepo(repo, job.requester_email) as pushlog:
```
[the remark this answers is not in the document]
shtrom: I added an option to disable the pushlog per repo.

shtrom:
@cgsheeh lando pushlog PR

shtrom:
@zzzeid @cgsheeh good for another look!

zzzeid:
> @zzzeid @cgsheeh good for another look!

Thanks will take a look at this today.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,144 @@
...
+@contextmanager
+def PushLogForRepo(repo: Repo, user: str):
+    """
+    Context manager allowing to incrementally build push information, and only submit it
+    when complete.
+
+    WARNING: Do not use record_push() on the returned PushLog, as the context manager
+    will take care of it automatically. Calling it separately may lead to duplicates.
+    """
+    if repo.pushlog_disabled:
```
cgsheeh, line 24: Nit: this could be much flatter if we did and early `return` in the `pushlog_disabled` path. Also, we could run the `record_push` outside of the `else` clause since it runs on success regardless.

cgsheeh (dismissed):
Looking very good! Had a few nits, a question or two and a few typos to correct. Nothing major on my end.

src/lando/pushlog/models/commit.py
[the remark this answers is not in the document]
zzzeid: To reword "if we don't want to restrict the ability to delete repos" -- I think [DO_NOTHING](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.DO_NOTHING) will trigger an `IntegrityError` if you try to delete a referenced foreign key that is in use. So in effect, you won't be able to delete a repo in that case.

src/lando/main/scm/hg.py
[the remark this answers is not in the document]
zzzeid: Aware of the arithmetic aspect, "empty parent" answers my question :)

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -382,6 +224,207 @@ def run_job(self, job: LandingJob) -> bool:
...
+        # Get the changeset hash of the first node.
+        commit_id = scm.head_ref()
+
+        repo_push_info = f"tree: {repo.tree}, push path: {repo.push_path}"
+        try:
+            # We need to add the commits to the pushlog _before_ pushing, so we can
+            # compare the current stack to the last upstream.
+            # We'll only confirm them if all succeeds.
+            for commit in scm.describe_local_changes():
+                pushlog.add_commit(commit)
```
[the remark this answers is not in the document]
zzzeid: +1 to this, ideally there is only one line in the `try` block.

src/lando/main/scm/git.py
```diff
@@ -240,12 +239,17 @@ def process_merge_conflict(
...
 
     def describe_commit(self, revision_id: str = "HEAD") -> Commit:
-        """Return Commit metadata."""
-        separator = "".join(
-            [chr(c) for c in random.choices(range(ord("A"), ord("Z")), k=16)]
-        )
-        format = separator.join(
+        return self._describe_commits(revision_id)[0]
+
+    def describe_local_changes(self) -> list[Commit]:
```
zzzeid, line 244: Docstring missing.

src/lando/main/tests/conftest.py
```diff
@@ -33,8 +33,14 @@ def git_repo(tmp_path: Path, git_repo_seed: Path) -> Path:
...
-    subprocess.run(["git", "show"], check=True, cwd=repo_dir)
-    subprocess.run(["git", "branch"], check=True, cwd=repo_dir)
+
+    # Create a separate base branch for branch tests.
+    subprocess.run(["git", "checkout", "-b", "dev"], check=True, cwd=repo_dir)
+    subprocess.run(
+        ["git", "commit", "--allow-empty", "-m", "dev"], check=True, cwd=repo_dir
+    )
+
+    subprocess.run(["git", "checkout", "main"], check=True, cwd=repo_dir)
```
zzzeid, line 43: nit

```suggestion
    for command in (                                                                                                                                                                                                  
        ["git", "checkout", "-b", "dev"],
        ["git", "commit", "--allow-empty", "-m", "dev"],
        ["git", "checkout", "main"],
    ):
        subprocess.run(command, check=True, cwd=repo_dir)
```

src/lando/pushlog/pushlog.py
```diff
@@ -21,46 +21,49 @@ def PushLogForRepo(repo: Repo, user: str):
     WARNING: Do not use record_push() on the returned PushLog, as the context manager
     will take care of it automatically. Calling it separately may lead to duplicates.
```
zzzeid, line 22: Wondering if there's a way to detect/enforce this.

zzzeid (dismissed):
Few small comments otherwise looks good.

src/lando/main/tests/conftest.py
shtrom: Fixed where the code moved.

src/lando/pushlog/pushlog.py
shtrom: Ah, yep! I added a boolean to record the recording, and I can throw a RuntimeException if we do it twice.

src/lando/main/scm/abstract_scm.py
```diff
@@ -195,3 +207,11 @@ def format_stack_tip(self, commit_message: str) -> Optional[list[str]]:
...
         Returns a list containing a single string representing the ID of the newly created commit.
         """
+
+    @staticmethod
+    def _separator():
+        """Generate a long random string usable as a separator when parsing
+        semi-structured multiline text output"""
+        return "".join(
+            [chr(c) for c in random.choices(range(ord("A"), ord("Z")), k=16)]
+        )
```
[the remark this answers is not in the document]
shtrom: Neat!

src/lando/pushlog/models/push.py
shtrom: Removed.

src/lando/pushlog/models/commit.py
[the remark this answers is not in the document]
shtrom: Yep, you're right.

```
IntegrityError at /admin/main/repo/

update or delete on table "main_repo" violates foreign key constraint "pushlog_file_repo_id_abc97f4d_fk_main_repo_id" on table "pushlog_file"
DETAIL:  Key (id)=(2) is still referenced from table "pushlog_file".
```

I'll use `SET_NULL`, but I hope we never get to this point 😉

src/lando/pushlog/models/commit.py
```diff
@@ -0,0 +1,246 @@
...
+
+        return Commit(
+            repo=repo,
+            hash=scm_commit.hash,
+            author=scm_commit.author,
+            datetime=scm_commit.datetime,
+            desc=scm_commit.desc,
+            parents=scm_commit.parents,
+            files=scm_commit.files,
+        )
```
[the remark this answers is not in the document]
shtrom: The goal of the dataclass is specifically to be free of the ORM dependency. It's an SCM-only model.

That said, it would make sense if the ORM object where a descendent of the dataclass.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -224,131 +183,14 @@ def run_job(self, job: LandingJob) -> bool:
...
-                    message=message + f"\n{e}",
-                )
-                self.notify_user_of_landing_failure(job)
-                return True
-
-            # Run through the patches one by one and try to apply them.
-            for revision in job.revisions.all():
-
+        with PushLogForRepo(repo, job.requester_email) as pushlog:
+            with scm.for_push(job.requester_email):
```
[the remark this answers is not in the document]
shtrom: ah! nice!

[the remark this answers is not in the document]
shtrom: FWIW, I also looked at putting the PushLog into the `for_push` context manager, to make it more transparent. Unfortunately, it started mixing Django DB objects into the SCM classes, so I preferred to backtrack.

shtrom:
@zzzeid @cgsheeh good for another look!

shtrom:
I pushed this branch to dev https://github.com/mozilla-conduit/lando/actions/runs/13888718562

Waiting for after the migration to new lando to merge.

src/lando/pushlog/pushlog.py
zzzeid: It's very conventional to not set a `[]` and instead set `None`. I think it's also better not to add a workaround to satisfy type hint annotations.

Lastly, passing an empty list here would also trigger lines 66-69, so I think it is better all around to set it to `None` and always set the default within the function.

src/lando/main/migrations/0019_repo_pushlog_disabled.py
zzzeid, line 1: Migration needs to be redone here.

zzzeid (dismissed):
Few minor comments.

src/lando/main/tests/conftest.py
```diff
@@ -51,14 +53,20 @@ def git_setup_user():
...
+    _run_commands(
+        [
+            ["git", "config", "user.name", "Py Test"],
+            ["git", "config", "user.email", "pytest@lando.example.net"],
+        ],
+        str(repo_dir),
     )
 
 
+def _run_commands(commands: list[list[str]], cwd: str):
```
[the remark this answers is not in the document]
shtrom: This is an excellent question. I'm glad you asked.

😅

src/lando/pushlog/models/commit.py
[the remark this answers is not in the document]
cgsheeh: What is the limitation the the ORM dependency that stops us from using it in cases where the dataclass version is used instead?

Side-nit: the two objects are both named `Commit`, a name change like `CommitData` for the dataclass version would make it easier to follow which object is used where.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,150 @@
...
+        # We keep a record of the Pushlog in the extra, in addition to printing
+        # details in the log.
+        logger.error(
+            f"Failed to record push log due to: {e}\n{pushlog}",
+            extra={"pushlog": pushlog},
+        )
+        raise e
+
+
+class PushLog:
```
cgsheeh, line 47: We should add a docstring for this.

src/lando/pushlog/pushlog.py
```diff
@@ -0,0 +1,150 @@
...
+    repo: Repo
+    push: Push
+    user: str
+
+    is_confirmed: bool = False
+    is_recorded: bool = False
+
+    commits: list
```
cgsheeh, line 55: `list[Commit]`?

src/lando/pushlog/tests/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+from lando.pushlog.models import Commit, File, Push, Tag
+
+
+@pytest.fixture
+def make_repo():
+    def repo_factory(seqno: int) -> Repo:
+        """Create a non-descript repository with a sequence number in the test DB."""
+        return Repo.objects.create(
+            name=f"repo-{seqno}",
+            scm_type="git",
```
cgsheeh, line 16: We should use `SCM_TYPE_GIT` here.

cgsheeh (approved):
Meta-nit: please don't use single letter variable names. :sweat_smile:

LGTM, I had a few small things I pointed out and a question about the `Commit` dataclass. I do think we should rename that class since we have two classes with the same names and very similar fields.

src/lando/pushlog/models/commit.py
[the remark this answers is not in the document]
shtrom: It's not a limitation as much as trying to keep things separated. The SCM object at the moment have no dependency on anything related to the ORM, which I'd rather keep this way.

The ORM Commit only comes in as it's necessary for the Pushlog (not the main app), so it feels like it's muddying up the waters a bit to loop that back as a dependency.

That said, I'm not a fan of the duplication either... I'll see how I can flatten this a bit.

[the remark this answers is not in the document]
shtrom: Hum, yeah, that's not trivial. There's a lot that goes into a Django ORM class beyond just attribute names. I'm pretty sure they still duck-type the same, though.

But I've renamed `scm.Commit` to `scm.CommitData`, so it's less confusing.

zzzeid (approved):
lgtm!

PR #197, opened 2025-01-17
diff_warnings: port functionality (bug 1941342)

zzzeid:
- port diff warnings endpoint handling
- unskip tests
- update tests, views, forms to work with new implementation
- rename diff_warning.error_breakdown back to diff_warning.data


src/lando/api/views.py
```diff
@@ -1 +1,111 @@
...
+from lando.main.models.revision import DiffWarning, DiffWarningStatus
+from lando.utils.phabricator import get_phabricator_client
+
+
+def phabricator_api_key_required(func: callable) -> Callable:
+    """A simple wrapper that checks for a valid Phabricator API token."""
+
+    @wraps(func)
+    def _wrapper(self, request, *args, **kwargs):
+        HEADER = "X-Phabricator-Api-Key"
```
shtrom, line 17: Should it be `HTTP_X_Phabricator_API_Key` now, as in the test fixtures? https://github.com/mozilla-conduit/lando/pull/197/files#diff-3f084e2a5f612fc35bc5c68a611d1ad8959f632315c6e9115149298061dcc371R15

src/lando/api/views.py
```diff
@@ -1 +1,111 @@
...
+        """
+        # TODO: validate whether revision/diff exist or not.
+        form = self._get_form(request.method, request.POST)
+        if form.is_valid():
+            data = form.cleaned_data
+            warning = DiffWarning(**data)
+            warning.save()
+            return JsonResponse(warning.serialize(), status=201)
+        else:
+            return JsonResponse({"errors": dict(form.errors)}, status=400)
```
shtrom, line 88:
```suggestion

        return JsonResponse({"errors": dict(form.errors)}, status=400)
```

src/lando/api/views.py
```diff
@@ -1 +1,111 @@
...
+    def get(self, request, **kwargs):
+        """Return a list of active revision diff warnings, if any."""
+        form = self._get_form(request.method, request.GET)
+        if form.is_valid():
+            warnings = DiffWarning.objects.filter(**form.cleaned_data).all()
+            return JsonResponse(
+                [warning.serialize() for warning in warnings], status=200, safe=False
+            )
+        else:
+            return JsonResponse({"errors": dict(form.errors)}, status=400)
```
shtrom, line 111:
```suggestion

        return JsonResponse({"errors": dict(form.errors)}, status=400)
```

src/lando/main/models/revision.py
```diff
@@ -146,10 +146,10 @@ class DiffWarning(BaseModel):
...
         return {
-            "id": self.id,
             "diff_id": self.diff_id,
+            "group": self.group,
+            "id": self.id,
             "revision_id": self.revision_id,
-            "status": self.status.value,
-            "group": self.group.value,
+            "status": self.status,
             "data": self.data,
```
shtrom, line 154: Nit: we could sort that (:

shtrom (approved):
Couple of non-functional suggestions, and one question about the `HTTP_X_Phabricator_API_Key` header, but LGTM otherwise.

If the questions is irrelevant, then I think it's good to merge.

src/lando/api/views.py
```diff
@@ -1 +1,111 @@
...
+class LegacyDiffWarningView(View):
+    """
+    This class provides the API controllers for the legacy `DiffWarning` model.
+
+    These API endpoints can be used by clients (such as Code Review bot) to
+    get, create, or archive warnings.
+    """
+
+    @classmethod
+    def _get_form(cls, method: str, data: dict):
```
cgsheeh, line 42: Nit: this function is harder to follow than if we directly build the form in the corresponding handler.

src/lando/api/views.py
```diff
@@ -1 +1,111 @@
...
+
+        Returns:
+            dict: a dictionary representation of the object that was created.
+        """
+        # TODO: validate whether revision/diff exist or not.
+        form = self._get_form(request.method, request.POST)
+        if form.is_valid():
+            data = form.cleaned_data
+            warning = DiffWarning(**data)
+            warning.save()
```
cgsheeh, line 85: Nit: `DiffWarnings.objects.create()` does these two lines in a single step.

cgsheeh (approved):

src/lando/api/views.py
zzzeid: These are case-insensitive, but good to change it for consistency .

zzzeid:
Comments addressed, as well as fixed a couple of issue with the tests to make them more consistent with how the endpoints are actually used. Lastly, added csrf_exempt to the endpoints since we do not need to check the token for these endpoints (used by external service and are authenticated with a phabricator token).

src/lando/api/views.py
zzzeid: Moved forms so they are in separate classes and in each handler.

src/lando/api/tests/test_diff_warnings.py
```diff
@@ -6,84 +8,93 @@
...
+def test_diff_warning_create_bad_request_no_message(client, phab_header):
     """Ensure a request with incorrect data returns an error."""
     response = client.post(
-        "/diff_warnings",
-        json={"revision_id": 1, "diff_id": 1, "group": "LINT", "data": {}},
-        headers=phab_header,
+        "/api/diff_warnings/",
+        data={"revision_id": 1, "diff_id": 1, "group": "LINT", "data": {}},
+        content_type="application/json",
+        **phab_header,
```
shtrom, line 41: Nit: according to https://docs.djangoproject.com/en/5.0/topics/testing/tools/#making-requests, 

> Keyword arguments starting with a `HTTP_` prefix are set as headers, but the `headers` parameter should be preferred for readability.

Both approaches work, so 🤷🏼‍♂️

PR #198, opened 2025-01-20
tests: handle skipped tests (bug 1897504)

zzzeid:
- fix landing job cancel tests
- remove unneeded phabricator exception test
- fix bug in JS code that called endpoint without trailing slash


shtrom (approved):
LGTM!

PR #200, opened 2025-01-23
landing_worker: add and use metadata from Revision rather than HgPatchHelper (bug 1936171)

shtrom:
The ultimate goal of this change is to deprecate the back-and-forth use
of the HgPatchHelper. This first change allows to confine the parsing of
an Hg patch to the Revision object, which allows exposing cleaned-up
attribute to the rest of the system.


shtrom:
@cgsheeh this could be useful for the automation API #194, so you don't have to mess around with the HgPatchHelper when creating the Revisions.

src/lando/main/models/revision.py
```diff
@@ -104,6 +115,89 @@ def serialize(self) -> dict[str, Any]:
...
+        return metadata.get("commit_message")
+
+    @property
+    def timestamp(self) -> Optional[str]:
+        metadata = self._parse_metadata_from_patch()
+        if ts := metadata.get("timestamp"):
+            # Some codepaths (via Phabricator) have the timestamp set as an int.
+            # We make sure it's always a string.
+            return str(ts)
+        return None
```
zzzeid, line 151: Feels a bit repetitive to call `_parse_metadata_from_patch` in each method. Maybe all these values should be determined at the same time, whenever `self.patch_data` is set.

src/lando/main/models/revision.py
```diff
@@ -104,6 +115,89 @@ def serialize(self) -> dict[str, Any]:
...
+
+    @property
+    def patch_helper(self) -> HgPatchHelper:
+        """Create and cache an HgPatchHelper to parse the raw patch with Hg metadata."""
+        if not self._patch_helper:
+            self._patch_helper = HgPatchHelper(StringIO(self.patch))
+            if not self._patch_helper.diff_start_line:
+                raise NoDiffStartLine
+
+        return self._patch_helper
```
zzzeid, line 199: These two methods are hg-specific, wonder if we should be making them more scm-agnostic?

zzzeid (changes requested):
Couple of small comments, otherwise looks good!

src/lando/main/models/revision.py
shtrom: Yes, the purpose of this change is a stop gap to start hiding the HgPatchHelper.

Historically, we transformed Phab revision into a Hg Patch, which we no longer need to do.

In this first cut, we hide the patch helper from everything else, and let the interface of the Revision be sufficient, and we can change the implementation later on.

src/lando/main/models/revision.py
shtrom: The metadata is only parsed one, and cached, as part of the `_parse_metadata_from_patch()` method https://github.com/mozilla-conduit/lando/pull/200/files#diff-6f00d8fe4977345e9360a87afcf4861324a3b6ff012d6a6a5716640b4b44c04dR153-R168

But calling it in every method makes sure we call it the first time we ever need any metadata.

zzzeid: The repetitiveness aspect (i.e., calling `_parse_metadata_from_patch` in every method) was more so what I meant, not so much efficiency. I think changing `_parse_metadata_from_patch` to a property (e.g., `_parsed_metadata_from_patch`) and then using it directly in these methods would alleviate this issue.

This would become:


```python
    @property
    def timestamp(self) -> Optional[str]:
        if ts := self._parsed_metadata_from_patch.get("timestamp"):
            # Some codepaths (via Phabricator) have the timestamp set as an int.
            # We make sure it's always a string.
            return str(ts)
        return None
```

shtrom: Ah! Yeah, that's better.

zzzeid (approved):
nit: the first commit message line looks a bit too long and might get clipped when landing.

PR #201, opened 2025-01-23
landing_worker: various changes to configuration (bug 1940565)

zzzeid:
This PR mainly adds the ability to configure landing workers on-the-fly without
the need to restart them after changes.

- implement functionality to update worker config on the fly
- remove configuration variables that relate to workers
- pass worker instance to base worker
- implement functionality in worker to fetch up to date config
- rename fields for consistency
- update tests and fixtures
- add improved string representation for workers


src/lando/api/legacy/workers/base.py
```diff
@@ -135,12 +125,19 @@ def _start(self, max_loops: int | None = None, *args, **kwargs):
         # NOTE: The worker will exit when max_loops is reached, or when the stop
         # variable is changed to True.
         loops = 0
+        sleep_seconds = self.worker_instance.sleep_seconds
```
zzzeid, line 128: This does not get updated on-the-fly, should probably move it to be in the loop.

src/lando/api/tests/test_transplants.py
```diff
@@ -709,17 +708,19 @@ def test_integrated_transplant_records_approvers_peers_and_owners(
...
-    )
+    repo = Repo.objects.get(name="mozilla-central")
+
+    repo.url = hg_server
+    repo.required_permission = SCM_LEVEL_3
+    repo.push_path = hg_server
+    repo.pull_path = hg_server
+    repo.system_path = hg_clone.strpath
+
+    repo.save()
```
shtrom, line 722: Nit: We can probably reuse the `repo_mc` fixture here.

src/lando/api/tests/test_worker.py
```diff
@@ -4,14 +4,18 @@
...
 
 @mock.patch.dict(os.environ, {LandingWorker.SSH_PRIVATE_KEY_ENV_KEY: ""})
 @pytest.mark.django_db
 def test_Worker__no_SSH_PRIVATE_KEY(caplog):
     # The worker will read the environment and try to handle the SSH_PRIVATE_KEY if
     # present.
-    w = LandingWorker([], with_ssh=True)
+    worker_instance = Worker.objects.create(
+        name="test-worker",
+    )
```
shtrom, line 17: Nit: Could we use the `landing_worker_instance` fixture here?

src/lando/main/models/worker.py
```diff
@@ -27,7 +27,16 @@ class Worker(BaseModel):
     )
 
     def __str__(self):
-        return self.name
+        if self.is_paused:
+            state = "PAUSED"
+        elif self.is_stopped:
+            state = "STOPPED"
+        else:
+            state = "RUNNING"
```
shtrom, line 35: Nit: Would it make sense to have a single `state` value, rather than separate booleans? All three seem mutually exclusive.

shtrom (approved):
LGTM. Couple of nits, but non-blocking.

shtrom (approved):
Look great!

PR #204, opened 2025-01-27
create_environment_repos: allow running command on prod (bug 1944072)

zzzeid:
- also add newest ESR repo
- improve logging

### DIFF
--- a/src/lando/utils/management/commands/create_environment_repos.py
+++ b/src/lando/utils/management/commands/create_environment_repos.py
@@ -252,6 +252,15 @@ REPOS = {
             "milestone_tracking_flag_template": "cf_status_firefox_esr{milestone}",
             "commit_flags": [DONTBUILD],
         },
+        {
+            "name": "mozilla-esr128",
+            "short_name": "esr128",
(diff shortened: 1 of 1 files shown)


src/lando/utils/management/commands/create_environment_repos.py
```diff
@@ -252,6 +252,15 @@
             "milestone_tracking_flag_template": "cf_status_firefox_esr{milestone}",
             "commit_flags": [DONTBUILD],
         },
+        {
```
cgsheeh, line 255: We define these repos as dicts, then when using them we do `Repo.objects.create(**definition)`. Perhaps it would make more sense to define these as `Repo` and then create them with `repo.save()` instead.

cgsheeh (approved):

src/lando/utils/management/commands/create_environment_repos.py
zzzeid: On production this is basically a single use script. On other environments, it will be reused so it is worth improving it some more if we feel the need. I think this is pretty good as is for the time being, but we can consider these improvements for the future.

PR #207, opened 2025-01-29
middleware: allow superusers to use Lando in maintenance (bug 1944335)

zzzeid:


cgsheeh (approved):
Could you add a comment specifying the purpose of this `if` statement?

PR #212, opened 2025-01-30
workers: don't fail on ssh_private_key missing or empty in env (bug 1944728)

shtrom:


cgsheeh (approved):

zzzeid (approved):
This is good, wonder if in the case of the git workers on remote environments, also set `use_ssh` to `False` since this will not be configured anyway.

shtrom:
> This is good, wonder if in the case of the git workers on remote environments, also set use_ssh to False since this will not be configured anyway.

We could, though at least in suite, we still use the SSH key to authenticate to git.test. It's a corner case, but using SSH for git is legitimate, so I think this is probably OK like this.

PR #218, opened 2025-01-31
repo: make default_branch optional (bug 1936863)

zzzeid:


src/lando/main/models/repo.py
```diff
@@ -118,7 +118,7 @@ def scm(self) -> AbstractSCM:
         return self._scm
 
     def __str__(self):
-        return f"{self.name} ({self.default_branch})"
+        return f"{self.name} ({self.scm_type})"
```
shtrom, line 121: We may want to keep the default branch visible if present here, as this will help differentiating Git repos as long as they are in Repo+Branch mode.

shtrom (changes requested):
LGTM, but let's keep the branch in the `__str__` for Git repos.

PR #219, opened 2025-02-05
GitSCM: fix default_branch and target_cset bugs (bug 1945697)

shtrom:


cgsheeh (approved):

src/lando/main/scm/git.py
```diff
@@ -111,8 +116,10 @@ def push(
 
         command += [push_path]
 
-        if push_target:
-            command += [f"HEAD:{push_target}"]
+        if not push_target:
+            push_target = self.default_branch
+
+        command += [f"HEAD:{push_target}"]
```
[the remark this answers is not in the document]
shtrom: Yep. All the additional commits have been implicitly added to HEAD. I don't think we even know the work branch name in other places (we could read it out, but it would be no better than using HEAD) 

We should already be on the work branch from the `update_repo` call, but that's the only time we make the switch explicitly, everything after that assumes it's in the right position.

Could be better in theory, and more explicit about the branch to work on, but I think it would add a lot of overhead vs. setting HEAD once, and assuming it is correct (<- famous last words right there.)

src/lando/main/scm/git.py
```diff
@@ -278,8 +294,13 @@ def update_repo(self, pull_path: str, target_cset: Optional[str] = None) -> str:
...
         )
+
+        # Create a new work branch, named after the current time, to work in.
+        # Ideally, we'd use the revision number, too, but it's not available to the SCM.
+        # A date is good enough for now, if we need to dig into issues.
+        work_branch = f"lando-{datetime.now().strftime(ISO8601_TIMESTAMP_BASIC)}"
         self._git_run(
-            "checkout", "--force", "-B", branch, f"origin/{branch}", cwd=self.path
+            "checkout", "--force", "-B", work_branch, target_cset, cwd=self.path
         )
```
zzzeid, line 304: Should we clean up after this? Otherwise we will accumulate a bunch of branches that will never be used again.

src/lando/main/scm/git.py
```diff
@@ -267,8 +274,17 @@ def update_repo(self, pull_path: str, target_cset: Optional[str] = None) -> str:
...
+        A new work branch will be created, using the current date in its name.
         """
-        branch = target_cset or self.default_branch
+        if not target_cset:
+            target_cset = self.default_branch
+
+        remote_branch = f"origin/{target_cset}"
+        if self._git_run("branch", "--list", "--remote", remote_branch, cwd=self.path):
+            # If the branch exists remotely, make sure we get the up-to-date version.
+            target_cset = remote_branch
```
zzzeid, line 286: Curious why we are going with that `target_cset` terminology instead of a more git-centric one?

zzzeid (commented):
Just a couple of comments, can be addressed in a followup if necessary.

src/lando/main/scm/git.py
shtrom: Weight of history. It's what the standalone `HgRepo` object was using, and it got into the `abstract_scm`.

For git I'd use something like `base_commit`, but the meaning isn't changed.

src/lando/main/scm/git.py
shtrom: (copying response to correct comment) Weight of history. It's what the standalone `HgRepo` object was using, and it got into the `abstract_scm`.

For git I'd use something like `base_commit`, but the meaning isn't changed.

PR #220, opened 2025-02-05
pulse: pulse notifications (bug 1940614)

shtrom:


src/lando/pulse/pulse.py
```diff
@@ -0,0 +1,112 @@
...
+        branches = {}
+        if push.commits.count():
+            commit = push.commits.latest()
+            branches = {push_data["branch"]: commit.hash}
+
+        # XXX: to be implemented
+        tags = {}
+
+        if not branches and not tags:
+            logger.error(
```
cgsheeh, line 94: Seems like we log and continue here when we've detected something is wrong. Should we fail?

src/lando/pulse/pulse.py
```diff
@@ -0,0 +1,112 @@
...
+                "type": "push",
+                "repo_url": push_data["repo_url"],
+                "branches": branches,
+                "tags": {},
+                "time": push.datetime.strftime("%s"),
+                "push_id": push_data["push_id"],
+                "user": push_data["user"],
+                # XXX
+                "push_json_url": "FIXME",
+                "push_full_json_url": "FIXME",
```
cgsheeh, line 109: I guess we don't have a URL since we don't have the web view yet. :sweat_smile:  Can we add a bug number?

src/lando/pulse/tests/test_pulse.py
```diff
@@ -0,0 +1,123 @@
...
+    message = messages[0][0]["payload"]
+    assert message["push_id"] == push.push_id
+    assert message["type"] == "push"
+    assert message["repo_url"] == push.repo_url
+    assert repo.default_branch in message["branches"]
+    assert message["branches"][repo.default_branch] == commit.hash
+    assert not message["tags"]
+    assert message["user"] == push.user
+    # assert message['push_json_url'] == push.push_json_url
+    # assert message['push_full_json_url'] == push.push_full_json_url
```
cgsheeh, line 123: Bug number for the URLs here, so we don't forget them. :)

cgsheeh (changes requested):

src/lando/pulse/pulse.py
```diff
@@ -0,0 +1,112 @@
...
+    producer: kombu.Producer
+
+    def __init__(self, producer: kombu.Producer | None = None) -> None:
+        if not producer:
+            producer = self._make_producer()
+
+        self.producer = producer
+
+    @classmethod
+    def _make_producer(cls) -> kombu.Producer:
```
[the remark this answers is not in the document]
shtrom: I think I prefer to retain something that gives us a naked `kombu.Producer` separately to a fully-built PulseNotifier.

However, I could add `settings` as a parameter to this method for more flexibility?

src/lando/pulse/pulse.py
```diff
@@ -0,0 +1,112 @@
...
+    @classmethod
+    def pulse_message_for_push(cls, push: Push):
+        push_data = model_to_dict(push)
+
+        branches = {}
+        if push.commits.count():
+            commit = push.commits.latest()
+            branches = {push_data["branch"]: commit.hash}
+
+        # XXX: to be implemented
```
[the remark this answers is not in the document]
shtrom: No. I'll add one. We'll have to look at how the automation API generates this.

https://bugzilla.mozilla.org/show_bug.cgi?id=1957547

src/lando/pulse/pulse.py
shtrom: Should be a warning.

src/lando/pulse/pulse.py
shtrom: Added (:

src/lando/pulse/pulse.py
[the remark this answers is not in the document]
cgsheeh: Ah yeah, leaving a dedicated function for the `Producer` is a good idea. I was suggesting we could move that logic into a different constructor and leave `__init__` to take a non-`Optional`.

src/lando/pulse/pulse.py
cgsheeh: Makes sense, warning and having the message published as a no-op works for me. :)

src/lando/pulse/pulse.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @classmethod
+    def _make_producer(cls) -> kombu.Producer:
+        if settings.PULSE_HOST.startswith("memory"):
+            message = (
+                "PULSE_HOST set to a `memory` location. "
+                + "This should not be the case in non-local deployments. "
+                + str(settings.PULSE_HOST)
+            )
+            if Environment.is_remote:
```
cgsheeh, line 29: This should be `settings.ENVIRONMENT.is_remote`.

cgsheeh (dismissed):
LGTM, just need to use `settings.ENVIRONMENT.is_remote` instead of `Environment.is_remote`.

cgsheeh (approved):

src/lando/utils/management/commands/create_environment_repos.py
```diff
@@ -163,6 +198,14 @@ def handle(self, *args, **options):
                         f"Repo {definition['name']} already exists or could not be added, skipping."
                     )
                 )
+            except Exception as e:
```
zzzeid, line 201: Can you be more specific here about which exception you are catching? E.g., if you've run into a new one that was not excpected, you may be able to add it to the previous `except` block. This script is really mainly geared towards local environments anyway.

src/lando/pushlog/pushlog.py
```diff
@@ -133,9 +133,17 @@ def record_push(self) -> Optional[Push]:
             push.commits.add(commit)
 
         push.save()
+        self.push = push
```
zzzeid, line 136: How is this value used?

src/lando/pushlog/pushlog.py
```diff
@@ -133,9 +133,17 @@ def record_push(self) -> Optional[Push]:
...
+        self.is_recorded = True
+
         logger.info(f"Successfully saved {push}")
 
-        self.is_recorded = True
+        try:
+            # XXX: share this instance
+            notifier = PulseNotifier()
+            notifier.notify_push(push)
+        except Exception as exc:
```
zzzeid, line 145: Are there any specific exceptions that could be anticipated here?

zzzeid (dismissed):

src/lando/pulse/pulse.py
```diff
@@ -0,0 +1,108 @@
...
+    def _make_producer(cls) -> kombu.Producer:
+        if settings.PULSE_HOST.startswith("memory"):
+            message = (
+                "PULSE_HOST set to a `memory` location. "
+                + "This should not be the case in non-local deployments. "
+                + str(settings.PULSE_HOST)
+            )
+            if settings.ENVIRONMENT.is_remote:
+                # XXX: we should verify this much earlier on
+                raise RuntimeError(message)
```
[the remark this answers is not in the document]
shtrom: Ah, cool. I added it as a `deploy=True` check. Do we do a `check --deploy` somewhere during deployment? I haven't found any instance of it in the code.

src/lando/pushlog/pushlog.py
```diff
@@ -133,9 +133,17 @@ def record_push(self) -> Optional[Push]:
...
+        self.push = push
+        self.is_recorded = True
+
         logger.info(f"Successfully saved {push}")
 
-        self.is_recorded = True
+        try:
+            # XXX: share this instance
+            notifier = PulseNotifier()
+            notifier.notify_push(push)
```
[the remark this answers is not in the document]
shtrom: I'd rather keep this as such for the moment, to catch any unexpected exception from the PulseNotifier that would otherwise kill the landing.

We'll have to come back to this and decouple this logic in https://bugzilla.mozilla.org/show_bug.cgi?id=1957523 anyway, so that is only temporary.

src/lando/pushlog/pushlog.py
```diff
@@ -133,9 +133,17 @@ def record_push(self) -> Optional[Push]:
...
 
         push.save()
+        self.push = push
+        self.is_recorded = True
+
         logger.info(f"Successfully saved {push}")
 
-        self.is_recorded = True
+        try:
+            # XXX: share this instance
```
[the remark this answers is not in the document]
shtrom: Ah, yes. A reminder. It should happen as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1957523, too.

src/lando/pushlog/pushlog.py
shtrom: Apparently it no longer is. It was at some point in the past.

src/lando/pushlog/pushlog.py
shtrom: Not really, the point is to prevent unexpected exceptions from failing the push https://github.com/mozilla-conduit/lando/pull/220#discussion_r2030352190

src/lando/pulse/pulse.py
```diff
@@ -0,0 +1,108 @@
...
+                "type": "push",
+                "repo_url": push.repo_url,
+                "branches": branches,
+                "tags": tags,
+                "time": push.datetime.strftime("%s"),
+                "push_id": push.push_id,
+                "user": push.user,
+                # XXX: https://bugzilla.mozilla.org/show_bug.cgi?id=1957549
+                "push_json_url": "FIXME",
+                "push_full_json_url": "FIXME",
```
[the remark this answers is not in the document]
shtrom: I'm not certain how the URL will be built yet. Presumably a base-url and a well-known REST path.

The current URLs look like 
* `https://hg.mozilla.org/try/json-pushes?version=2&startID=120039&endID=120040`
* `https://hg.mozilla.org/try/json-pushes?version=2&full=1&startID=120039&endID=120040`, which is a mix of prefixes is suffixes. It may be tricky to keep as just a string, but the additional logic is also not trivial, so I'm not sure it can be a settings-only thing.

There is enough uncertainty that I'd rather not guess ahead too much until we're sure.

src/lando/pulse/pulse.py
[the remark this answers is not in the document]
zzzeid: Yes, see for example https://github.com/mozilla-it/webservices-infra/blob/main/lando/k8s/lando/templates/deployment.yaml#L116.

src/lando/pushlog/pushlog.py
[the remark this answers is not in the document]
zzzeid: Ok. Ideally if you need to catch different exceptions from `notifier = PulseNotifier()` and `notifier.notify_push(push)`, those happen in different try blocks, but if this is getting updated/refactored in the near future anyway it can wait til then.

zzzeid (dismissed):

zzzeid (approved):

PR #224, opened 2025-02-07
lando-api: merge fixes from Flask LandoAPI (Bug 1944562)

cgsheeh:


zzzeid (changes requested):
Left comments that refer to what's in here in #227.

src/lando/api/tests/test_landings.py
zzzeid, line 1: This also looks like it is out of place and should not be part of the fixup, these changes are already in `main`. I will pause the review here, and continue after the changes are fixed, since the diff still does not appear correctly.

zzzeid (changes requested):
Pausing the review about 1/3 of the way as per comment I left on `src/lando/api/tests/test_landings.py`.

src/lando/api/legacy/validation.py
```diff
@@ -23,7 +23,7 @@ def parse_landing_path(landing_path: list[dict]) -> list[tuple[int, int]]:
             (revision_id_to_int(item["revision_id"]), int(item["diff_id"]))
             for item in landing_path
         ]
-    except (ValueError, TypeError) as e:
+    except (ValueError, TypeError, KeyError) as e:
```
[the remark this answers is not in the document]
cgsheeh: In `src/lando/api/tests/test_transplants.py::test_transplant_wrong_landing_path_format` we started hitting a `KeyError`. Without digging too deep, I think the issue is that in `connexion` the missing `diff_id` field in the last section of the test causes an error before the `/transplants` code is reached, whereas in Django we end up in this code path where we finally noticed the `diff_id` key field is missing. It may have something to do with the checks system as well, but since it's a straightforward fix I added the new exception type here and moved on.

src/lando/api/tests/conftest.py
```diff
@@ -361,21 +361,6 @@ def _get_landing_worker(scm_type):
...
-        mem["set"] = True
-        mem["val"] = val
-
-    monkeypatch.setattr(
-        "lando.api.legacy.transplants.tokens_are_equal",
-        lambda t1, t2: mem["val"] if mem["set"] else tokens_are_equal(t1, t2),
-    )
-    return set_value
-
-
```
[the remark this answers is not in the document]
cgsheeh: See [this commit](https://github.com/mozilla-conduit/lando-api/commit/8ecc15c235c6b4694da816b3c43febd0c925f82f).

src/lando/api/tests/test_transplants.py
```diff
@@ -1824,3 +1863,64 @@ def test_warning_multiple_authors(phabdouble, mocked_repo_config, create_state):
     assert (
         warning.details == "Revision has multiple authors: alice, bob."
     ), "Multiple authors on a revision should return a warning."
+
+
+@pytest.mark.django_db(transaction=True)
+def test_transplant_on_linked_legacy_repo(
```
zzzeid, line 1869: I think this was probably accidentally removed in the merge, and re-added here. It should be removed from this PR and rolled up in the merge commit (this file was one that had conflicts).

src/lando/api/tests/mocks.py
```diff
@@ -1819,3 +1815,86 @@ def _build_handlers(self):
             if hasattr(getattr(self, a), "_conduit_method")
         ]
         return {handler._conduit_method: handler for handler in handlers}
+
+
+class TreeStatusDouble:
```
zzzeid, line 1820: This change looks like it was accidentally removed in the merge PR and re-added here. It is already in `main`.

zzzeid (changes requested):
Looks good, two outstanding issues (and one previous unresolved comment) and we should be good to go.

src/lando/api/legacy/transplants.py
```diff
@@ -770,6 +767,16 @@ def blocker_unsupported_repo(
...
 
+    if not stack_state.landing_assessment:
+        return
+
+    # If the landing repo on the current revision is set as a legacy repo
+    # of another repo, then change the landing repo to match the target.
+    set_repo = stack_state.landable_repos[repo]
+    stack_state.landing_assessment.landing_repo = (
+        set_repo if not set_repo.is_legacy else set_repo.new_target
+    )
```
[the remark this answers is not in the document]
cgsheeh: The "landing repo" is only relevant for a landing assessment. It is possible to have revisions with different repos in a stack, so the check doesn't make full sense at the "stack assessment" level. It is also possible to have a valid landing path within a stack that has multiple repos, when the path has the same repo for each revision. So we can't do the more simple check of saying any stack with multiple repos is invalid for landing.

I agree that putting this here is a little wonky. Perhaps as a follow-up we could find the landing repo when we build the landing assessment state, instead of here in this check, and then the check could instead just inspect the state and see if a valid landing repo was found (ie `if not landing_assessment.landing_repo`).

src/lando/api/tests/test_transplants.py
cgsheeh: This wasn't removed in the merge PR, the merge PR pre-dates this change. This change was added during the work week, so I had to merge changes landed after I started the initial merge back into this branch. I can't remove it from the merge commit because it doesn't exist there yet.

zzzeid: >This wasn't removed in the merge PR, the merge PR pre-dates this change

This is no longer true as of yesterday, after you force pushed as seen [above](https://github.com/mozilla-conduit/lando/pull/224#event-16462467560) and reverted some files to main.

Everything added in the work week should already be part of the "merge PR" (i.e., [api-merge-without-fixes](https://github.com/mozilla-conduit/lando/tree/api-merge-without-fixes), #225). This PR (#224) should only have the fixes. I think this is the same case (and fix) as what you did with the other files you already fixed in the merge commit in #225 yesterday (e.g., `src/lando/api/legacy/workers/base.py`).

zzzeid (changes requested):
There are merge conflicts in this branch, and it looks like the changes in `api-merge-without-fixes` have not been pulled back into this yet (as was done in the previous round of review). This is evident from the number of changes visible in this diff (should be similar to the previous round of review) since the histories have diverged.

image

```
git checkout api-merge
git merge origin/api-merge-without-fixes
```

Should fix this.

cgsheeh:
> There are merge conflicts in this branch, and it looks like the changes in `api-merge-without-fixes` have not been pulled back into this yet (as was done in the previous round of review). This is evident from the number of changes visible in this diff (should be similar to the previous round of review) since the histories have diverged.
> image
> 
> ```
> git checkout api-merge
> git merge origin/api-merge-without-fixes
> ```
> 
> Should fix this.

Done.

src/lando/api/legacy/transplants.py
```diff
@@ -859,6 +866,7 @@ def blocker_single_landing_repo(
 
     # Set the landing repo field on the `LandingAssessmentState`.
     landing_repo = stack_state.landable_repos.get(repo_phid)
+
     stack_state.landing_assessment.landing_repo = landing_repo
```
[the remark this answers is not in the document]
zzzeid: Ping on this comment? Not sure if it was missed or not needed here. Mainly asking from a consistency perspective.

src/lando/api/tests/test_transplants.py
zzzeid: This still shows up here? This indicates there is something wrong in the merge PR. This test is already in `main`.

zzzeid (changes requested):
There is still a change that does not belong in this PR, but it looks like it is the last such one (`src/lando/api/tests/test_transplants.py`). This indicates that this section of the code was removed when resolving the merge conflict in #225, and needs to be re-added back to #225.

zzzeid (changes requested):
This PR can not be reviewed while it still has unresolved conflicts with the target branch. Once these conflicts are fixed I will review it (this is the same workflow as the last round of review.)

src/lando/api/legacy/transplants.py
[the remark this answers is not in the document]
cgsheeh: I moved this logic into `blocker_single_landing_repo`, where we usually set the landing repo for a revision. I also added a check that the landing repo is in `landable_repos`, so now we only set the landing repo in a single location and have verified the repo is valid. :+1:

src/lando/api/legacy/transplants.py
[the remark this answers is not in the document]
zzzeid: >The "landing repo" is only relevant for a landing assessment

Makes sense.

I think we should file a bug to track this. I think the main issue is that the blocker/check is altering the state of the target repo -- maybe this can happen somewhere higher up in the chain (this is more so an issue with the original implementation though, not so much the "porting" so it's out of scope here to some extent anyway). Ideally the check would get the target repo already in the correct form.

src/lando/api/legacy/transplants.py
```diff
@@ -866,10 +854,15 @@ def blocker_single_landing_repo(
...
+    if repo_phid not in stack_state.landable_repos:
+        return "Landing repository is not supported by Lando."
 
-    stack_state.landing_assessment.landing_repo = landing_repo
+    # If the landing repo on the current revision is set as a legacy repo
+    # of another repo, then change the landing repo to match the target.
+    set_repo = stack_state.landable_repos[repo_phid]
+    stack_state.landing_assessment.landing_repo = (
+        set_repo if not set_repo.is_legacy else set_repo.new_target
+    )
```
zzzeid, line 865: Not sure if this will have impact on functionality at this time, but it's a little odd that `stack_state.landing_assessment.landing_repo` will have different values in different checks. Maybe these should be set elsewhere to apply to all checks? Maybe there should be something like `stack_state.landing_repo_mapping` which could already have the correct mapping from a revision PHID to a target repo.

zzzeid (approved):
Looks good to me. I think the issue with `stack_state.landing_assessment.landing_repo` should be tracked in a separate bug, as it could result in an inconsistency between checks when it comes to the target repo.

I assume that the legacy repo target functionality has been tested and works as expected based on these changes (i.e., a revision with a repo with a `new_target` set lands in that repo as expected).

src/lando/api/legacy/transplants.py
cgsheeh: I filed [bug 1951169](https://bugzilla.mozilla.org/show_bug.cgi?id=1951169) to track this.

PR #225, opened 2025-02-19
Merge remainder of lando-api repo

cgsheeh:


zzzeid (changes requested):
This should be redone to align with #226.

zzzeid (changes requested):
This PR should consist only of the merge, without any additional commits, as it will be merged as-is into the repo. Currently it contains this additional commit: `revert some files to origin/main`.

As per previous discussion, this should be "hard reset" to match #226 or you should use #226 as the target branch of your other PR (#224) and close this one.

zzzeid (changes requested):
This looks good, just needs the two updates from the comments I left in #224.

nit: For consistency, I would suggest renaming the commit message / PR title to:
`Merge lando-api repo` which would line up with the original merge PR (#13).

zzzeid (changes requested):
See my comments earlier today in #224 which would affect this PR.

zzzeid (approved):
Looks good to me. Only nit is to rename the PR title to something like **`Merge remainder of lando-api repo`** for consistency/clarity. Also I just enabled "merge commits" on this repo, so make sure to pick that when merging this PR (but not #224).

cgsheeh:
> Looks good to me. Only nit is to rename the PR title to something like `Merge remainder of lando-api repo` for consistency/clarity. Also I just enabled "merge commits" on this repo, so make sure to pick that when merging this PR (but not #224).

Done! :+1:

src/lando/api/legacy/transplants.py
```diff
@@ -356,26 +568,71 @@ def warning_code_freeze(*, repo, **kwargs):
...
+    stack_state: StackAssessmentState,
+) -> Optional[str]:
     """Check the user has a proper auth0 email."""
+    if not stack_state.landing_assessment:
+        return None
+
     return (
         None
-        if lando_user.email
+        if stack_state.landing_assessment.auth0_user.email
```
zzzeid, line 614: @cgsheeh FYI after this merge, `lando_user` got renamed back to `auth0_user` and was not updated in the fixup PR. Not a blocker, but it should be renamed back to `lando_user` everywhere for consistency as this convention is used elsewhere, and references to `auth0_user` were entirely removed from the repo.

PR #232, opened 2025-03-06
conftest: mark `overrides` as possibly being `None`

cgsheeh:
Since the default value is `None`, `overrides` is not always `dict`.


shtrom (approved):
lgtm

zzzeid (approved):

PR #237, opened 2025-03-12
transplants: rename `auth0_user` to `lando_user` (Bug 1953213)

cgsheeh:


zzzeid (changes requested):
Other files still user auth0_user, those should also be changed.

zzzeid (approved):
Never mind, since those are treestatus related and will be revisited later.

PR #238, opened 2025-03-14
GitSCM: rename app_privkey to private_key (bug 1950893)

shtrom:


cgsheeh (approved):
I guess we're at a minor risk of this variable name being changed and this bug being regressed. Maybe we could add a comment that the variable name is important?

Will we know this works only after a deploy or did we test it somehow?

zzzeid (approved):
Since this does not block the upcoming transition (only an issue for the git landing worker) we can hold off on merging/deploying this until after the transition.

shtrom:
> Will we know this works only after a deploy or did we test it somehow?

We'll need a deploy. Do we have sentry enable on lower envs?

cgsheeh:
> > Will we know this works only after a deploy or did we test it somehow?
> 
> We'll need a deploy. Do we have sentry enable on lower envs?

Yes, we have Sentry enabled on dev/stage. :+1:

shtrom:
Tested in dev. Will land on Monday.

PR #239, opened 2025-03-14
jinja: fix linkified transplant URLs for Git (bug 1934483)

shtrom:


cgsheeh (approved):

src/lando/jinja.py
```diff
@@ -211,9 +211,15 @@ def linkify_transplant_details(text: str, transplant: dict) -> str:
...
-    replace = r'<a href="{repo_url}/rev/\g<1>">{repo_url}/rev/\g<1></a>'.format(
-        repo_url=transplant["repository_url"]
-    )
+
+    parsed_repo_url = urllib.parse.urlsplit(transplant["repository_url"])
+    # We assume HG by default (legacy path), but use a Github-like path if 'git' is
+    # present in the netloc.
+    link_template = r'<a href="{repo_url}/rev/\g<1>">{repo_url}/rev/\g<1></a>'
+    if "git" in parsed_repo_url.netloc:
+        link_template = r'<a href="{repo_url}/commit/\g<1>">{repo_url}/commit/\g<1></a>'
```
zzzeid, line 220: I think as an improvement to this in the future, `link_template` could be a parameter that is set or determined on the `Repo` object, and instead of having a `transplant` dict here, we could just be passing the landing job object itself. Part of de-legacy-fying Lando.

zzzeid (approved):

src/lando/jinja.py
zzzeid: Filed [bug 1955337](https://bugzilla.mozilla.org/show_bug.cgi?id=1955337).

PR #244, opened 2025-03-19
landing_worker: use `mach format` for autoformatting (Bug 1807712)

cgsheeh:
Now that `mach format` runs `eslint` as a formatter without fixes,
we can use `mach format` to autoformat patches, removing the need
to maintain our own list of valid formatters.


shtrom (approved):
lgtm.

This was a short-lived fix 😆

zzzeid (approved):

PR #246, opened 2025-03-20
legacy: remove unneeded serialization of transplants (bug 1955337)

zzzeid:


cgsheeh (approved):

src/lando/main/models/landing_job.py
```diff
@@ -347,28 +347,14 @@ def transition_status(
...
-                self.updated_at.astimezone(datetime.timezone.utc).isoformat()
-            ),
-        }
+    @property
+    def legacy_details(self) -> str:
+        """Return a string of the landed commit id or the error details."""
+        return (
+            self.error or self.landed_commit_id
+            if self.status in (LandingJobStatus.FAILED, LandingJobStatus.CANCELLED)
+            else self.landed_commit_id or self.error
```
shtrom, line 356: Would there be a way to simplify this expression? The duplication of terms is breaking my brain 😅

src/lando/jinja.py
```diff
@@ -72,15 +72,15 @@ def calculate_duration(start: str, end: Optional[str] = None) -> dict[str, int]:
...
+def tostatusbadgeclass(landing_job: LandingJob) -> str:
     mapping = {
         "aborted": "Badge Badge--negative",
         "submitted": "Badge Badge--warning",
         "in_progress": "Badge Badge--warning",
         "landed": "Badge Badge--positive",
         "failed": "Badge Badge--negative",
     }
-    return mapping.get(status["status"].lower(), "Badge Badge--negative")
+    return mapping.get(landing_job.status.lower(), "Badge Badge--negative")
```
shtrom, line 83: Should we add some resiliency for https://bugzilla.mozilla.org/show_bug.cgi?id=1956970, e.g, if `landing_job.status == None`?

src/lando/main/models/landing_job.py
zzzeid: I will take a stab at simplifying it.

zzzeid: I went back to check, this code is very old. I don't think we need to still handle the case of showing a "landed commit ID" when a job is failed or deferred, since that should not ever happen (`job.landed_commit_id` is only set when a job is marked as `LANDED`).

shtrom (approved):

PR #247, opened 2025-03-20
landing_worker: use active_repos when filtering next job (bug 1955367)

zzzeid:


cgsheeh (approved):
Can we add a test for this?

zzzeid:
> Can we add a test for this?

Added a test to validate that the repos returned are correct, however we don't have an integration test that runs the worker with jobs without explicitly specifying a job (e.g., with a `worker.start(max_loops=...)` kind of implementation). I think this would be a good addition but I will file a separate bug for this so we can unblock the current bug.

PR #252, opened 2025-03-24
code style: add noqa (manual and auto) (bug 1920792)

zzzeid:


dklawren (approved):

cgsheeh (changes requested):
I think we should use ruff's [per-file ignore rules](https://docs.astral.sh/ruff/settings/#lint_per-file-ignores) for much of these annotations instead. For example many of the methods in test directories should be exempt from type annotation requirements and it would be easier to ignore those rules entirely in `tests/` directories than adding so many `noqa` comments, and having to continue adding `noqa` to new code in the future.

In the non-tests directories, we should look at adding type hints instead of adding a comment granting a typing exception.

cgsheeh (approved):
LGTM. We could also consider skipping `migrations/` directories for these checks since those are mostly generated files.

zzzeid:
> LGTM. We could also consider skipping `migrations/` directories for these checks since those are mostly generated files.

In this case it is actually manually-added code that @shtrom added that was flagged, none of the autogenerated stuff.

PR #267, opened 2025-04-02
landing_worker: raise `TemporaryFailureException` instead of `return False`

cgsheeh:


shtrom (approved):

zzzeid:
Is there a bug or issue for this, or was this a preemptive fix?

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -250,7 +250,7 @@ def apply_and_push(
             job.transition_status(LandingJobAction.DEFER, message=message)
 
             # Try again, this is a temporary failure.
-            return False
+            raise TemporaryFailureException(message) from e
```
zzzeid, line 253: This is the only instance where we use `from` when raising a `TemporaryFailureException` (two other instances in this file).

PR #276, opened 2025-04-03
jinja: update linkify_transplant_details with landing_job (bug 1955337)

zzzeid:


src/lando/jinja.py
```diff
@@ -203,23 +203,23 @@ def linkify_revision_ids(text: str) -> str:
     return re.sub(search, replace, str(text), flags=re.IGNORECASE)
 
 
-def linkify_transplant_details(text: str, transplant: dict) -> str:
+def linkify_transplant_details(text: str, landing_job: LandingJob) -> str:
     # The transplant result is not always guaranteed to be a commit id. It
     # can be a message saying that the landing was queued and will land later.
-    if transplant["status"].lower() != "landed":
+    if landing_job.status.lower() != "landed":
```
cgsheeh, line 209: We should compare with `LandingJobStatus` here instead.

src/lando/jinja.py
```diff
@@ -203,23 +203,23 @@ def linkify_revision_ids(text: str) -> str:
     return re.sub(search, replace, str(text), flags=re.IGNORECASE)
 
 
-def linkify_transplant_details(text: str, transplant: dict) -> str:
+def linkify_transplant_details(text: str, landing_job: LandingJob) -> str:
```
cgsheeh, line 206: Could we make this a method on the `LandingJob` instead of using a template filter? ie calling this would be `LandingJob.linkify_transplant_details(text)`.

cgsheeh (changes requested):

src/lando/jinja.py
zzzeid: I don't think this functionality belongs on the model, specifically adding the html tags, etc... it's a specific UI feature that probably still belongs in a jinja filter, though we should probably rework these legacy filters at some point. Though I think the actual URL value should be stored on the model (see my comment on Slack earlier).

There are also two other similar methods in this module that take in a LandingJob and do something similar with the status.

cgsheeh (approved):

PR #278, opened 2025-04-03
maintenance: improve command and set sensible defaults (bug 1955422)

zzzeid:


src/lando/utils/management/commands/maintenance.py
```diff
@@ -20,56 +20,59 @@ def add_arguments(self, parser):  # noqa: ANN001
...
 
-        if not check:
-            action_mapping[action]()
-            self.stdout.write(f"Turned maintenance mode {action}.")
-            sys.exit()
-
+        # If there are unapplied migrations, then both the web interface and the
+        # workers need to be put in maintenance.
         unapplied_migrations = (
             subprocess.run(["lando", "migrate", "--check", "--noinput"]).returncode != 0
```
shtrom, line 66: Unrelated to this PR, but should we use something like `call_command` rather than forking a subprocess? https://docs.djangoproject.com/en/5.0/ref/django-admin/#running-management-commands-from-your-code

shtrom (approved):

src/lando/utils/management/commands/maintenance.py
zzzeid: If I remember correctly, I think I ran into trouble capturing `returncode` from `call_command`, which may be why I reverted to this. I filed a bug to tackle this in the future, I'm sure there's a way to do it using `call_command`.

PR #280, opened 2025-04-03
phabricator: get_conduit_result_data implementation (bug 1953929)

zzzeid:
- add method to aggregate result data from multiple requests

### DIFF
--- a/src/lando/api/tests/test_phabricator_client.py
+++ b/src/lando/api/tests/test_phabricator_client.py
@@ -2,6 +2,8 @@
 Tests for the PhabricatorClient
 """
 
+import unittest.mock as mock
+
 import pytest
 import requests
 import requests_mock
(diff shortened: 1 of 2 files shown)


src/lando/utils/phabricator.py
```diff
@@ -212,6 +212,21 @@ def call_conduit(self, method: str, **kwargs) -> Any:  # noqa: ANN401
         PhabricatorAPIException.raise_if_error(response)
         return response.get("result")
 
+    def get_conduit_result_data(self, method: str, **kwargs) -> dict:
```
cgsheeh, line 215: Type hint could be more specific here: `dict[str, list[Any]]:`

Bikeshed: `call_conduit_paginated()` would make it clearer what this method is doing. A more detailed docstring would also be nice.

src/lando/api/tests/test_phabricator_client.py
```diff
@@ -83,3 +85,24 @@ def test_phabricator_exception(get_phab_client):
...
+    first_batch = []
+    second_batch = []
+    for i in range(100):
+        first_batch.append(f"batch-1-{i}")
+        second_batch.append(f"batch-2-{i}")
+
+    initial_result = {"data": first_batch.copy(), "cursor": {"after": 1234}}
+    second_result = {"data": second_batch.copy(), "cursor": {}}
+
+    mock_call_conduit.side_effect = [initial_result, second_result]
```
cgsheeh, line 103: Optimally we would use the `PhabricatorDouble` to create 2 pages of `differential.diff.search` (or some other endpoint) results and ensure the new retrieval method correctly returns all the results. This seems reasonable for now, but we should do that as a follow-up.

cgsheeh (approved):

src/lando/api/tests/test_phabricator_client.py
zzzeid: I think increasing the integration reach of this test could be beneficial, but don't think it's necessary since we are not interacting directly with the Phabricator API or care about the internals of it, instead relying on another instance method that has an expected response.

I think if we wanted to test this via PhabricatorDouble, we should add a test elsewhere (maybe in `test_revisions`) that does a more proper end-to-end test by creating revisions, diffs, etc. and then calling the API. As a side note, it looks like the only endpoint that properly supports this in the mocked `PhabricatorDouble` class is `transaction_search` so this wouldn't work if we tried it today.

src/lando/utils/phabricator.py
zzzeid: nit: I think "collated" is more appropriate here, since we're "de-paginating" results rather than paginating them.

cgsheeh (approved):

PR #284, opened 2025-04-08
pulse: pulse management commands (bug 1958734)

shtrom:


zzzeid (dismissed):

shtrom:
(I had to rebase, and merging main into it was conflicting a lot more than I was comfortable with given how minute the changes here are).

src/lando/pulse/tests/test_pulse_notify.py
```diff
@@ -0,0 +1,183 @@
...
+        "repo": repo.name,
+        "push_id": push_id,
+    }
+    if force_flag:
+        opts[force_flag] = True
+    call_command("pulse_notify", stdout=out, **opts)
+
+    user_message = out.getvalue()
+    assert "Notifying" in user_message
+    assert str(push) in user_message
```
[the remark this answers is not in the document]
shtrom: I tend to think it's more robust like this (: but maybe not robust to the same things.

My reasoning is that rather than binding to a full string, I test that the important elements (“Notifying” and the descriptor of the push) are present. This way we can alter the string without having to change the test. So the test is robust to (most) non-functional changes to the string (e.g., punctuation, additional details, ...).

The flip side is that if another string has `Notifying` and the `str(push)` in it, the test may pass erroneously. So in that respect, it is also less robust.

TBF, I think either occurrences (string change and other string present) are pretty much as unlikely as each other.

A middle-ground solution may be a single match on a static substring containing the key elements of the message.

src/lando/pulse/management/commands/pulse_notify.py
```diff
@@ -0,0 +1,87 @@
...
+        parser.add_argument(
+            "-O",
+            "--force-out-of-order",
+            default=False,
+            action=argparse.BooleanOptionalAction,
+        )
+
+    def handle(self, *args, **options) -> None:
+        repo_name = options["repo"]
+        push_id = options.get("push_id")
```
zzzeid, line 33: nit: This also technically defaults to `0` no?

src/lando/pulse/management/commands/pulse_declare.py
```diff
@@ -0,0 +1,20 @@
+from django.core.management.base import BaseCommand, CommandError
+
+from lando.pulse.pulse import PulseNotifier
+
+
+class Command(BaseCommand):
+    help = "Declare the Pulse exchange"
+    name = "pulse_declare"
+
+    def handle(self, *args, **options) -> None:
```
zzzeid, line 10: nit: Would be nice to include a bit here about why this command is needed.

zzzeid (approved):

src/lando/pulse/management/commands/pulse_notify.py
shtrom: Ah! Yes!

PR #285, opened 2025-04-09
automation-actions: implement `create-commit` action (Bug 1959441)

cgsheeh:
Implement the `create-commit` action, which is similar to the
existing `add-commit` action, except the commit data is passed
as a diff and metadata fields directly.

### DIFF
--- a/src/lando/headless_api/api.py
+++ b/src/lando/headless_api/api.py
@@ -178,6 +178,48 @@ class AddCommitAction(Schema):
         return True
 
 
+class CreateCommitAction(Schema):
+    """Create a new commit from a diff and metadata."""
+
+    action: Literal["create-commit"]
+    author: str
+    commit_message: str
+    date: datetime.datetime
+    diff: str
+
+    def process(
+        self, job: AutomationJob, repo: Repo, scm: AbstractSCM, index: int
+    ) -> bool:
+        """Create a new commit from a diff and metadata."""
+        try:
+            scm.apply_patch(
+                self.diff,
+                self.commit_message,
+                self.author,
+                self.date.isoformat(),
+            )
+        except PatchConflict as exc:
+            message = (
+                f"Merge conflict while creating commit in `create-commit`, "
+                f"action #{index}.\n\n"
+                f"{str(exc)}"
+            )
+            raise AutomationActionException(
+                message=message, job_action=JobAction.FAIL, is_fatal=False
+            )
+        except Exception as e:
+            message = (
+                f"Aborting, could not create commit from `create-commit`, "
+                f"action #{index}."
+                f"\n{e}"
+            )
+            raise AutomationActionException(
+                message=message, job_action=JobAction.FAIL, is_fatal=False
+            )
+
+        return True
+
+
 class MergeOntoAction(Schema):
     """Merge the current branch into the target commit."""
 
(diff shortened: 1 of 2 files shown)


src/lando/headless_api/api.py
```diff
@@ -178,6 +178,48 @@ def process(
...
+                self.date.isoformat(),
+            )
+        except PatchConflict as exc:
+            message = (
+                f"Merge conflict while creating commit in `create-commit`, "
+                f"action #{index}.\n\n"
+                f"{str(exc)}"
+            )
+            raise AutomationActionException(
+                message=message, job_action=JobAction.FAIL, is_fatal=False
```
zzzeid, line 208: It would be great to add more documentation about `is_fatal` in `AutomationActionException`, it's unclear what makes an error fatal or not. Is this equivalent to temporary/permanent?

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -589,6 +589,109 @@ def test_automation_job_add_commit_fail(
...
+    )
+
+    hg_automation_worker.worker_instance.applicable_repos.add(repo)
+
+    # Mock `phab_trigger_repo_update` so we can make sure that it was called.
+    mock_trigger_update = mock.MagicMock()
+    monkeypatch.setattr(
+        "lando.api.legacy.workers.automation_worker.AutomationWorker.phab_trigger_repo_update",
+        mock_trigger_update,
+    )
```
zzzeid, line 635: out of scope, but noticing this repeated in many places, might make sense to just add it to the fixture, and maybe pass a flag to disable it if needed.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -589,6 +589,109 @@ def test_automation_job_add_commit_fail(
...
+--- a/test.txt
++++ b/test.txt
+@@ -1,1 +1,2 @@
+ TEST
++adding another line
+""".lstrip()
+
+
+@pytest.mark.django_db
+def test_automation_job_create_commit_success_hg(
```
zzzeid, line 603: This test and the one below can be combined into a parameterized test, as they mostly use the same code. The worker and repos can be abstracted similar to how other tests are abstracted without having to do anything SCM specific in the tests. For an example, see tests in test_landings.py. Basically adding a `get_automation_worker` fixture, since you already are fetching `repo_mc` based on the scm type.

Only the happy path is tested, a test should be added to test the exception handling that is implemented as well.

zzzeid (changes requested):

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -589,6 +589,109 @@ def test_automation_job_add_commit_fail(
...
 
 
+PATCH_DIFF = """
+diff --git a/test.txt b/test.txt
+--- a/test.txt
++++ b/test.txt
+@@ -1,1 +1,2 @@
+ TEST
++adding another line
+""".lstrip()
```
shtrom, line 599: We probably already have a fixture for this.

shtrom: Ah, no, not the patch alone. Nevermind.

src/lando/headless_api/api.py
cgsheeh: Yes it is, we should switch to using `TemporaryFailureException` and `PermanentFailureException` at some point in the future. I wrote this code before those existed. :sweat_smile:

shtrom (approved):

src/lando/headless_api/tests/test_automation_api.py
cgsheeh: File https://bugzilla.mozilla.org/show_bug.cgi?id=1960790 for this. :)

PR #288, opened 2025-04-10
admin: add `formatted_replacements` to `LandingJob` admin page (Bug 1959743)

cgsheeh:


zzzeid (approved):
Feel free to add to https://github.com/mozilla-conduit/lando/pull/255 for admin changes that don't need to go our right away too.

PR #292, opened 2025-04-17
automation-actions: implement `tag` action (Bug 1959583)

cgsheeh:
Implement the `tag` automation action. The action creates
a tag at either the currently checked out commit during
action processing, or at a specific commit if passed.

Update the `run_automation_job` logic to track new tags
and push them separately on Git, since Git does not create
new commits for tags.

We also updated `set_landing_system.py` to work with
the `tag` command. Since the `extra` option is passed
explicitly as `None`, using `.setdefault` does not set
the value to a dict, so we update the logic to ensure it
is set appropriately.


src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -885,6 +885,256 @@ def test_automation_job_merge_onto_fail(
...
+    automation_worker.worker_instance.applicable_repos.add(repo)
+
+    assert not automation_worker.run_automation_job(job)
+    job.refresh_from_db()
+    assert job.status == JobStatus.FAILED
+    assert "Aborting, could not perform `tag`, action #1" in job.error
+
+
+@pytest.mark.django_db
+def test_automation_job_tag_success_hg(
```
shtrom, line 1066: No failure case for hg?

shtrom (changes requested):
We should merge  `push_tags` into `push`, and let the TagAction return a list of additional refs to push.

This would allow us any run of an automation schedule results in a single push, that we can confirm.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -885,6 +885,256 @@ def test_automation_job_merge_onto_fail(
...
+    assert job.status == JobStatus.LANDED
+
+    # Tag should be on the most recent commit.
+    expected_commit = subprocess.run(
+        ["git", "rev-parse", "HEAD"],
+        cwd=repo.system_path,
+        capture_output=True,
+        check=True,
+        text=True,
+    ).stdout.strip()
```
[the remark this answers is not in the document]
cgsheeh: D'oh. :)

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -885,6 +885,256 @@ def test_automation_job_merge_onto_fail(
...
+    request,
+):
+    repo = repo_mc(SCM_TYPE_HG)
+
+    repo_path = Path(repo.system_path)
+
+    # Create a new commit that will be tagged.
+    _create_hg_commit(request, repo_path)
+    expected_commit = subprocess.run(
+        ["hg", "log", "-r", ".", "-T", "{node}"],
```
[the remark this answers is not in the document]
cgsheeh: We can't use `head_ref` without being inside a `for_push` context manager, so I left these as regular `hg` calls for now.

src/lando/main/scm/git.py
```diff
@@ -540,3 +540,19 @@ def merge_onto(
...
+        tag_command = ["tag", name]
+
+        if target:
+            tag_command.append(target)
+
+        self._git_run(*tag_command, cwd=self.path)
+
+    def push_tag(self, tag: str, remote: str):
+        """Push the tag with name `tag` to `remote`."""
+        self._git_run("push", remote, f"refs/tags/{tag}", cwd=self.path)
```
zzzeid, line 558: This would be better located inside the existing `push` method (similar to what @shtrom said elsewhere). Otherwise all the auth and extraneous push parameters will be lost, and I'm not sure this would work anyway as it exists now given that it does not implement auth/token parameters in the push.

src/lando/main/tests/test_git.py
```diff
@@ -503,3 +503,91 @@ def test_GitSCM_merge_onto(
...
+    git_setup_user: Callable,
+    request: pytest.FixtureRequest,
+    tmp_path: Path,
+):
+    clone_path = tmp_path / request.node.name
+    clone_path.mkdir()
+
+    scm = GitSCM(str(clone_path))
+    scm.clone(str(git_repo))
+    git_setup_user(str(clone_path))
```
zzzeid, line 575: This comment applies to other similar tests in this module.

These lines should be in a fixture (repeated above).
Side note, why not use `repo_mc` to keep things consistent/abstracted with the push/pull paths and other tests? Slightly more high level but still worth it if it simplifies these tests I think.

src/lando/headless_api/tests/test_automation_api.py
zzzeid, line 1066: This test is very similar to `test_automation_job_tag_success_git_new_commit`, and could benefit from being parameterized or use of fixtures to avoid repetition.

zzzeid (changes requested):

src/lando/headless_api/tests/test_automation_api.py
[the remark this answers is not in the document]
shtrom: fine! 😆

src/lando/main/scm/abstract_scm.py
```diff
@@ -229,3 +229,14 @@ def merge_onto(
 
         Return the SHA of the newly created merge commit.
         """
+
+    @abstractmethod
+    def tag(self, name: str, target: str | None):
+        """Create a new tag called `name` on the `target` commit.
+
+        If `target` is `None`, use the currently checked out commit.
+        """
```
shtrom, line 238: For the sake of maintaining the Pushlog, could this return a `CommitData` object for the commit pointed to by the Tag?

I think implementations could just return `self.describe_commit(name)`

Then we can do something like this in the action:

```
tagged_commit_data = scm.tag(name, target)
pushlog.add_tag(name, tagged_commit_data)
```

src/lando/main/tests/test_git.py
cgsheeh: I used the existing conventions in the other tests in this file, which don't use the `repo_mc` fixture. I agree that we should try and keep things more consistent, but let's do it as a follow-up since fixing all the tests in this file is out of scope.

src/lando/main/scm/git.py
```diff
@@ -115,14 +116,20 @@ def push(
...
+        push_command = [*common_args, f"HEAD:{push_target}"]
 
-        self._git_run(*command, cwd=self.path)
+        self._git_run(*push_command, cwd=self.path)
+
+        # If any tags were passed, ensure they are pushed.
+        if tags:
+            for tag in tags:
+                push_tags_command = [*common_args, f"refs/tags/{tag}"]
+                self._git_run(*push_tags_command, cwd=self.path)
```
shtrom, line 132: I think we can push the tags at the same time 
image

shtrom (changes requested):
Let's push the tags at the same time as the branches.

shtrom (approved):
lgtm!

PR #294, opened 2025-04-18
pushlog: add add_tag (bug 1957547)

shtrom:


src/lando/conftest.py
```diff
@@ -572,11 +572,13 @@ def tag_factory(repo: Repo, seqno: int, commit: Commit) -> Tag:
 
 @pytest.fixture
 def make_push():
-    def push_factory(repo: Repo, commits: list[Commit]):
+    def push_factory(repo: Repo, commits: list[Commit] = [], tags: list[Tag] = []):
```
cgsheeh, line 575: You shouldn't use `[]` as a default in a keyword argument. Use `None` as the default and create the list inside the function if you need an empty list as the default instead.

src/lando/pushlog/models/commit.py
```diff
@@ -239,13 +239,52 @@ class Tag(models.Model):
         on_delete=models.SET_NULL,
     )
 
+    _scm_commit: CommitData | None = None
+
+    @staticmethod
+    def for_scm_commit(repo: Repo, name: str, scm_commit: CommitData):  # noqa: ANN205
```
cgsheeh, line 245:
```suggestion
    def for_scm_commit(repo: Repo, name: str, scm_commit: CommitData) -> typing.Self:
```

src/lando/pushlog/models/commit.py
```diff
@@ -239,13 +239,52 @@ class Tag(models.Model):
...
+        """Create a partial Tag object, based on CommitData not present in the DB
```
cgsheeh, line 246: Nit: first line of a docstring should be a single line, followed by a newline.

src/lando/pushlog/pushlog.py
```diff
@@ -58,22 +58,30 @@ class PushLog:
...
+            # We cannot use the default value of a mutable type argument,
+            # as it will get reused on every initialisation.
             commits = []
         self.commits = commits
 
+        if not tags:
+            # We cannot use the default value of a mutable type argument,
+            # as it will get reused on every initialisation.
+            tags = []
+        self.tags = tags
```
cgsheeh, line 83: Nit: this could also be written as `self.tags = tags or []`

src/lando/pushlog/models/commit.py
```diff
@@ -239,13 +239,52 @@ class Tag(models.Model):
         on_delete=models.SET_NULL,
     )
 
+    _scm_commit: CommitData | None = None
+
+    @staticmethod
```
cgsheeh, line 244: Looks like this should be `@classmethod` and the use `cls` instead of `Tag` below?

cgsheeh (changes requested):
LGTM, except we should use `None` as the default argument in `push_factory` to avoid confusing side effects in tests.

src/lando/pushlog/models/commit.py
```diff
@@ -239,13 +239,52 @@ class Tag(models.Model):
...
         return f"{self.__class__.__name__}(repo={self.repo!r}, name={self.name}, commit={self.commit})"
 
     def __str__(self) -> str:
-        return (
-            f"Tag {self.name} in {self.repo.url} pointing to Commit {self.commit.hash}"
-        )
+        try:
+            return f"Tag {self.name} in {self.repo.url} pointing to Commit {self.commit.hash}"
+        except Tag.commit.RelatedObjectDoesNotExist:
+            return f"Partial tag {self.name} in {self.repo.url} pointing to Commit {self._scm_commit.hash}"
```
zzzeid, line 290: This should be rewritten to expose the exception outside the string formatting.

src/lando/pushlog/models/commit.py
```diff
@@ -239,13 +239,52 @@ class Tag(models.Model):
...
     )
 
+    _scm_commit: CommitData | None = None
+
+    @staticmethod
+    def for_scm_commit(repo: Repo, name: str, scm_commit: CommitData):  # noqa: ANN205
+        """Create a partial Tag object, based on CommitData not present in the DB
+        yet."""
+        try:
+            # If a commit already exists in the DB, use that
```
zzzeid, line 249:
```suggestion
            # If a commit already exists in the DB, use that.
```

src/lando/pushlog/pushlog.py
```diff
@@ -58,22 +58,30 @@ class PushLog:
...
         self.repo = repo
         self.user = user
 
         if not commits:
-            # We cannot use the default value of the argument as a mutable type, and will
-            # get reused on every initialisation.
+            # We cannot use the default value of a mutable type argument,
+            # as it will get reused on every initialisation.
             commits = []
         self.commits = commits
```
zzzeid, line 77: Same as Connor's comment below.

src/lando/pushlog/models/commit.py
```diff
@@ -239,13 +239,52 @@ class Tag(models.Model):
...
+        If the target commit is not an ORM model yet, this method will find or create
+        it as needed, and maintain the DB relations.
+        """
+        # First make sure we don't deal with stale data.
+        try:
+            self.refresh_from_db()
+        except Tag.DoesNotExist:
+            # We're OK if this commit doesn't exist in the DB yet; we're just about to
+            # write it.
+            pass
```
zzzeid, line 272: This is a little strange here. What's the purpose of this? I.e., what data could be stale?

I think there might be some unexpected behaviour in the future, since it will revert any edits to the instance before saving it. Regardless, better to check for `self.pk` than to catch an exception here.

src/lando/pushlog/models/commit.py
```diff
@@ -239,13 +239,52 @@ class Tag(models.Model):
...
+
+        if self._scm_commit:
+            self.commit = Commit.from_scm_commit(self.repo, self._scm_commit)
+            self._scm_commit = None
+
+        super(Tag, self).save(*args, **kwargs)
```
zzzeid, line 278:
```suggestion
        super().save(*args, **kwargs)
```

zzzeid (changes requested):

src/lando/conftest.py
shtrom: Yep. In this case, it's not at risk because I'm iterating over the elements of the list, rather than owning/modifying it, but fair enough.

Probably one of my biggest annoyance with Python 😅

src/lando/pushlog/models/commit.py
shtrom: We are dealing with a weird case here: we have a partially-created model that hasn't be saved yet, so trying to deal with relations to other raises that exception, but it's not a problem. I tried just inspecting `self.commit` to see if it was None-ish, but even that raised the exception... So I ended up just capturing the exception as a poor-man's conditional. 

Note that, with https://bugzilla.mozilla.org/show_bug.cgi?id=1957526, this problem will go away.

src/lando/pushlog/models/commit.py
shtrom: I need to be use words with fewer letters 😅

src/lando/pushlog/models/commit.py
shtrom: As we create the object in memory before committing it to the DB (so we have something we can create relations to [which won't work until we commit it to the DB anyway...]), the refresh serves to fetch the latest version (which there shouldn't be because we only have a single worker processing a single job at any time), _before_ we make any local modification further down the line.

So... Yeah, it does look like overkill, as we're enabling something that we don't do, and protecting against something that's structurally no gonna happen.

cgsheeh (approved):
LGTM.

PR #297, opened 2025-04-22
git: re-raise exceptions from Git in `apply_patch` (Bug 1961715)

cgsheeh:
Exceptions thrown by `_git_run` while applying patches are silently
swallowed while trying to convert them to `PatchConflict` exceptions.
Re-raise them here so any non-standard errors are seen.


cgsheeh:
This is failing because it catches an issue in tests, which is fixed in #292.

shtrom (approved):

PR #299, opened 2025-04-23
pushlog: make tags read-only in admin

shtrom:


shtrom:
> ## Pull Request Overview

No, not you, go away.

cgsheeh (approved):

PR #300, opened 2025-04-23
pushlog: new command pushlog_add_gap (bug 1962039)

shtrom:


zzzeid (approved):
It's unclear from the bug or this PR under what circumstances this command should be used. Would be helpful to add something about that.

PR #302, opened 2025-04-23
headless_api: store Pydantic models in DB using `.model_dump(mode="json")` (Bug 1962191)

cgsheeh:
Using `.dict()` converts the Pydantic models to a `dict`, but does
not properly convert Python types like `datetime` into a JSON
serializeable format. Switch to using `.model_dump(mode="json")`
which does the correct conversion.

### DIFF
--- a/src/lando/headless_api/tests/test_automation_api.py
+++ b/src/lando/headless_api/tests/test_automation_api.py
@@ -377,7 +377,7 @@ def test_automation_job_create_commit_request(client, repo_mc, headless_user):
             {
                 "action": "create-commit",
                 "commitmsg": "test commit message",
-                "date": "2025-04-22T18:30:27.786900+00:00",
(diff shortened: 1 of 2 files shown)


src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -367,6 +367,37 @@ def test_automation_job_create_api(client, hg_server, hg_clone, headless_user):
...
+        content_type="application/json",
+        headers={
+            "User-Agent": "Lando-User/testuser@example.org",
+            "Authorization": f"Bearer {token}",
+        },
+    )
+
+    assert (
+        response.status_code == 202
+    ), "Successful submission should result in `202 Accepted` status code."
```
shtrom, line 398: Should we assert that the `date` in the model data is a string here?

shtrom (approved):

PR #305, opened 2025-04-24
phabricator: support limits when collating (bug 1957891)

zzzeid:
- ignore limit when calling call_conduit
- return data based on original limit after collating

For queries that are expected to return greater than 100 results
setting the limit does not make sense as Phabricator limits results
to 100 anyway. So in those cases, we ignore the limit set, and then
discard any excess records. This is done for compatibility at this
time since various methods currently set a limit.

### DIFF
--- a/src/lando/utils/phabricator.py
+++ b/src/lando/utils/phabricator.py
@@ -214,6 +214,15 @@ class PhabricatorClient:
 
     def call_conduit_collated(self, method: str, **kwargs) -> dict[str, list[Any]]:
         """Continuously call call_conduit and return the collated data in one dict."""
+
+        # NOTE: if a limit is passed, it is ignored since Phabricator already limits
+        # results to 100, and raises and error if a limit is passed that is greater
+        # than 100. This is kept for backwards compatibility.
+        # See bug 1962523 for more information.
+        limit = None
+        if "limit" in kwargs:
+            limit = kwargs.pop("limit")
+
         result = self.call_conduit(method, **kwargs)
         if not result:
             return {"data": []}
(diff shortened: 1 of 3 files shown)


zzzeid:
Filed [bug 1962523](https://bugzilla.mozilla.org/show_bug.cgi?id=1962523) to track deprecating the use of "limits" altogether.

src/lando/utils/phabricator.py
```diff
@@ -225,6 +234,8 @@ def call_conduit_collated(self, method: str, **kwargs) -> dict[str, list[Any]]:
             )
             if result and "data" in result and result["data"]:
                 data += result["data"]
+        if limit:
+            data = data[:limit]
```
shtrom, line 238: It looks like we might miss out on data beyond the limit here.

Is there a way to determine that more data was available? Or will the caller just loop until no data is returned?

zzzeid: That is correct, we're ignoring anything beyond the limit but this is to keep the behaviour as the rest of the code expects it to be. I put more details in [bug 1962523](https://bugzilla.mozilla.org/show_bug.cgi?id=1962523) and in the PR description. If more data was available, we would ignore it just as a standard `call_conduit` call with a limit would also ignore any additional data since in those cases, we only make one call to the API and we don't check additional pages.

In a nut shell, setting a limit beyond 100 is not supported by the API, and currently various parts of the code in Lando rely (in a hacky[1] way) on setting a limit to ensure that the returned data length is as expected (see [here](https://github.com/mozilla-conduit/lando/blob/main/src/lando/api/legacy/stacks.py#L73) for an example).

I think we need to reimplement these checks in order to fix this properly, and then we can get rid of using limits. I.e., replace the following with something else:

```python
if len(revs["data"]) != len(revision_phids):
    raise ValueError("Mismatch in size of returned data.")
```

---

[1] I don't think this code actually does anything other than potentially catch cases where less than the expected number of entries is returned.

shtrom (approved):

PR #307, opened 2025-04-25
middleware: add maintenance message functionality (bug 1957093)

zzzeid:


cgsheeh (approved):
What's the use case for this?

zzzeid:
> What's the use case for this?

I mentioned one in the bug: "This way, more context can be provided to users who encounter the maintenance mode message."

For example, "A new version of Lando is being deployed, please try again shortly."

Or, "Lando is undergoing unscheduled maintenance, expected to take another 30 minutes."

This to make it less necessary for people to reach out on Slack or Matrix asking what is on with Lando, and reduce the need for us to do manual messaging.

cgsheeh:
> > What's the use case for this?
> 
> I mentioned one in the bug: "This way, more context can be provided to users who encounter the maintenance mode message."
> 
> For example, "A new version of Lando is being deployed, please try again shortly."
> 
> Or, "Lando is undergoing unscheduled maintenance, expected to take another 30 minutes."
> 
> This to make it less necessary for people to reach out on Slack or Matrix asking what is on with Lando, and reduce the need for us to do manual messaging.

Ah okay, so we will be manually updating the message via the admin UI. I was wondering if you planned to add this to our deploy scripts in some way. Makes sense to me. :+1:

zzzeid:
> > > What's the use case for this?
> > 
> > 
> > I mentioned one in the bug: "This way, more context can be provided to users who encounter the maintenance mode message."
> > For example, "A new version of Lando is being deployed, please try again shortly."
> > Or, "Lando is undergoing unscheduled maintenance, expected to take another 30 minutes."
> > This to make it less necessary for people to reach out on Slack or Matrix asking what is on with Lando, and reduce the need for us to do manual messaging.
> 
> Ah okay, so we will be manually updating the message via the admin UI. I was wondering if you planned to add this to our deploy scripts in some way. Makes sense to me. 👍

Yeah we could have a default message stored, and if we want to update it manually before triggering a deploy that would be very easy. This is pretty basic but in the future we could have a drop down or something like that.

PR #308, opened 2025-04-25
create_environment_repos: remove prod repos from script (bug 1959193)

zzzeid:
These repos are no longer needed in the script, since it has already been run.

### DIFF
--- a/src/lando/utils/management/commands/create_environment_repos.py
+++ b/src/lando/utils/management/commands/create_environment_repos.py
@@ -5,15 +5,13 @@ from django.db.utils import IntegrityError
 from lando.environments import Environment
 from lando.main.models import (
     DONTBUILD,
-    SCM_ALLOW_DIRECT_PUSH,
     SCM_CONDUIT,
     SCM_LEVEL_1,
-    SCM_LEVEL_3,
     Repo,
 )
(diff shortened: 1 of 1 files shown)


src/lando/utils/management/commands/create_environment_repos.py
```diff
@@ -153,23 +151,8 @@
             "required_permission": SCM_CONDUIT,
         },
     ],
-    Environment.production: [],
 }
 
-for branch in ["autoland", "beta", "esr115", "esr128", "release"]:
```
cgsheeh, line 159: We need to add `firefox-main`, can we use this script for that today before we remove it?

cgsheeh (changes requested):

src/lando/utils/management/commands/create_environment_repos.py
zzzeid: @cgsheeh sure, just edit the script and include only `main` in the list of branches.

zzzeid: You can also manually create it in the admin. This script was mostly to save time and avoid error because there were 5 repos.

cgsheeh (approved):

PR #313, opened 2025-04-30
uplift: pass empty tree hash when creating diff (bug 1963679)

zzzeid:
- Phabricator expects this value to be set for git commits
- Mercurial commits that are copied over do not include this value

See [bug 1963689](https://bugzilla.mozilla.org/show_bug.cgi?id=1963689) for full fix.


cgsheeh (approved):
LGTM.

PR #318, opened 2025-05-01
uplifts: use `repo.default_branch` as repo name for uplifts (Bug 1963870)

cgsheeh:
Use `repo.default_branch` as the repo name for Bugzilla
updates on uplift. In the new unified Git repo model,
the branch names map to the expected values in BMO
for tracking flags.


zzzeid (approved):
LGTM!

I assume there are no more uplifts from "legacy" repos, as for those the `default_branch` is not set.

Also the PR title needs to be updated.

cgsheeh:
> LGTM!
> 
> I assume there are no more uplifts from "legacy" repos, as for those the `default_branch` is not set.
> 
> Also the PR title needs to be updated.

Yes that's right, the old repos are no longer used for uplifts.

PR #321, opened 2025-05-02
admin: improvements to Pushlog, Revision and LandingJob admin pages (bug 1955972)

shtrom:


src/lando/pushlog/admin.py
```diff
@@ -29,14 +86,48 @@ class PushAdmin(PushLogAdmin):
         "branch",
         "datetime",
         "user",
-        "commits",
-        "tags",
+        # "commits",
+        # "tags",
```
cgsheeh, line 90: Why are these commented out?

cgsheeh (approved):

src/lando/pushlog/admin.py
shtrom: They are not directly editable anymore, as they have been added to the `exclude` list, and replaced with the Inline. 

So, tl;dr: no longer needed, will remove.

zzzeid (dismissed):
Minor comments otherwise lgtm.

zzzeid (approved):
The PR title is too long, and will get cropped when you merge, better to put the details in the description.

PR #322, opened 2025-05-05
pushlog_view: add --push-id argument

shtrom:


src/lando/pushlog/management/commands/pushlog_view.py
```diff
@@ -30,9 +32,16 @@ def handle(self, *args, **options):
         except Repo.DoesNotExist:
             raise CommandError(f"Repository not found: {repo_name}")
 
-        pushes = Push.objects.filter(repo=repo).order_by("-push_id")
-        if limit:
-            pushes = pushes[:limit]
+        if push_id > 0:
```
zzzeid, line 35: nit
```suggestion
        if push_id:
```

src/lando/pushlog/management/commands/pushlog_view.py
```diff
@@ -30,9 +32,16 @@ def handle(self, *args, **options):
...
+            pushes = Push.objects.filter(repo=repo, push_id=push_id)
+        else:
+            pushes = Push.objects.filter(repo=repo).order_by("-push_id")
+            if limit:
+                pushes = pushes[:limit]
+
+        if not pushes:
+            push_id_info = f" (push_id {push_id})" if push_id > 0 else ""
```
zzzeid, line 43: nit

```suggestion
            push_id_info = f" (push_id {push_id})" if push_id else ""
```

zzzeid (approved):

src/lando/pushlog/management/commands/pushlog_view.py
shtrom: Ah, but it's a user-controlled unsigned int. If the user passes, say, `-1`, I'd rather just show the full pushlog, rather than failing to find the non-existent push with id `-1`.

OTOH, we do have a clean way to report not-found log, so it wouldn't be a big issue 🤔

PR #325, opened 2025-05-06
automation_job: add error field (bug 1963807)

zzzeid:
- error is already being set on the instance
- add field to capture the actual error

See https://github.com/mozilla-conduit/lando/blob/main/src/lando/headless_api/models/automation_job.py#L115.


cgsheeh (approved):
Nice, I think I have a patch to add this somewhere as well. :sweat_smile: We will need to add display of this field to the `lando_cli` as well.

PR #326, opened 2025-05-07
headless_api: add support for RelBranches (Bug 1964716)

cgsheeh:
Add support for creating and updating commits on RelBranches
to the headless API. Add a new top-level `relbranch` field to
the `AutomationOperation` schema, which takes a `RelBranchSpecifier`
object. This object contains the name of the relbranch to push
changes to, and an optional `commit_sha` which is used as the
base when creating a new RelBranch.

The values in the `relbranch` field are saved as a new JSON
column in the `AutomationJob`. Before running the job, we
reference this field to determine if we should use RelBranch
specific base commits to apply actions to, and a RelBranch
specific push target ref. Since it is possible for the
branch to be new, and therefore not have an upstream tracking
branch to compare to, we update `scm.describe_local_changes`
with an optional argument which indicates the base commit
to compare against to find local changes.


src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -83,9 +83,12 @@ def run_automation_job(self, job: AutomationJob) -> bool:
             scm.for_push(job.requester_email),
             PushLogForRepo(repo, job.requester_email) as pushlog,
         ):
+            # Determine if a RelBranch should be used for the push.
+            target_cset, push_target = job.resolve_push_target_from_relbranch(repo)
```
zzzeid, line 87: nit: I think it would be better to have separate methods that each return `target_cset` and `push_target` separately, to move away from the paradigm of return and/or passing tuples that contain multiple things which the legacy Lando had a lot of.

src/lando/headless_api/models/automation_job.py
```diff
@@ -49,6 +49,9 @@ class AutomationJob(BaseModel):
     # Text describing errors when status != LANDED.
     error = models.TextField(default="", blank=True)
 
+    # Specification of RelBranch to push changes to.
+    relbranch_specifier = models.JSONField(null=True, blank=True)
```
zzzeid, line 53: Since this appears to be for only two fields, one of which is optional, why not have two separate fields rather than a JSON field? e.g. `relbranch_name` and `relbranch_commit_sha` or something similar. I think this would simplify the rest of the implementation.

src/lando/main/scm/hg.py
```diff
@@ -332,7 +332,7 @@ def describe_commit(self, revision_id: str = ".") -> CommitData:
         """Return Commit metadata."""
         return self._describe_revisions(revision_id)[0]
 
-    def describe_local_changes(self) -> list[CommitData]:
+    def describe_local_changes(self, **kwargs) -> list[CommitData]:
```
zzzeid, line 335: Looks like there is a divergence here in terms of what this method expects and uses in git vs. hg. Can this be made more explicit?

zzzeid (changes requested):
Looks good, just a couple of minor comments.

It looks like this change is mainly for git repos. Wondering how we should start differentiating between things supported in both SCM types vs. only one, since there is no obvious way to see that right now apart from which ones have tests.

src/lando/main/scm/hg.py
```diff
@@ -332,9 +332,12 @@ def describe_commit(self, revision_id: str = ".") -> CommitData:
...
         return self._describe_revisions(revision_id)[0]
 
-    def describe_local_changes(self) -> list[CommitData]:
+    def describe_local_changes(
+        self, target_cset: str | None = None
+    ) -> list[CommitData]:
         """Return a list of the Commits only present on this branch."""
-        return list(self._describe_revisions("::. and draft()"))
+        base = target_cset if target_cset else ""
+        return list(self._describe_revisions(f"{base}::. and draft()"))
```
shtrom, line 340: ... like this:

```suggestion
        return list(self._describe_revisions(f"{base_cset}::. and draft()"))
```

shtrom (commented):
LGTM. Couple of questions, nothing major. Main point: `base_cset` seems more descriptive than `target_cset` for `describe_local_changes`.

src/lando/main/scm/git.py
```diff
@@ -257,9 +257,17 @@ def describe_commit(self, revision_id: str = "HEAD") -> CommitData:
...
-        return list(reversed(self._describe_commits("@{u}..")))
+    def describe_local_changes(
+        self, target_cset: str | None = None
+    ) -> list[CommitData]:
+        """Return a list of the Commits only present on this branch.
+
+        Use the passed target changeset as the base commit. Otherwise, use the
+        configured upstream branch.
+        """
+        refspec = f"{target_cset}.." if target_cset else "@{u}.."
```
shtrom, line 268: ... for simplicity with the previous suggested change:

```suggestion
        refspec = f"{base_cset}.."
```

shtrom (commented):
Small addition for simplicity.

src/lando/headless_api/api.py
```diff
@@ -298,12 +298,25 @@ def resolve_action(action_data: dict) -> Action:
...
 
 
+class RelBranchSpecifier(Schema):
+    """Metadata requried to specify the RelBranch for pushing."""
+
+    # Name of the RelBranch for pushing.
+    branch_name: str
+
+    # Commit to point the RelBranch to, if it does not exist yet.
+    commit_sha: str | None = None
```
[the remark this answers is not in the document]
cgsheeh: The user-side looks for the `origin/<relbranch name>` remote branch, so there's no way for the user-side to see it without it existing on the server. If the user side didn't provide the `commit_sha` and the branch didn't exist, Lando would fail to find the remote branch and the job would fail, which is the behaviour we expect.

src/lando/main/scm/git.py
```diff
@@ -364,6 +367,11 @@ def update_repo(self, pull_path: str, target_cset: Optional[str] = None) -> str:
             cwd=self.path,
         )
 
+        remote_branch = f"origin/{target_cset}"
+        if self._git_run("branch", "--list", "--remote", remote_branch, cwd=self.path):
+            # If the branch exists remotely, make sure we get the up-to-date version.
+            target_cset = remote_branch
```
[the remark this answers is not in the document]
cgsheeh: In this case we're just asserting the remote branch exists, not inspecting the SHA our local copy thinks the remote branch points at. Fetching earlier here ensures we have all the remote branches, since the remote branch isn't created locally on push due to the way we push branches. We were still fetching and using the up-to-date version of the remote branch.

shtrom (approved):
LGTM, but see @zzeid's nit about `resolve_push_target_from_relbranch`

src/lando/headless_api/models/automation_job.py
```diff
@@ -49,8 +49,11 @@ class AutomationJob(BaseModel):
     # Text describing errors when status != LANDED.
     error = models.TextField(default="", blank=True)
 
-    # Specification of RelBranch to push changes to.
-    relbranch_specifier = models.JSONField(null=True, blank=True)
+    # Name of RelBranch to push changes to.
+    relbranch_name = models.JSONField(null=True, blank=True)
```
zzzeid, line 53:
```suggestion
    relbranch_name = models.CharField(null=True, blank=True)
```

zzzeid (changes requested):

PR #327, opened 2025-05-08
pushlog_view: filter by commits- or tags-only (bug 1963805)

shtrom:


src/lando/pushlog/management/commands/pushlog_view.py
```diff
@@ -35,19 +44,37 @@ def handle(self, *args, **options):
         if push_id:
             pushes = Push.objects.filter(repo=repo, push_id=push_id)
         else:
-            pushes = Push.objects.filter(repo=repo).order_by("-push_id")
+            pushes = Push.objects.filter(repo=repo)
+
+            if commits_only:
+                if tags_only:
+                    self.stdout.write(
```
cgsheeh, line 51: Seems like these arguments should be in a [mutually exclusive group](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) instead.

src/lando/pushlog/tests/test_pushlog_view.py
```diff
@@ -83,8 +83,33 @@ def test_pushlog_view_single_push(
     }
 
     call_command("pushlog_view", stdout=out, **opts)
+    output = out.getvalue()
+
+    assert str(commit) in output
+    assert str(tag) not in output, "Other Pushes than selected found in outpu"
```
cgsheeh, line 89:
```suggestion
    assert str(tag) not in output, "Other Pushes than selected found in output."
```

cgsheeh (approved):

PR #334, opened 2025-05-16
automation_worker: add checks before pushing (bug 1964556)

shtrom:


src/lando/main/tests/test_git.py
```diff
@@ -133,6 +136,46 @@ def test_GitSCM_clean_repo(
...
+    # with a different base commit.
+    trim_from_re = r"^From [^\n]+\n"
+    expected_patch = re.sub(trim_from_re, "", expected_patch, count=1)
+    new_patch = re.sub(trim_from_re, "", new_patch, count=1)
+
+    # The git version used when we created our fixture may no longer match the one on
+    # the system running the test.
+    unify_git_version_re = r"\d+(\.\d+)+$"
+    expected_patch = re.sub(unify_git_version_re, "GIT.VERS.ION", expected_patch)
+    new_patch = re.sub(unify_git_version_re, "GIT.VERS.ION", new_patch)
```
zzzeid, line 167: nit: I think it would be cleaner to explicitly ignore this, rather than modifying it in the expected output. This may mean that we would be going through the patch a little differently when comparing, rather than what you have on line 176.

src/lando/main/tests/test_git.py
```diff
@@ -133,6 +136,46 @@ def test_GitSCM_clean_repo(
...
+
+    # The original patch may have a `[PATCH]` in the subject that we don't want to
+    # retain on application and subsequent export.
+    expected_patch = re.sub(r"Subject: \[PATCH\]", "Subject:", expected_patch, count=1)
```
zzzeid, line 171: Similar comment here.

zzzeid (changes requested):
Most nits (except one comment). Otherise lgtm.

src/lando/main/tests/test_git.py
shtrom: Keeping discussion in previous mention.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -715,6 +734,141 @@ def test_automation_job_create_commit_success(
...
+    automation_worker = get_automation_worker(scm_type)
+
+    automation_worker.worker_instance.applicable_repos.add(repo)
+
+    # Mock `phab_trigger_repo_update` so we can make sure that it was called.
+    mock_trigger_update = mock.MagicMock()
+    monkeypatch.setattr(
+        "lando.api.legacy.workers.automation_worker.AutomationWorker.phab_trigger_repo_update",
+        mock_trigger_update,
+    )
```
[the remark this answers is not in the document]
shtrom: https://bugzilla.mozilla.org/show_bug.cgi?id=1969192

src/lando/api/legacy/hgexports.py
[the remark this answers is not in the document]
shtrom: https://bugzilla.mozilla.org/show_bug.cgi?id=1969347

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -229,18 +231,20 @@ def run_automation_job(self, job: AutomationJob) -> bool:
...
         """Run landing checks on the stack of commits generated by the job.
 
         Returns a list of error messages.
         """
-        patch_helpers = []
-        # Create PatchHelpers from commits
-        for commit in commits:
-            patch_helpers.append(scm.get_patch_helper(commit.hash))
+        patch_helpers = filter(
+            lambda x: x is not None,
```
zzzeid, line 241: Using a variable name other that `x` would be better here, and also would be good to add a comment as to when the argument is expected to be None.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -734,91 +736,128 @@ def test_automation_job_create_commit_success(
     assert len(job.landed_commit_id) == 40, "Landed commit ID should be a 40-char SHA."
 
 
+BAD_ACTION_TYPES = [
```
zzzeid, line 739: nit: Can we use something more specific than `BAD_` as a prefix? Maybe `INVALID_` or `DISALLOWED_` or something else more specific.

PR #337, opened 2025-05-23
scm.git: add colon to userinfo regex (bug 1965019)

zzzeid:


shtrom (approved):
LGTM, but +1 for the test.

cgsheeh (approved):

PR #339, opened 2025-05-27
landing_worker: add pre- and post-landing blocking checks (bug 1968614)

shtrom:


cgsheeh (changes requested):
For regular landings, the checks should run via the `run_landing_checks` API so users see the issues with their landing request and fix them before attempting to land. See [`blocker_prevent_symlinks`](https://github.com/mozilla-conduit/lando/blob/main/src/lando/api/legacy/transplants.py#L874) for an example of how to use the `DiffAssessor` on a `StackAssessmentState` to present the error as a blocker at landing time.

shtrom:
@cgsheeh @zzzeid this is good for another look, particularly the use of `run_landing_checks` in the transplants

It conflicts with its base branch ATM, as I needed to change things in #334, but I'll merge it back in the end.

shtrom:
@zzzeid @cgsheeh good for another look (but start with #334)

src/lando/api/legacy/transplants.py
```diff
@@ -871,6 +872,35 @@ def blocker_single_landing_repo(
     )
 
 
+def blocker_prevent_nsprnss_files(
+    revision: dict, diff: dict, stack_state: StackAssessmentState
+) -> Optional[str]:
+    """Block revisions which contain changs to the NSS or security directories."""
```
zzzeid, line 878:
```suggestion
    """Block revisions which contain changes to the NSS or security directories."""
```

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -338,10 +361,23 @@ def apply_and_push(
         # Get the changeset hash of the first node.
         commit_id = scm.head_ref()
 
+        new_commits = scm.describe_local_changes()
+
+        check_errors = self.run_landing_checks(scm, new_commits)
+
+        if check_errors:
+            message = "Some checks weren't successful:\n" + "\n".join(check_errors)
```
zzzeid, line 369: nit
```suggestion
            message = "Some checks failed before attempting to land:\n" + "\n".join(check_errors)
```

src/lando/api/tests/test_landings.py
```diff
@@ -553,6 +555,78 @@ def test_merge_conflict(
...
 
 
+@pytest.mark.parametrize(
+    "repo_type,bad_commit_type",
+    # We make a cross-product of all the SCM and all the bad actions.
+    # As we don't want a cross-product of bad actions and reasons, we bundle them in a
+    # tuple, that we deconstruct in the test.
+    itertools.product(
+        [SCM_TYPE_HG, SCM_TYPE_GIT],
+        # All of BAD_COMMIT_TYPES, but not wpt
```
zzzeid, line 565: I think this and maybe other instances of "bad" may be outdated based on the target branch.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -867,18 +702,24 @@ def test_automation_job_create_commit_failed_check(
...
 ):
-    bad_action, reason = failed_check_action_reason(bad_action_reason)
+    bad_action, reason = failed_check_reason(bad_action_type)
+    # Warning: I suspect, but haven't been able to confirm, that doing this
+    # may directly modify the data in the fixture. If, in the future, the
+    # failed_check_reason fixtures appears to contain weird unexpected content in other
+    # test cases, look with suspicion towards this.
+    bad_action["action"] = "create-commit"
+    bad_action["date"] = 0
+    bad_action["diff"] = check_diff(bad_action_type)
```
zzzeid, line 717: Would be better to pass these as arguments to the fixture, and override them in the fixture instead (common idiom). `bad_action_type` is passed twice, once to `failed_check_reason` which returns `bad_action`, and then again to override the value in `bad_action["diff"]`. This can be cleaned up I think.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -937,22 +778,32 @@ def test_automation_job_create_commit_failed_check_override(
...
         status=JobStatus.SUBMITTED,
         requester_email="example@example.com",
         target_repo=repo,
     )
+
+    # Create a failing commit.
+    no_bug_action = failed_check_reason("nobug")[0]
+    no_bug_action["action"] = "create-commit"
+    no_bug_action["date"] = 0
+    no_bug_action["diff"] = check_diff("valid")
```
zzzeid, line 801: Same comment here re: passing arguments to the fixture to clean this up.

zzzeid (changes requested):
nit: PR title should be a bit more specific.

src/lando/headless_api/tests/test_automation_api.py
shtrom: I got myself into a position where I have slightly different needs from the fixture, so the second need extends the first one... But I think I can just create a second fixture to hide this logic in.

cgsheeh (dismissed):
From the code POV this LGTM. :)

We should fix the typos and remove the commented out code before landing.

PR #340, opened 2025-05-27
api: add support for bidirectional mapping (bug 1963822)

zzzeid:
- add CommitMap model and migration file
- add various methods to support populating the DB
- add command that pre-populates the DB
- add methods that query pushlog to catch up on missing data
- add bidirectional views
- add tests

### DIFF
--- a/src/lando/conftest.py
+++ b/src/lando/conftest.py
@@ -907,6 +907,7 @@ def commit_maps():
     return list(CommitMap.objects.all().order_by("id"))
 
 
+@pytest.fixture
 def mock_landing_worker_phab_repo_update(monkeypatch):
     mock_trigger_update = mock.MagicMock()
     monkeypatch.setattr(


cgsheeh (commented):
My initial thought is we should use Django-Ninja to define these endpoints for consistency with the headless API. Any reason we shouldn't use that here?

zzzeid:
> My initial thought is we should use Django-Ninja to define these endpoints for consistency with the headless API. Any reason we shouldn't use that here?

We could, it's just not part of the api app currently. I didn't want to put these endpoints in the headless_api app since it is not related to automation. I'm hesitant to mix two different types of endpoints (e.g., standard "class based" views, and ninja ones) given that there are other endpoints in this app that do not use ninja. Also, I'm taking advantage of class based views in this case.

I'll give it some thought, though since there is no handling of form data, I don't know what advantage using ninja would provide here.

src/lando/utils/management/commands/process_git_hg_mapping_file.py
```diff
@@ -0,0 +1,64 @@
...
+
+import requests
+from django.core.management.base import BaseCommand
+from django.db.utils import IntegrityError
+
+from lando.main.models import CommitMap
+
+
+class Command(BaseCommand):
+    help = "Download and process the git to hg mapping file" ""
```
cgsheeh, line 14:
```suggestion
    help = "Download and process the git to hg mapping file"
```

src/lando/api/tests/test_views.py
```diff
@@ -0,0 +1,47 @@
+from unittest import mock
+
+import pytest
+
+
+@pytest.mark.django_db(transaction=True)
+def test__views__git2hgCommitMapView(commit_maps, authenticated_client, monkeypatch):
+    mock_catch_up = mock.MagicMock()
+    monkeypatch.setattr("lando.api.views.CommitMap.catch_up", mock_catch_up)
+    response = authenticated_client.get(f"/api/git2hg/test_git_repo/{'a' * 40}")
```
cgsheeh, line 10: Tests should be using the unauthenticated client, since consumers in automation won't be authenticated.

src/lando/api/tests/test_views.py
```diff
@@ -0,0 +1,47 @@
...
+    assert response.status_code == 200
+    assert response.json() == {"git_hash": "a" * 40, "hg_hash": "b" * 40}
+
+
+@pytest.mark.django_db(transaction=True)
+def test__views__hg2gitCommitMapView(commit_maps, authenticated_client, monkeypatch):
+    mock_catch_up = mock.MagicMock()
+    monkeypatch.setattr("lando.api.views.CommitMap.catch_up", mock_catch_up)
+    response = authenticated_client.get(f"/api/hg2git/test_git_repo/{'b' * 40}")
```
cgsheeh, line 19: Could we reference a known good SHA from `commit_maps` here instead of hard-coding the SHA?

src/lando/api/views.py
```diff
@@ -104,3 +110,40 @@ class Form(forms.Form):
             )
 
         return JsonResponse({"errors": dict(form.errors)}, status=400)
+
+
+class CommitMapBaseView(View):
+    """CommitMap base view to be extended for bidirectional git - hg mapping."""
+
```
cgsheeh, line 117: Should we define `scm = None` as a default here?

src/lando/utils/management/commands/process_git_hg_mapping_file.py
```diff
@@ -0,0 +1,64 @@
...
+        for row in self.rows:
+            try:
+                CommitMap.objects.create(
+                    git_hash=row["git"],
+                    hg_hash=row["hg"],
+                    git_repo_name="firefox",
+                )
+            except IntegrityError:
+                self.stdout.write(f"Skipped {row}.")
+                skipped_count += 1
```
cgsheeh, line 59: What might cause us to skip rows here? Do we have a plan if some rows don't import correctly?

src/lando/main/models/commit.py
```diff
@@ -0,0 +1,82 @@
...
+
+class CommitMap(BaseModel):
+    """Map a git hash to an hg hash, based on a specific repo."""
+
+    HGMO_PUSHLOG_TEMPLATE = "https://hg.mozilla.org/{}/json-pushes"
+    REPO_MAPPING = (("firefox", "mozilla-unified"),)
+
+    git_hash = models.CharField(default="", unique=True)
+    hg_hash = models.CharField(default="", unique=True)
+    git_repo_name = models.CharField(default="")
```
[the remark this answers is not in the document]
cgsheeh: Makes sense to me. Could you add a comment about this for future reference?

src/lando/main/models/commit_map.py
```diff
@@ -0,0 +1,84 @@
...
+        for push in list(push_data.values()):
+            hg_changesets = push["changesets"]
+            git_changesets = push["git_changesets"]
+            if len(hg_changesets) != len(git_changesets):
+                raise ValueError(
+                    "Number of hg changesets does not match number of git changesets: "
+                    f"{len(hg_changesets)} vs {len(git_changesets)}"
+                )
+            for hg_changeset, git_changeset in zip(
+                hg_changesets, git_changesets, strict=False
```
zzzeid, line 76: note: `strict=True` fails only when a length mismatch occurs during an iteration, but not before iterating.

src/lando/main/models/commit.py
```diff
@@ -0,0 +1,82 @@
...
+        # NOTE: multiple changesets may be included in the response.
+        for push in list(push_data.values()):
+            changesets = push["changesets"]
+            git_changesets = push["git_changesets"]
+            if len(changesets) != len(git_changesets):
+                raise ValueError(
+                    "Number of changesets does not match number of git changesets: "
+                    f"{len(changesets)} vs {len(git_changesets)}"
+                )
+            for i in range(len(changesets)):
```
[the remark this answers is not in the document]
cgsheeh: Nice, that's much better. :) Usually `range(len(thing))` is a giveaway that there's a cleaner way to run the iteration.

zzzeid:
I will update the mapping file and merge + deploy this on Monday.

src/lando/api/views.py
```diff
@@ -104,3 +110,45 @@ class Form(forms.Form):
...
+    scm = None
+
+    def get(
+        self, request: WSGIRequest, git_repo_name: str, commit_hash: str
```
shtrom, line 121: nit: depending on the file, we have type annotations of `request` as `django.http.HttpRequest` or this `WSGIRequest`. We should probably try to choose one for consistency, though I'm not sure which one, as I don't know what the difference is off the bat, do you know?

src/lando/api/views.py
```diff
@@ -104,3 +110,45 @@ class Form(forms.Form):
...
+    ) -> JsonResponse:
+        hash_field = f"{self.scm}_hash"
+        filters = {hash_field: commit_hash, "git_repo_name": git_repo_name}
+        commits = CommitMap.objects.filter(**filters)
+
+        if not commits.exists():
+            CommitMap.catch_up(git_repo_name)
```
shtrom, line 128: How long does this take? Is it likely to last more than a request lifetime, and if so what would happen if the client request dies? Can we leapfrog?

Maybe it's something we could delegate to a celery worker?

src/lando/main/models/commit_map.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class CommitMap(BaseModel):
+    """Map a git hash to an hg hash, based on a specific repo."""
+
+    HGMO_PUSHLOG_TEMPLATE = "https://hg.mozilla.org/{}/json-pushes"
+    REPO_MAPPING = (("firefox", "mozilla-unified"),)
+
+    git_hash = models.CharField(default="", unique=True)
+    hg_hash = models.CharField(default="", unique=True)
```
shtrom, line 15: I think we need the unique constraints to cover the git_repo (git_repo first, then hash).

src/lando/main/models/commit_map.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # NOTE: This value is set because multiple Lando repos can map to a single hg repo.
+    # This is because currently a separate repo object is created for each branch of a
+    # git repo, though those might map to a single hg repo (e.g., mozilla-unified).
+    git_repo_name = models.CharField(default="")
```
shtrom, line 20: Does this mean we may have the same hash mapping multiple times (once per repo)?

(My previous comment was just about preventing a serendipitous hash collision, but this may precipitate the issue.)

src/lando/main/models/commit_map.py
```diff
@@ -0,0 +1,90 @@
...
+
+from lando.main.models import BaseModel
+from lando.main.models.repo import Repo
+
+
+class CommitMap(BaseModel):
+    """Map a git hash to an hg hash, based on a specific repo."""
+
+    HGMO_PUSHLOG_TEMPLATE = "https://hg.mozilla.org/{}/json-pushes"
+    REPO_MAPPING = (("firefox", "mozilla-unified"),)
```
shtrom, line 12: Could we put this in the `settings`? (The `REPO_MAPPING` will be tricky... Maybe some mapping such as `REPO_MAPPING_firefox=mozilla-unified` that we parse and put back into an array of tuples.

src/lando/main/models/commit_map.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # NOTE: This value is set because multiple Lando repos can map to a single hg repo.
+    # This is because currently a separate repo object is created for each branch of a
+    # git repo, though those might map to a single hg repo (e.g., mozilla-unified).
+    git_repo_name = models.CharField(default="")
+
+    @classmethod
+    def get_hg_repo_name(cls, git_repo_name: str) -> str:
+        """Return mapped repo name or `git_repo_name` by default."""
+        return dict(cls.REPO_MAPPING).get(git_repo_name, git_repo_name)
```
shtrom, line 25: Could we make the mapping a dict to start with?

src/lando/main/models/commit_map.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @classmethod
+    def get_pushlog_url(cls, git_repo_name: str) -> str:
+        """Return pushlog URL based on provided repo name."""
+        return cls.HGMO_PUSHLOG_TEMPLATE.format(cls.get_hg_repo_name(git_repo_name))
+
+    @classmethod
+    def _find_last_node(cls, git_repo_name: str) -> "CommitMap":
+        """Return last CommitMap object that was stored for given repo."""
+        return cls.objects.filter(git_repo_name=git_repo_name).latest("created_at")
```
shtrom, line 40: nit: this is the creation date in the DB, rather than than of the commit, and this is loosely related to the topological order of the commits depending on how the insertion is done.

I had the same issue in the PushLog, and decided that it was fine, but it's worth keeping in mind and making the assumption explicitely.

src/lando/utils/management/commands/process_git_hg_mapping_file.py
```diff
@@ -0,0 +1,64 @@
...
+        self._prepare_rows()
+        count = 0
+        skipped_count = 0
+
+        for row in self.rows:
+            try:
+                CommitMap.objects.create(
+                    git_hash=row["git"],
+                    hg_hash=row["hg"],
+                    git_repo_name="firefox",
```
shtrom, line 55: Same issue re: commit date/topological order vs created_at. Perhaps we could separate `CommitMap.fetch` and `CommitMap.push` so we can use the latter twice, rather than writing similar code.

src/lando/main/models/commit_map.py
```diff
@@ -0,0 +1,90 @@
...
+                )
+
+            for hg_changeset, git_changeset in zip(
+                hg_changesets, git_changesets, strict=True
+            ):
+                params = {
+                    "hg_hash": hg_changeset,
+                    "git_hash": git_changeset,
+                    "git_repo_name": git_repo_name,
+                }
```
shtrom, line 88: Could we set the `created_at` value to match the timestamp of the commit, or use a dedicated column for this commit information? Otherwise, it will be the creation date of the object in the DB.

src/lando/utils/management/commands/process_git_hg_mapping_file.py
```diff
@@ -0,0 +1,64 @@
...
+            f.extractall(Path("/tmp"))
+
+        with file_path.open("r") as f:
+            self.rows += list(csv.DictReader(f))
+
+        # Download second (remainder) backup file.
+        url = (
+            "https://gist.github.com/zzzeid/8442141e8b12dfdede1f325a42b7d0f7/"
+            "raw/4672ef4820b401a3157f28be277b70398cb057a6/git2hg_2.csv"
+        )
```
shtrom, line 41: Could we make this an argument and/or a setting?

src/lando/main/models/commit_map.py
shtrom: Ah, also, we know the length of a commit SHA, so we could limit the column size. I'd also suggest not having a defaultand, as we'd run into the uniqueness constraint issue again.

shtrom (changes requested):
A few comments. Mostly nits, though I think it's necessary to fix the DB uniqueness constraints.

src/lando/main/models/commit_map.py
zzzeid: It is possible but unlikely (mainly if the two repos are for different products, i.e., different repos on hgmo).

src/lando/main/models/commit_map.py
zzzeid: See https://github.com/mozilla-conduit/lando/pull/340#discussion_r2159131505.

src/lando/main/models/commit_map.py
zzzeid: I don't think there is much to be gained from this (see my other comment about the purpose of `created_at` when sorting). Doing this would actually make things more challenging to sort.

src/lando/api/views.py
shtrom: As a nit, I'd say we should stick to the HttpRequest superclass, unless we are using things that are specific to WSGI.

src/lando/main/models/commit_map.py
shtrom: Worse comes to worse, it's one deployment away (:

src/lando/api/views.py
zzzeid: The entire app is wired to be a wsgi one. So in effect, things are quite specific to wsgi even though in theory we could change things in the future to support asgi. That is to say, this argument will never be anything other than WSGIRequest unless we explicitly change the structure of this project.

I am not too picky here and think both work, but I do think that WSGIRequest is a more appropriate type especially since we never actually call any of these methods directly, but they are handled by the dispatcher.

Happy to do either, so we can make this more consistent but let's do that in a separate PR.

src/lando/main/models/commit_map.py
```diff
@@ -74,9 +74,11 @@ def fetch_push_data(cls, git_repo_name: str, **kwargs) -> dict:
 
         # We don't care about the key, as it is just the push ID.
         # NOTE: multiple changesets may be included in the response.
-        for push in list(push_data.values()):
-            hg_changesets = push["changesets"]
-            git_changesets = push["git_changesets"]
+
+        pushes = sorted(push_data.keys())
+        for push_id in pushes:
```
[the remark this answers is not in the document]
zzzeid: I don't think it makes a difference, but I added this to be more explicit so that we know a little more explicitly what is happening in this call.

PR #341, opened 2025-05-28
setup_dev: create automation workers

shtrom:


src/lando/utils/management/commands/setup_dev.py
```diff
@@ -19,28 +21,35 @@ def setup_workers(self):
...
+                workers[worker_name] = worker
 
         for repo in Repo.objects.all():
             # Associate all repos with applicable worker.
-            self.stdout.write(
-                f"Adding {repo} ({repo.scm_type}) to {workers[repo.scm_type]}."
-            )
-            workers[repo.scm_type].applicable_repos.add(repo)
+            for worker in workers.values():
+                if not worker or worker.scm != repo.scm_type:
```
zzzeid, line 45: What's the `not worker` check for? Presumably, `workers` contains non-null values so this feel redundant.

src/lando/utils/management/commands/setup_dev.py
```diff
@@ -19,28 +21,35 @@ def setup_workers(self):
...
+                    continue
+                self.stdout.write(f"Adding {repo} ({repo.scm_type}) to {worker}.")
+                worker.applicable_repos.add(repo)
```
zzzeid, line 48: Would this be adding all repos to all workers (e.g., landing and automation?) Maybe OK for time being, but a more realistic setup would be to only have a subset of repos be added to either type of worker, since that's how we have it in prod.

zzzeid (changes requested):

src/lando/utils/management/commands/setup_dev.py
shtrom: The check on `scm_type` with the `continue` on lines 45--46 makes sure we only associate all repos _of the matching type_ to the worker.

src/lando/utils/management/commands/setup_dev.py
shtrom: This was to pacify the static analyser, as `workers` is initialised with `None` values at first https://mozilla.slack.com/archives/C4D3JFF26/p1748999453647419

But by the time we're here, it should no longer be the case...

I think I need to change the way `workers` is first declared (:

src/lando/utils/management/commands/setup_dev.py
```diff
@@ -42,10 +44,12 @@ def setup_workers(self):
         for repo in Repo.objects.all():
             # Associate all repos with applicable worker.
             for worker in workers.values():
-                if not worker or worker.scm != repo.scm_type:
+                if worker.scm != repo.scm_type:
                     continue
-                self.stdout.write(f"Adding {repo} ({repo.scm_type}) to {worker}.")
-                worker.applicable_repos.add(repo)
+                # if not worker.applicable_repos.filter(repo=repo):
+                if repo.worker_set.exists():
```
zzzeid, line 50: nit: we should probably put a nicer `related_name` in that model, so we can do something like `repo.workers`.

zzzeid (approved):

src/lando/utils/management/commands/setup_dev.py
```diff
@@ -14,33 +16,44 @@ class Command(BaseCommand):
...
-        }
+        scm_types = [
+            SCM_TYPE_GIT,
+            SCM_TYPE_HG,
+        ]
+
+        worker_types = [
+            "",  # landing workers
+            "-automation-worker",
+        ]
```
cgsheeh, line 27: Nit: I'm going to remove the hg automation worker shortly as it is unused. Perhaps we could have an explicit list of the expected workers, `hg`, `git` and `git-automation-worker`, to be more precise.

src/lando/utils/management/commands/setup_dev.py
```diff
@@ -14,33 +16,44 @@ class Command(BaseCommand):
...
         for repo in Repo.objects.all():
             # Associate all repos with applicable worker.
-            self.stdout.write(
-                f"Adding {repo} ({repo.scm_type}) to {workers[repo.scm_type]}."
-            )
-            workers[repo.scm_type].applicable_repos.add(repo)
+            for worker in workers.values():
+                if worker.scm != repo.scm_type:
+                    continue
+                # if not worker.applicable_repos.filter(repo=repo):
```
cgsheeh, line 49: Looks like a leftover comment from development.

cgsheeh (approved):

src/lando/utils/management/commands/setup_dev.py
shtrom: I'd rather not fiddle with this here, but I agree.

src/lando/utils/management/commands/setup_dev.py
shtrom: Ok, I added a more flexible mapping :+1:.

cgsheeh (approved):

PR #342, opened 2025-05-29
settings: bump DATA_UPLOAD_MAX_MEMORY_SIZE (bug 1964604)

shtrom:


src/lando/settings.py
```diff
@@ -48,6 +48,9 @@
 SESSION_COOKIE_SECURE = True
 CSRF_COOKIE_SECURE = True
 
+# Set the upload size to 10x the default value (2.5MiB); see bug 1964604.
+DATA_UPLOAD_MAX_MEMORY_SIZE = 26214400
```
zzzeid, line 52: It wouldn't hurt to treat this like other variables, i.e., have a default but try to fetch from the environment.

shtrom: Very good point!

PR #344, opened 2025-06-03
hgexports: use bytes when initializing email message (bug 1969532)

zzzeid:
- work around issue with `email.message.get_content` corrupting message with unexpected characters
- add headers to test patch to make sure the correct code paths are triggered in the test

### DIFF
--- a/src/lando/api/legacy/hgexports.py
+++ b/src/lando/api/legacy/hgexports.py
@@ -302,8 +302,9 @@ class GitPatchHelper(PatchHelper):
 
     def __init__(self, fileobj: io.StringIO):
         super().__init__(fileobj)
-        self.message = email.message_from_string(
-            self.patch.read(), policy=default_email_policy
+        message_bytes = fileobj.read().encode("utf-8")
+        self.message = email.message_from_bytes(
+            message_bytes, policy=default_email_policy
         )
         self.message.set_charset("utf-8")
         self.commit_message, self.diff = self.parse_email_body(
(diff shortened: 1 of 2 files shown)


shtrom (approved):

shtrom:
I did some testing in dev and haven't found any issue with
1. landing the the file with the problematic content using lando-cli https://github.com/mozilla-conduit/ff-test/blob/b62b992f50a9dd52c6e7593a76fff07054cb05ca/third_party/libwebrtc/moz-patch-stack/0091.patch#L5601
2. landing commits with UTF-8 changes https://phabricator-dev.allizom.org/D3970

This is good to merge and deploy :shipit:

cgsheeh (commented):
Can we add a test to this that shows it correctly parsing some currently-bad input?

Given the issues we've seen with treating patches as strings, I wonder if we should go back to treating them as `bytes` and use `git am` to apply them directly, even if just for the automation API use case.

zzzeid:
> Can we add a test to this that shows it correctly parsing some currently-bad input?
> 
> Given the issues we've seen with treating patches as strings, I wonder if we should go back to treating them as `bytes` and use `git am` to apply them directly, even if just for the automation API use case.

The issue is not the input itself, but the content transfer encoding header and how `email.message` is handling it. After verifying that this is indeed the root cause of this issue, I updated the patch itself that we have in the tests.

cgsheeh (approved):

zzzeid:
>Do we need a patch to lando_cli to change how we run git format-patch? If we can make things more consistent on both ends that would be desirable.

I think this requires more research, but it would definitely be good to make it more consistent with the patches we had previously (if that's even possible), since there may be some other unknown issues that may arise. It may mean that we have to do more manual parsing instead.

PR #365, opened 2025-06-16
repo: add a flag to enable/disable diff inspection hooks (Bug 1972404)

cgsheeh:
Add new `hooks_enabled` flag on the `Repo` object to control whether
diff-inspection checks should run for automation jobs against the repo.
Add a parameter to the `test_automation_job_create_commit_failed_check`
test to assert failing diffs will not cause landing failures due to the
hooks when the flag is disabled.


ahochheiden (approved):

zzzeid (commented):
lgtm!

PR #374, opened 2025-06-19
headless_api: add an action to apply patches as binary (Bug 1972956)

cgsheeh:
Add a new `add-commit-base64` action which accepts a base64
encoded patch in the format of the target repo to apply patches
to. This action will directly apply patches, using `git am` for
Git and `hg import` for Mercurial. Add an `apply_patch_bytes`
method on the `AbstractSCM` class and implementations for each
concrete SCM.


src/lando/main/tests/test_git.py
```diff
@@ -217,6 +218,53 @@ def test_GitSCM_apply_get_patch_merge(
...
+    commit = scm.describe_commit()
+
+    expected_patch = patch_str
+    new_patch = scm.get_patch(commit.hash)
+
+    assert new_patch, f"Empty patch unexpectedly generated for {commit.hash}"
+
+    # The git version stamp varies. Strip it from the output before comparing.
+    remove_git_version_re = r"\d+(\.\d+)+$"
+    no_version_patch = re.sub(remove_git_version_re, "", new_patch)
```
shtrom, line 240: I think we have to add a fixture for this next time round. I wrote the same logic in another test.

shtrom (approved):
LGTM!

src/lando/main/scm/hg.py
```diff
@@ -289,6 +289,14 @@ def apply_patch(
...
             )
 
+    def apply_patch_bytes(self, patch_bytes: bytes):
+        """Apply the given `hg export` to the repo directly."""
+        with tempfile.NamedTemporaryFile(mode="wb", suffix=".patch") as tmp_file:
+            tmp_file.write(patch_bytes)
+            tmp_file.flush()
+
+            self.run_hg(["import", tmp_file.name])
+
```
zzzeid, line 299: Wondering what the value is of continuing to add implementations for `hg` (particularly in the context of automation) given that there is no plan to use these, and we also decommissioned the unused worker? I also don't think this code path is tested in this PR.

This should probably just raise `NotImplementedError`.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -597,6 +598,54 @@ def test_automation_job_add_commit_success_git(
...
+    )
+    AutomationAction.objects.create(
+        job_id=job,
+        action_type="add-commit-base64",
+        data={
+            "action": "add-commit-base64",
+            "content": patch_b64,
+        },
+        order=0,
+    )
```
zzzeid, line 627: Ideally these would go into a fixture that can be customized as needed, instead of recreating jobs/actions for every test from scratch.

src/lando/main/tests/test_git.py
zzzeid: nit: Given that this is repeated twice in this PR, I think now would be a good time to fix this (here and test_GitSCM_apply_patch_bytes_base64)

zzzeid (changes requested):

src/lando/main/scm/git.py
```diff
@@ -210,6 +210,14 @@ def apply_patch(
 
                     raise exc
 
+    def apply_patch_bytes(self, patch_bytes: bytes):
+        """Apply the given `git format-patch` to the repo directly."""
+        with tempfile.NamedTemporaryFile(mode="wb", suffix=".patch") as tmp_file:
+            tmp_file.write(patch_bytes)
+            tmp_file.flush()
+
+            self._git_run("am", tmp_file.name, cwd=self.path)
```
shtrom, line 219: As you suggested, adding `--keep-cr` is needed.

```suggestion
            self._git_run("am", "--keep-cr", tmp_file.name, cwd=self.path)
```

shtrom (approved):
Good to go, with the `--keep-cr` suggestion applied.

src/lando/headless_api/tests/test_automation_api.py
cgsheeh: Agreed, there is a lot of repeated boilerplate here as well. This would be a larger change and out of scope for this PR, filed https://bugzilla.mozilla.org/show_bug.cgi?id=1973247 to track this.

zzzeid (approved):

PR #377, opened 2025-06-23
tests: add a fixture to create automation jobs (Bug 1973247)

cgsheeh:
Add a new `automation_job` fixture for creating automation
jobs and associated actions. This removes much of the boilerplate
around assigning the `job_id` and ordering the actions, as well
as specifying the `action_type` when it is already defined
in the `action` JSON definition.

### DIFF
--- a/src/lando/headless_api/tests/test_automation_api.py
+++ b/src/lando/headless_api/tests/test_automation_api.py
@@ -28,14 +28,35 @@ from lando.main.tests.test_git import _create_git_commit
 from lando.pushlog.models import Push
 
 
+@pytest.fixture
+def automation_job():
+    """Create an automation job from the specified actions."""
+
+    def create_automation_job(actions: list[dict], **job_args):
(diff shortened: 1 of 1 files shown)


src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -28,14 +28,35 @@
 from lando.pushlog.models import Push
 
 
+@pytest.fixture
+def automation_job():
+    """Create an automation job from the specified actions."""
+
+    def create_automation_job(actions: list[dict], **job_args):
```
zzzeid, line 35: why not add return type here also since there are other type hints?

zzzeid (approved):

src/lando/headless_api/tests/test_automation_api.py
cgsheeh: Good call, added.

PR #378, opened 2025-06-23
tests: add fixtures to mock out `phab_trigger_repo_update` for workers (Bug 1969192)

cgsheeh:
Add fixtures to mock out the `phab_trigger_repo_update` method on
the landing and automation workers. We currently add this code
snippet to many tests manually, so this fixture allows us to mock
out the call to create a new Celery job without polluting the test
with irrelevant mocks.


shtrom (approved):
Lovely.

PR #383, opened 2025-06-24
scm: remove `get_current_node` in favour of `head_ref` (Bug 1961081)

cgsheeh:
These methods perform the same function. As `head_ref` is
older, remove `get_current_node` and convert usages to
`head_ref`. The method was never added to `AbstractSCM`
so this change is only required in `HgSCM` and `GitSCM`.


shtrom (approved):
Yes... Hahaha... YES!

PR #411, opened 2025-07-03
requirements: upgrade various packages (bug 1976221)

zzzeid:
This PR will preclude various other open dependabot PRs (including some with dependency conflicts).


cgsheeh (approved):
LGTM. We should probably have a bug for this as it is essentially a code change.

shtrom (approved):

zzzeid:
> LGTM. We should probably have a bug for this as it is essentially a code change.

I think given that we use dependabot, and those do not have bugs associated with them, this may not require a bug unless perhaps an actual code change is required to accommodate an upgraded package. Sometimes dependabot doesn't get it right or there is a large accumulation of upgrades, so these PRs are needed. We could also have one bug that stays open but not sure what value that would add other than keeping track of all the PRs.

PR #417, opened 2025-07-10
headless_api: more granular read-only fields on AutomationJob *b

shtrom:


src/lando/headless_api/admin.py
```diff
@@ -24,18 +22,13 @@ class AutomationActionJobInline(ReadOnlyInline):
...
-            formfield.widget.can_change_related = False
-            formfield.widget.can_delete_related = False
-        return formfield
+    def has_add_permission(self, request, obj=None) -> bool:  # noqa: ANN001
+        """Forbid addition of any action object from the inline interface."""
+        return False
+
+    def has_delete_permission(self, request, obj=None) -> bool:  # noqa: ANN001
+        """Forbid deletion of any action object from the inline interface."""
+        return False
```
cgsheeh, line 31: Nit: I think the types here are `request: HttpRequest, obj: AutomationAction | None = None`, we could add those and remove the `noqa`.

shtrom: The type-hinting of `request` is tricky (because Django doesn't use type-hints internally). We had a chat with @zzzeid the other day, and concluded that we should use `WSGIRequest` (which is a subclass of `HttpRequest` https://github.com/django/django/blob/0fe218842e0e396e3ab3982bd21227968a9e7fd8/django/core/handlers/wsgi.py#L56), as we use `uwsgi`.

PR #420, opened 2025-07-11
utils: use raw query to pause workers for maintenance (bug 1945758)

shtrom:
This is necessary as a requirements of the fix for bug 1945758. This allows us to update the worker database schemas. Without this deployments will fail as the maintenance mode is switched on, and new columns are fetched from the yet unmigrated database, leading to failures.

### DIFF
--- a/src/lando/utils/management/commands/maintenance.py
+++ b/src/lando/utils/management/commands/maintenance.py
@@ -23,7 +23,7 @@ class Command(BaseCommand):
     def _pause_workers(self):
         """Pause all workers."""
         # We explicitely select the `is_paused` field, and defer the others.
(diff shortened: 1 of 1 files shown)


src/lando/utils/management/commands/maintenance.py
```diff
@@ -22,7 +22,11 @@ def add_arguments(self, parser):  # noqa: ANN001
 
     def _pause_workers(self):
         """Pause all workers."""
-        workers = Worker.objects.all()
+        # We explicitely select the `is_paused` field, and defer the others.
+        # This allows a new version of lando to Pause workers during an update, even
```
shtrom, line 26:
```suggestion
        # This allows a new version of lando to pause workers during an update, even
```

PR #438, opened 2025-07-22
repo: do not default to SSH for pull path (bug 1958211)

zzzeid:
- defaulting to SSH is no longer needed and can be manually set
- add default branch (main) if this is a git repo
- do some clean up and add comments


shtrom (approved):
lgtm

PR #440, opened 2025-07-23
testing: add shims to mock authentication locally (Bug 1978881)

cgsheeh:
Add a `MockAuthBackend` which overrides `authenticate` to always
return a mocked local user. Add a `MockAuthMiddleware` to log the
user in to a session.

With these shims loaded, running Lando locally with `docker compose`
will show an authenticated view.

### DIFF
--- a/src/lando/middleware.py
+++ b/src/lando/middleware.py
@@ -6,12 +6,18 @@ from io import StringIO
 from typing import Optional
 
 from django.conf import settings
+from django.contrib.auth import authenticate, get_user_model, login
+from django.contrib.auth.backends import BaseBackend
+from django.contrib.auth.models import Permission
+from django.contrib.contenttypes.models import ContentType
 from django.core.handlers.wsgi import WSGIRequest
 from django.http import HttpRequest, HttpResponse
 from django.template.loader import render_to_string
 from django.urls import resolve
 
 from lando.main.models import ConfigurationKey, ConfigurationVariable
+from lando.main.models.profile import Profile
+from lando.treestatus.models import Tree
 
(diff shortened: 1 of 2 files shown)


cgsheeh:
@zzzeid @shtrom posting this as a draft so we can discuss the best approach for using this in the repo. Test failures are expected since it's currently loading into the main `settings.py` file.

My initial thought is we should add this in a separate `settings.py` file, and have that file load via an environment variable or CLI argument when starting Lando. Any other ideas are appreciated.

zzzeid:
> @zzzeid @shtrom posting this as a draft so we can discuss the best approach for using this in the repo. Test failures are expected since it's currently loading into the main `settings.py` file.
> 
> My initial thought is we should add this in a separate `settings.py` file, and have that file load via an environment variable or CLI argument when starting Lando. Any other ideas are appreciated.

Initial thoughts:
- I am not sure we should unconditionally authenticate users, this would at least break invalid auth workflows (e.g., providing an incorrect user/password), and seems like way more than what is needed for local testing.
- I think rather than creating a separate auth backend, adding a script that creates test users locally along with their required permissions would be better and more streamlined. There already is an auth backend that allows logging in with a user/password, so this can be used instead, and we already create an admin user in this way.

Aside from the comments above:

> My initial thought is we should add this in a separate `settings.py` file, and have that file load via an environment variable or CLI argument when starting Lando. Any other ideas are appreciated.

See implementation of test_settings.py and remote_settings.py which could be used in a similar way for local_settings.py.

cgsheeh:
Nowadays we have `setup_dev.py`, which is the preferred location for adding test users/repos/configs/etc.

PR #458, opened 2025-08-08
compose: mount staticfiles in nginx proxy (bug 1981857)

shtrom:
Unlike the `lando` service, where we mount the whole codebase, nothing was mounted into the `proxy`.

Therefore, it could not find the staticfiles as directed by the `nginx` configuration.


cgsheeh (approved):
Confirmed this resolves the bug. Thank you for the quick fix!

PR #459, opened 2025-08-08
try_api: OAuth bearer token support (bug 1979246)

shtrom:


shtrom:
We can't merge this right now, as we're waiting for the upstream Django-OIDC fix to land, but this is otherwise good for a first look.

src/lando/try_api/api.py
```diff
@@ -0,0 +1,29 @@
...
+        # to pass it. But we need to inherit from HttpBearer for Auth to work.
+        oidc_auth = LandoOIDCAuthenticationBackend()
+        return oidc_auth.authenticate(request)
+
+
+api = NinjaAPI(urls_namespace="try", auth=GlobalAuth())
+
+
+@api.get("/__userinfo__")
+def userinfo(request: WSGIRequest) -> JsonResponse:
```
cgsheeh, line 26: A docstring describing what this is for would be great.

src/lando/try_api/api.py
```diff
@@ -0,0 +1,29 @@
...
+    if not settings.ENVIRONMENT.is_lower():
+        raise HttpError(404, "Not Found")
```
cgsheeh, line 28: This will need to be removed before landing/deployment to prod?

src/lando/try_api/api.py
```diff
@@ -0,0 +1,29 @@
...
+from ninja.errors import HttpError
+from ninja.security import HttpBearer
+
+from lando.main.auth import LandoOIDCAuthenticationBackend
+
+logger = logging.getLogger(__name__)
+
+
+class GlobalAuth(HttpBearer):
+    def authenticate(self, request: WSGIRequest, token: str) -> User:
```
cgsheeh, line 15: We should have docstrings for the class and `authenticate` methods.

We've talked about moving human users of the automation API to this authentication method in the future. I think this is fine to keep here for now, but we will end up moving it somewhere available to the `try_api` and `headless_api` apps in the future if that happens.

cgsheeh (changes requested):

src/lando/try_api/admin.py
zzzeid, line 1: This file can be removed.

src/lando/try_api/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class TryConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "try"
```
zzzeid, line 6:
```suggestion
    name = "lando.try"
```

Conventionally this is also named after the module (`try_api` in this case). Any reason to deviate from that?

src/lando/try_api/models.py
zzzeid, line 1: This file can be removed.

src/lando/try_api/tests.py
zzzeid, line 1: This file should be removed. It can cause import problems. `try_api/tests/__init__.py` is missing and should be added.

zzzeid (changes requested):

zzzeid (approved):
lgtm, though I think the nit I left would make this a little better.

PR #473, opened 2025-08-13
try_api: implement /patches endpoint (bug 1979247)

shtrom:


src/lando/utils/management/commands/setup_dev.py
```diff
@@ -22,6 +22,8 @@ def setup_workers(self):
             },
             SCM_TYPE_HG: {
                 WorkerType.LANDING: "",
+                WorkerType.AUTOMATION: "-automation-worker",
+                WorkerType.TRY: "-try-worker",
```
shtrom, line 26: Remove

src/lando/main/models/worker.py
```diff
@@ -18,6 +18,7 @@ class WorkerType(models.TextChoices):
 
     LANDING = "LANDING", gettext_lazy("Landing worker")
     AUTOMATION = "AUTOMATION", gettext_lazy("Automation worker")
+    TRY = "TRY", gettext_lazy("Try worker")
```
shtrom, line 21: Remove.

src/lando/try_api/management/commands/start_try_worker.py
shtrom, line 1: Remove.

src/lando/try_api/worker.py
shtrom, line 1: Remove.

src/lando/try_api/api.py
```diff
@@ -1,9 +1,225 @@
...
+                detail=f"Could not determine the equivalent base commit for {target_commit_hash} in {repo.scm_type} for {mapping_repo}. Please try again later.",
+                status=status,
+            )
+
+    try_job = LandingJob.objects.create(
+        target_repo=repo,
+        requester_email=request.auth.email,
+        target_commit_hash=target_commit_hash,
+        status=JobStatus.SUBMITTED,
+        priority=-10,
```
cgsheeh: Why is the priority changed from the default here?

src/lando/utils/auth.py
```diff
@@ -60,6 +63,24 @@ def authenticate(self, request: WSGIRequest, token: str) -> User:
...
+    required_permission: str
+
+    def __init__(self, required_permission: str, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.required_permission = required_permission
+
+    @override
+    def authenticate(self, request: WSGIRequest, token: str) -> User:
+        user = super().authenticate(request, token)
+        # XXX Check the user's own permissions, not delegated from other roles.
```
cgsheeh: What does this comment mean? We don't currently leverage Django permission groups, but if we did start using them this would still be okay, no?

cgsheeh (changes requested):

src/lando/try_api/api.py
```diff
@@ -1,9 +1,225 @@
...
+        try:
+            if repo.scm_type == SCMType.HG:
+                target_commit_hash = CommitMap.git2hg(mapping_repo, target_commit_hash)
+            else:
+                target_commit_hash = CommitMap.hg2git(mapping_repo, target_commit_hash)
+        except CommitMap.DoesNotExist:
+            status = 400
+            error = (
+                f"Could not determine the equivalent base commit for {target_commit_hash} in {repo.scm_type} for {mapping_repo}. Please try again later.",
+            )
```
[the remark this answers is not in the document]
shtrom: 🤦🏼

src/lando/try_api/api.py
shtrom: Mostly because it doesn't hurt: It's not making assumptions on the worker setup, so this guarantees that they are, indeed, lower priority.

src/lando/main/scm/consts.py
```diff
@@ -18,3 +18,11 @@ class MergeStrategy(str, enum.Enum):
 
     # Use the target branch's tree, ignoring the current tree.
     THEIRS = "theirs"
+
+
+@enum.unique
```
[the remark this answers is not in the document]
shtrom: Ha! This works.

We can replace all the `SCM_TYPE_xx` with `SCMType.xx`, too.

The only issue is that this couples the django `models.TextChoices` into the otherwise decoupled `scm` module. But decoupling it requires extra handling and/or additional dependencies (see https://github.com/django/new-features/issues/94), which doesn't seem worth it.

src/lando/try_api/api.py
```diff
@@ -1,9 +1,225 @@
...
+    try_job.add_revisions(revisions)
+    try_job.sort_revisions(revisions)
+
+    return 201, JobResponse(
+        id=try_job.id,
+    )
+
+
+@api.get("/api/jobs/{int:try_job_id}/", url_name="api-job")
+def get_job_json(request: WSGIRequest, try_job_id: int) -> JsonResponse:
```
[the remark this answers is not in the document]
shtrom: This was supposed to be a backward compat with the old lando API, e.g. https://api.lando.services.mozilla.com/landing_jobs/171365, but I already landed a separate implementation of it.

So no longer needed :+1:

src/lando/try_api/tests/test_api.py
```diff
@@ -0,0 +1,203 @@
...
+from lando.main.models.jobs import JobStatus
+from lando.main.models.landing_job import LandingJob
+from lando.main.models.repo import Repo
+from lando.try_api.api import api, legacy_api
+
+
+@patch("lando.try_api.api.AccessTokenAuth.authenticate")
+def test_legacy_try_patches_invalid_user(
+    mock_authenticate: Mock,
+    ninja_api_client: Callable,
```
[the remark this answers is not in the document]
shtrom: It was introduced in #767. I just tried using the Django test `Client` instead, which kinda works, but it seems not to go through the authentication middleware (which gets patched in with `@patch("lando.utils.auth.AccessTokenAuth.authenticate"`) that are set for Ninja with `auth=...`, so some of the authentication-specific tests fail.

src/lando/try_api/api.py
```diff
@@ -1,9 +1,225 @@
...
+        target_commit_hash=target_commit_hash,
+        status=JobStatus.SUBMITTED,
+        priority=-10,
+    )
+
+    # Create Revision objects from patches and associate them with the job
+    revisions = []
+    patch_helper_class = PATCH_HELPER_MAPPING[patches.patch_format]
+
+    for patch_data in patches.patches:
```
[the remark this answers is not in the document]
shtrom: Added error handling, RFC-7807 errors, and tests 👌🏼

shtrom:
@cgsheeh good for another look. I'm purposefully leaving the conflicts with `main` to fix for last.

Reviewing the changes and code coverage, I now realise that I have introduce some logic to use the `ProblemException` logic in some error handling paths, while we already had some `return status, ProblemDetail( ... )` in others.

Which one do you think we should prefer?

src/lando/main/models/repo.py
```diff
@@ -203,6 +201,9 @@ def is_try(self) -> bool:
     # Use this field to enable/disable access to this repo via the automation API.
     automation_enabled = models.BooleanField(default=False)
 
+    # Use this field to enable/disable access to this repo via the try API.
+    try_enabled = models.BooleanField(default=False)
```
zzzeid: nit: maybe `is_try_repo` is a better descriptor here, since a repo is either a try repo, or it isn't, and this fact is immutable.

src/lando/main/scm/consts.py
```diff
@@ -18,3 +17,15 @@ class MergeStrategy(str, enum.Enum):
 
     # Use the target branch's tree, ignoring the current tree.
     THEIRS = "theirs"
+
+
+# @enum.unique
+# class SCMType(str, enum.Enum):
```
zzzeid: This should be removed.

src/lando/try_api/api.py
```diff
@@ -1,9 +1,243 @@
...
+        "responses": {
+            200: {
+                # XXX: This should not happen, but NinjaAPI does't let me disable this.
+                "description": "Not used.",
+                "content": {},
+            },
+            201: {"description": "Push was submitted successfully."},
+            400: {
+                "description": "Invalid request.",
+                # "content": {"application/problem+json": {"schema": ProblemDetail}},
```
zzzeid: is this meant to be commented out?

src/lando/try_api/api.py
```diff
@@ -1,9 +1,243 @@
...
+        int, Field(description="The ID of the job created for this submission.")
+    ]
+
+
+@api.post(
+    "/patches",
+    summary="Submit a set of patches to the Try server.",
+    url_name="api-patches",
+    response={201: JobResponse, 400: ProblemDetail, 404: ProblemDetail},
+    openapi_extra={
```
zzzeid: Is this needed? What does it provide that isn't already provided elsewhere?

src/lando/try_api/api.py
```diff
@@ -1,9 +1,243 @@
...
+        error = f"Repo {repo_name} is not a Try repository."
+        logger.info(
+            error,
+            extra={"user": request.user.email, "token": request.auth},
+        )
+        return status, ProblemDetail(
+            title="Not a Try repository", detail=error, status=status
+        )
+
+    # XXX: We'll need a more flexible way to set this.
```
zzzeid: Can we not use `Repo.legacy_source` to determine this?

src/lando/try_api/api.py
```diff
@@ -1,9 +1,243 @@
...
+    patch_helper_class = PATCH_HELPER_MAPPING[patches.patch_format]
+
+    for patch_no, patch_data in enumerate(patches.patches):
+        # Decode the base64 patch data to bytes
+        try:
+            decoded_patch_bytes = base64.b64decode(patch_data)
+        except Exception as exc:
+            raise BadRequestProblemException(
+                title="Invalid base64 patch data",
+                detail=f"Invalid base64 data for patch {patch_no}: {exc}",
```
zzzeid: are we exposing some internal errors here with `{exc}`?

zzzeid (changes requested):

shtrom (commented):
Forgot to submit previous comments 🤦🏼

src/lando/main/models/repo.py
shtrom: I just renamed it to `is_try`, and removed the `@property` that we previously had that was checking the repo name.

src/lando/utils/management/commands/create_environment_repos.py
```diff
@@ -135,14 +158,20 @@
         },
         # Use real `try` for testing since `try` is a testing environment anyway.
         {
-            "name": "try",
+            "name": "firefox-try",
```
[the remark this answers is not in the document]
shtrom: Ah, yes, that's a straggler. We decided to keep using `try` for as long as it was in HG.

src/lando/utils/auth.py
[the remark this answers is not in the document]
shtrom: I think the whole of this file is Ninja-specific for the moment. It does provide helpers and an API, so we could split the API to a separate submodule. Is this what you mean?

src/lando/try_api/api.py
shtrom: Maybe? What would we have to set the legacy source to on the Try repo?

Looking at the `firefox-autoland` repo, it has `autoland` as its legacy source.

I'm not sure the `CommitMaps` are tied to this, and the mappings are hardcoded there, too https://github.com/mozilla-conduit/lando/blob/27315c19ec74ec10eb87cb12ad1fd72622367444/src/lando/main/models/commit_map.py#L22

Here, we want to make to the matching Git repo for the Try repo, rather than Hg repo, so it feels like it would be the other way round?

Perhaps we need another attribute in the Repo model that tells us which main repo it supports. So `try` would have the attribute set to `firefox`, which would then allow use to look up mapping from `mozilla-unified`?

src/lando/try_api/api.py
shtrom: Unlikely, but we could tighten the Exception. My only concern here would be malformed base64, which would throw a `binascii.Error`.

src/lando/try_api/api.py
```diff
@@ -1,9 +1,243 @@
...
+
+            # Extract patch information using PatchHelper
+            author_name, author_email = patch_helper.parse_author_information()
+            commit_message = patch_helper.get_commit_description()
+            timestamp = patch_helper.get_timestamp()
+            diff = patch_helper.get_diff()
+        except ValueError as exc:
+            raise BadRequestProblemException(
+                title="Invalid patch data",
+                detail=f"Invalid patch data for patch {patch_no}: {exc}",
```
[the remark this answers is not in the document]
shtrom: I think this one is OK, as we are only catching ValueErrors than may get thrown from parsing the provided patch in the patch_helper.

shtrom:
Now based on #477 as it relies on the permission checking logic.

src/lando/main/auth.py
```diff
@@ -115,9 +115,13 @@ def __call__(self, f: Callable) -> Callable:
         @functools.wraps(f)
         def wrapper(request, *args, **kwargs):  # noqa: ANN001
             if not request.user.has_perm(f"main.{self.required_permission}"):
-                raise PermissionDenied()
+                raise PermissionDenied(
+                    f"Permission level {self.required_permission} is necessary"
+                )
             return f(request, *args, **kwargs)
 
+        return wrapper
```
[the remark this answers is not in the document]
zzzeid: Probably ok to keep the fix here, but maybe ensure there's a reference to it in the commit message.

src/lando/main/models/profile.py
```diff
@@ -143,3 +143,31 @@ def update_permissions(self):
...
+            # permissions for superusers. Here, we want to check permissions that have been
+            # explicitely given to the user from LDAP groups.
+            if self.user.user_permissions.filter(codename=permission):
+                return True
+        else:
+            # If the user is not a superuser, we can skip the DB round-trip.
+            if f"main.{permission}" in self.user.get_user_permissions():
+                return True
+
+        return False
```
zzzeid: As mentioned elsewhere, we should separate this change from this PR, and apply this to anything else that uses permissions. I also think this should be specific to repo permissions, i.e., maybe rename it to `has_repo_permission` and limit permission checks to SCM_* ones.

src/lando/ui/tests/test_ui.py
zzzeid: Ideally the `SCMType` changes would have gone into a separate PR, to make it simpler to review this one.

src/lando/try_api/api.py
zzzeid: I think we should keep the priority set to default, for the time being. Things are already prioritized in a FIFO way, setting priority thus far has been mostly useful to re-prioritize something to go to the top of the queue (or bottom) on a case by case basis.

src/lando/try_api/api.py
zzzeid: I'm OK with setting a try repo reference here, but also OK with a direct mapping, but we should probably move the mapping to `models.Repo`, and include the source repo and the target try repo (i.e., `{"firefox": "try"}`).

src/lando/try_api/api.py
```diff
@@ -1,9 +1,247 @@
...
+        logger.info(
+            error,
+        )
+        return status, ProblemDetail(
+            title="Not a Try repository", detail=error, status=status
+        )
+
+    if not request.user.profile.has_direct_permission(
+        repo.required_permission.removeprefix("main.")
+    ):
```
zzzeid: I don't think this is very clean here. We should just have something like checks repo permissions without altering the actual permission string when called. E.g., `request.user.profile.has_repo_permission(repo.required_permission)`.

src/lando/api/tests/test_transplants.py
```diff
@@ -1743,6 +1767,65 @@ def test_revision_has_data_classification_tag(
...
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "superuser,user_perms,group_perms,should_allow",
+    (
+        (False, [], [], False),
+        (False, [], ["scm_level_3"], False),
+        (False, ["scm_level_3"], [], True),
+        (True, [], [], False),
+        (True, [], ["scm_level_3"], False),
+        (True, ["scm_level_3"], [], True),
```
zzzeid: It's going to get messy if we start mixing permissions with and without the app name. We shouldn't assume permissions belong to `main` but rather explicitly state those where they are used.

zzzeid (changes requested):
Few more changes requested. Overall looking good, main comments are around permission checking.

src/lando/ui/tests/test_ui.py
shtrom: Yeah... It's another one of those changes that get too tangled for their own good.

src/lando/main/models/profile.py
shtrom: Ok, I fixed it in the #476, and merged it back here. I kept the `has_direct_perm` approach, as it's generic enough, but I made it take full perm strings with the app_label, to be consistent with the rest of Django.

shtrom:
@zzzeid all comments addressed. The permission check changes are in https://github.com/mozilla-conduit/lando/pull/476, which can be reviewed/landed independently.

I'm not sure why this PR here shows conflicts at the moment, as I think it's up-to-date with `main`. In any case, it'll likely need a rebasing before landing, as I suspect we'll get conflicts when applying the patches in sequence.

src/lando/utils/ninja_auth.py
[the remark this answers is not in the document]
shtrom: Lando uses `LandoOIDCAuthenticationBackend`, which `AccessTokenLandoOIDCAuthenticationBackend` subclasses. It's not used anywhere else in Lando, but it is indeed not Ninja-specific. I moved it to `main.auth`

src/lando/try_api/api.py
```diff
@@ -1,9 +1,252 @@
...
+    ] = "try"
+
+    base_commit: Annotated[
+        str,
+        Field(
+            description="The published base commit on which to apply `patches`",
+        ),
+        StringConstraints(min_length=40, max_length=40, pattern=r"^[0-9a-f]{40}$"),
+    ]
+    base_commit_vcs: Annotated[
```
zzzeid: This should be validated too, to be one of hg, git.

zzzeid: nit: vcs -> scm, here and elsewhere, for consistency.

src/lando/try_api/api.py
```diff
@@ -1,9 +1,252 @@
...
+
+    if not repo.user_can_push(request.user):
+        raise PermissionDenied(f"Missing permissions: {repo.required_permission}")
+
+    target_commit_hash = patches_request.base_commit
+    if patches_request.base_commit_vcs != repo.scm_type:
+        mapping_repo = CommitMap.TRY_REPO_MAPPING.get(repo.name)
+        if not mapping_repo:
+            status = 400
+            error = f"Don't know how to lookup commits from {patches_request.base_commit_vcs} to {repo.scm_type} for {repo_name}."
```
zzzeid:
```suggestion
            error = f"Unable to lookup commits from {patches_request.base_commit_vcs} to {repo.scm_type}. {repo_name} is not supported."
```

src/lando/try_api/api.py
```diff
@@ -1,9 +1,252 @@
...
+            else:
+                target_commit_hash = CommitMap.hg2git(mapping_repo, target_commit_hash)
+        except CommitMap.DoesNotExist:
+            status = 400
+            error = f"Could not determine the equivalent base commit for {target_commit_hash} in {repo.scm_type} for {mapping_repo}. Please try again later."
+            logger.warning(
+                error,
+            )
+            return status, ProblemDetail(
+                title="Error converting VCS commit IDs",
```
zzzeid: Should try and stick to using SCM everywhere for consistency (instead of VCS).

src/lando/try_api/api.py
```diff
@@ -1,9 +1,252 @@
...
+        # Create PatchHelper instance to parse the patch
+        patch_io = io.BytesIO(decoded_patch_bytes)
+        try:
+            patch_helper = patch_helper_class.from_bytes_io(patch_io)
+
+            # Extract patch information using PatchHelper
+            author_name, author_email = patch_helper.parse_author_information()
+            commit_message = patch_helper.get_commit_description()
+            timestamp = patch_helper.get_timestamp()
+            diff = patch_helper.get_diff()
```
zzzeid: Too many things in try block. The try block should contain only the line which may cause the exceptions that are being caught.

src/lando/main/auth.py
```diff
@@ -63,6 +66,32 @@ def update_user(self, user: User, claims: dict) -> User:
...
+                # aren't alrealdy registered.
+                try:
+                    return self.get_or_create_user(token, None, None)
+                except HTTPError as exc:
+                    if exc.response.status_code in [401, 403]:
+                        logger.warning(
+                            "failed to authenticate user from bearer token: %s", exc
+                        )
+                        return None
+                    raise exc
```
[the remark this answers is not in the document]
shtrom: As this is a shim, I'd rather not change the behaviour, which is consistent with the OIDC lib https://github.com/mozilla/mozilla-django-oidc/pull/551/changes#diff-957c3e74ed7279f104c8ae561b6207158af08c8a3bd7c2fb0a39ba55235379eaR301-R302.

The handling of the exception happens correctly further up in the middlewares.

src/lando/utils/management/commands/generate_requirements.py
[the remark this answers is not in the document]
shtrom: I think the file was DOS-line-ended.

It came from a merge of main into this branch.

src/lando/try_api/tests/test_api.py
```diff
@@ -0,0 +1,312 @@
...
+    user = scm_user([Permission.objects.get(codename="scm_level_1")], "password")
+    mock_authenticate = mock_authenticate_builder(user)
+
+    request_payload = {
+        "repo": "mozilla-central",  # from the mocked_repo_config
+        "base_commit": "0" * 40,
+        "base_commit_vcs": "git",
+        "patches": [
+            "YmFzZTY0Cg==",  # "base64"
+            "YmFzZTY0LXRvbwo=",  # "base64-too"
```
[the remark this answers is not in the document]
shtrom: That's what the comments are!

But yes (thinking back to the `xv` thing)

src/lando/main/auth.py
```diff
@@ -63,6 +66,32 @@ def update_user(self, user: User, claims: dict) -> User:
...
+
+        return super().authenticate(request, **kwargs)
```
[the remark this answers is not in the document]
zzzeid: >This is a fallback to the normal mechanism in case no HTTP_AUTHORIZATION header is present, so we can continue authenticating users with the normal flow if no token is present.

So I think the perspective I'm taking here is that by definition, `AccessTokenLandoOIDCAuthenticationBackend` is a separate backend (no matter what we want to call it, e.g., a shim or otherwise). Its name implies that it is for access token authentication. There's already another backend that handles the regular auth flow (i.e.,`LandoOIDCAuthenticationBackend`) and it's not necessary to duplicate its functionality in the access token flow.

Django already [falls back](https://docs.djangoproject.com/en/6.0/topics/auth/customizing/#specifying-authentication-backends) to the next backend in the list, if the current one fails.

This backend should only handle auth tokens, and fail if it can't.

src/lando/main/auth.py
```diff
@@ -63,6 +66,32 @@ def update_user(self, user: User, claims: dict) -> User:
         return super().update_user(user, claims)
 
 
+class AccessTokenLandoOIDCAuthenticationBackend(LandoOIDCAuthenticationBackend):
+    """Authenticates a user based on a Bearer access_token or the OIDC code flow.
+
+    This is a shim of the LandoOIDCAuthenticationBackend, borrowing code from mozilla-django-oidc#551.
+
+    https://github.com/mozilla/mozilla-django-oidc/pull/551"""
```
zzzeid: I know these references were copied over, but I don't think we should keep them here. This PR basically supersedes the one that is referenced. If anything, it may make more sense to reference this PR in the other one.

src/lando/main/auth.py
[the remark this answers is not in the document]
shtrom: Fair enough. I'm also realising we are already returning `None` in the path above anyway, so my previous point about interfaces is moot.

src/lando/utils/tests/test_ninja_auth.py
```diff
@@ -27,14 +27,67 @@ def test_authentication_valid_token(
...
+        "sub": "ad|Mozilla-LDAP|api-user",
+        "updated_at": "2026-01-23T00:55:19.962Z",
+    }
+
+    user_count = len(User.objects.all())
+
+    response = client.get(
+        "/auth/__userinfo__",
+        # The value of the token doesn't actually matter, as the output is controlled by
+        # the authenticator function, which we mock to return a User.
```
shtrom: Incorrect statement here.

```suggestion
```

src/lando/try_api/api.py
zzzeid: There's isn't a broad meta bug (yet). Probably wouldn't hurt to create one, but we could also use a keyword potentially.

shtrom:
@zzzeid I updated AccessTokenLandoOIDCAuthenticationBackend as per your suggestion, and I'm testing in develop. I'll merge this tomorrow.

shtrom:
rebased on main

PR #475, opened 2025-08-15
jobs: factorise common job logic into BaseJob (bug 1983208)

shtrom:
* Create a new `BaseJob` model and move all common attributes from LandingJob and AutomationJob to it, as well as queue-management logic.
* Also add a `to_dict` method for various JSON representation needs.
* Update `imports` throughout.


src/lando/headless_api/models/automation_job.py
```diff
@@ -72,74 +45,27 @@ def processing(self):
...
-            .filter(target_repo__in=repositories)
-            .order_by("-priority", "created_at")
-            .select_for_update()
-        )
+    def job_queue_query(
+        cls, repositories: Optional[Iterable[str]] = None, **kwargs
+    ) -> QuerySet:
+        """Return a query which selects the queued jobs."""
+        # XXX: If we can include IN_PROGRESS jobs, and sort by status_order, this method
+        # can be deleted in favour of the parent class's.
```
shtrom, line 62: @cgsheeh you may be the best placed to know:

I'm pretty sure https://github.com/mozilla-conduit/lando/pull/475/files#diff-857711dd1362a2dcfe7cc7b4a4ae6c412405c88e966aeb33600d517bcdd8354cR231-R238 would work equally well, so we can just remove the override.

src/lando/main/models/landing_job.py
```diff
@@ -281,31 +150,15 @@ def job_queue_query(
             grace_seconds (int): Ignore landing jobs that were submitted after this
                 many seconds ago.
         """
-        q = cls.objects.filter(status__in=JobStatus.pending())
-
-        if repositories:
-            q = q.filter(target_repo__in=repositories)
+        q = super().job_queue_query(repositories)
```
shtrom, line 153: This reorders how we build the QuerySet a bit. I'm assuming that the build order doesn't impact the final query when dealing with different parts (e.g., WHERE vs. ORDER BY).

src/lando/main/models/landing_job.py
```diff
@@ -247,31 +126,21 @@ def revisions_query(cls, revisions: Iterable[str]) -> QuerySet:
...
-                    f"{settings.PHABRICATOR_URL}/D{r.revision_id}" for r in j.revisions
-                ],
-                "status": j.status,
-                "updated_at": j.updated_at,
-            }
-            for j in jobs
-        ]
+    def to_dict(self) -> dict[str, Any]:
+        job_dict = super().to_dict()
+        job_dict["revisions"] = (
```
cgsheeh, line 131: Is this supposed to be a tuple?

src/lando/main/models/jobs.py
```diff
@@ -0,0 +1,277 @@
...
+            "created_at": self.created_at,
+            "error": self.error,
+            "id": self.id,
+            "requester": self.requester_email,
+            "status": self.status,
+            "updated_at": self.updated_at,
+        }
+
+        if self.target_repo:
+            job_dict["repository"] = (self.target_repo.short_name,)
```
cgsheeh, line 254: Is this supposed to be a tuple?

src/lando/headless_api/models/automation_job.py
cgsheeh: Yes, I think we can include in progress jobs and remove the override. 👍

src/lando/main/models/jobs.py
```diff
@@ -0,0 +1,277 @@
...
+
+from lando.main.models.base import BaseModel
+from lando.main.models.commit_map import CommitMap
+from lando.main.models.repo import Repo
+from lando.main.scm.consts import SCM_TYPE_HG
+
+logger = logging.getLogger(__name__)
+
+
+class JobStatus(models.TextChoices):
```
cgsheeh, line 20: Adding a docstring here would be nice.

src/lando/main/models/jobs.py
```diff
@@ -0,0 +1,277 @@
...
+        and then we order by creation time (older first).
+
+        Any `JobStatus.IN_PROGRESS` jobs are second. As there should
+        be a maximum of one (per repository), and with the assumption of a single worker
+        instance, a worker picking up an IN_PROGRESS job would mean that the job
+        previously crashed, and that the worker needs to restart processing.
+        """
+        return Case(
+            When(status=cls.SUBMITTED, then=1),
+            When(status=cls.IN_PROGRESS, then=2),
```
cgsheeh, line 42: I wonder if we should process `IN_PROGRESS` jobs first, since those would have been at the start of the queue when the worker crashed. Probably out of scope for this PR.

src/lando/main/models/landing_job.py
cgsheeh: I think assumption is correct.

cgsheeh (approved):
Looks great, thanks for cleaning this up!

Please see my comments about tuples in `to_dict`, but otherwise this LGTM.

src/lando/main/models/jobs.py
shtrom: Ditto 😅

src/lando/main/models/jobs.py
shtrom: I think the case to have IN_PROGESS jobs second is that if said job kills the worker badly, processing another SUBMITTED job next allows us to inch forwards, rather than be in a tight fail-loop that would block the processing of any message. Not great, but it turns a critical failure into a slow down.

There are two cases I can think of where we'd end up in this situation: either a poison pill job, in which case marking as FAILED would make sense, or the worker was killed due to other reasons (restarts, new deploy, cosmic rays, ...), in which case we may not want to set the job to FAILED (though it would be such a rare occurrence that, as long as there are notifications about the failure, we could probably take the hit without too much disruption).

In any case, I'd rather not change this as part of the refactor, so as to not muddle the waters, but I think it's worth considering our options.

src/lando/headless_api/models/automation_job.py
```diff
@@ -72,74 +45,27 @@ def processing(self):
...
-        if action in (JobAction.FAIL, JobAction.DEFER):
-            self.error = kwargs["message"]
-
-        if action == JobAction.LAND:
-            self.landed_commit_id = kwargs["commit_id"]
-
-        self.save()
+        job_dict = self.to_dict()
+
+        job_dict["job_id"] = job_dict.pop("id")
```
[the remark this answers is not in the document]
shtrom: Yes. I'll keep both in, so we're not in a lockstep to update and deploy the consumer.

src/lando/main/models/jobs.py
```diff
@@ -0,0 +1,277 @@
...
+        "duration_seconds",
+    )
+    list_filter = ("target_repo__name", "requester_email", "created_at")
+    readonly_fields = (
+        "attempts",
+        "duration_seconds",
+        "error",
+        "landed_commit_id",
+        "requester_email",
+    )
```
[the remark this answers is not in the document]
shtrom: Ah, yes. I deliberately put it there, as I feel like it would make more sense to manage Models and their admin page in the same spot. It does go against the current pattern, though. I'll move it back.

src/lando/headless_api/models/automation_job.py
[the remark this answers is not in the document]
shtrom: Here: https://github.com/mozilla-conduit/lando-cli/pull/12

zzzeid (approved):

PR #488, opened 2025-08-25
repo: add support for (git)attributes override (bug 1984942)

shtrom:


cgsheeh (dismissed):

zzzeid (approved):
Looks good, though may need some clarification around why we are checking the beginning of the string. Based on that, this may need to be changed.

src/lando/main/models/repo.py
```diff
@@ -111,6 +111,14 @@ def path(self) -> str:
     milestone_tracking_flag_template = models.CharField(blank=True, default="")
     product_details_url = models.CharField(blank=True, default="")
 
+    # By default, override any attribute forcing files to be displayed as diffs (rather
+    # than binaries).
+    attributes_override = models.TextField(
+        blank=True,
+        default="* !diff\n",
+        help_text="SCM-specific attribute override. For git, this is `.gitattributes` value.",
```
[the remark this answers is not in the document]
shtrom: It's not exactly that file, but an override for that file. I'll make reference to the `gitattributes(5)` manpage instead, so somebody chasing that up can get full context.

src/lando/main/scm/git.py
```diff
@@ -460,9 +460,11 @@ def clean_repo(
         # explicitly given an empty string.
         if attributes_override is not None:
             # $GIT_DIR/info/attributes has the highest precedence.
-            with open(f"{self._git_dir}/info/attributes", "w+") as fp:
-                if not "\n".join(fp.readlines()).startswith(attributes_override):
+            with open(f"{self._git_dir}/info/attributes", "a+") as fp:
+                fp.seek(0)
```
zzzeid, line 464: This line is probably redundant.

src/lando/main/scm/git.py
```diff
@@ -460,9 +460,11 @@ def clean_repo(
         # explicitly given an empty string.
         if attributes_override is not None:
             # $GIT_DIR/info/attributes has the highest precedence.
-            with open(f"{self._git_dir}/info/attributes", "w+") as fp:
-                if not "\n".join(fp.readlines()).startswith(attributes_override):
+            with open(f"{self._git_dir}/info/attributes", "a+") as fp:
```
zzzeid, line 463: Not sure this needs to be `a+`. Maybe this should be `r+`. Or this can be split into two `open` statements (`r` and `w`) and then you don't need to truncate or seek altogether.

zzzeid (approved):

src/lando/main/scm/git.py
shtrom: It's needed with the `a+` (:

PR #495, opened 2025-09-02
AutomationJob: add queue view (bug 1977975)

shtrom:


cgsheeh (approved):
LGTM. I think the number of automation jobs present at once is low enough that collecting the full queue and indexing in Python is fast enough.

PR #505, opened 2025-09-04
ui: show type attribute from BaseJob and subclasses (bug 1986843)

shtrom:


src/lando/main/models/jobs.py
```diff
@@ -89,6 +89,10 @@ class BaseJob(BaseModel):
     class Meta:
         abstract = True
 
+    # A human-friendly name of this type of job.
+    # To be overriden by subclasses.
```
cgsheeh, line 93:
```suggestion
    # To be overridden by subclasses.
```

PR #512, opened 2025-09-09
GitSCM: raise retriable SCMInternalServerError on git-level errors (bug 1986917)

shtrom:


cgsheeh (approved):

src/lando/conftest.py
```diff
@@ -1042,3 +1042,16 @@ def _new_tree(
...
+def active_mock() -> Callable:
+    def active_mock(obj: object, method: str) -> mock.MagicMock:
+        """Mock a method without preventing its side-effect from happening."""
+        original = getattr(obj, method)
+        mock_method = mock.MagicMock()
+        mock_method.side_effect = original
+        setattr(obj, method, mock_method)
+        return mock_method
+
+    return active_mock
```
zzzeid, line 1057: nit
```suggestion
    def _active_mock(obj: object, method: str) -> mock.MagicMock:
        """Mock a method without preventing its side-effect from happening."""
        original = getattr(obj, method)
        mock_method = mock.MagicMock()
        mock_method.side_effect = original
        setattr(obj, method, mock_method)
        return mock_method

    return _active_mock
```

zzzeid (approved):

PR #513, opened 2025-09-09
GitSCM: don't retag if the tag is already in the desired state (bug 1986917)

shtrom:


src/lando/main/scm/git.py
```diff
@@ -666,14 +667,25 @@ def tag(self, name: str, target: str | None):
...
+
         tag_command = ["tag", name]
 
         if target:
             tag_command.append(target)
 
         self._git_run(*tag_command, cwd=self.path)
-
-    @override
-    def push_tag(self, tag: str, remote: str):
```
cgsheeh, line 677: Good catch. :)

zzzeid (approved):

PR #515, opened 2025-09-10
workers: move shared logic to BaseWorker (bug 1958894)

shtrom:
* explicit base properties
* abstract `type` property
* add abstract `job_type` property
* factor `__init__()` in
* factor `loop()` in
* make `run_job()` part of the interface, and rename
  `run_automation_job()`
* add `call_task()` for generic async calls

### DIFF
--- a/src/lando/api/legacy/workers/base.py
+++ b/src/lando/api/legacy/workers/base.py
@@ -6,20 +6,59 @@ import logging
 import os
 import re
 import subprocess
+from abc import ABC, abstractmethod
 from time import sleep
 
+from celery import Task
+from django.db import transaction
+from kombu.exceptions import OperationalError
+
 import lando.utils.treestatus
+from lando.api.legacy.treestatus import TreeStatus
 from lando.main.models import Worker as WorkerModel
+from lando.main.models.jobs import BaseJob, JobStatus
+from lando.main.models.repo import Repo
+from lando.main.models.worker import WorkerType
 
 logger = logging.getLogger(__name__)
 
 
-class Worker:
+class Worker(ABC):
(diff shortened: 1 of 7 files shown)


src/lando/api/legacy/workers/base.py
```diff
@@ -6,21 +6,62 @@
...
+        This should be overridden by concrete implementations."""
+        raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def type(self) -> WorkerType:
+        """Type of this Worker.
+
+        This should be overridden by concrete implementations."""
+        raise NotImplementedError()
```
zzzeid, line 48: I would get rid of these methods in favour of setting the variables on the subclasses themslves. Also, should avoid using built-in function names as method names.

```suggestion
    job_type: LandingJob | AutomationJob
    worker_type: LandingWorker | AutomationWorker
```

Though for both of these variables, I would suggest automatically detecting that based on a class name mapping, since that information is available and doesn't need to be explicitly set.

src/lando/api/legacy/workers/base.py
```diff
@@ -162,6 +232,19 @@ def start(self, max_loops: int | None = None):
...
+        Args:
+            task: celery.Task to call
+            *args: list of argurents for the Task
+        """
+        try:
+            # Send a task to Celery.
+            task.apply_async(args=args)
+        except OperationalError as e:
+            # Log the exception but continue gracefully.
+            # The repo will eventually update.
```
zzzeid, line 248: I think this comment may be specific to `phab_trigger_repo_update`.

cgsheeh (approved):

PR #516, opened 2025-09-10
headless_api: add merge-remote action (bug 1980701)

jcristau:
This adds a merge-remote action to the headless API, that takes a git repository URL and commit hash to merge into the target repo.  The intended use case is importing code and its history from an existing repository, such as when https://github.com/mozilla/application-services will be merged into https://github.com/mozilla-firefox/firefox.


shtrom:
One thing that we should be careful with merges of native git branches is that cinnabar doesn't support signed commits. If we allow any signed commit to land into a repo synced to HgMO with git-hg-sync, this will immediately create a divergence which will be painful to resolve. 

I think it might be worth adding a dedicated landing check that would ensure that none of the commits are signed before pushing. -> https://bugzilla.mozilla.org/show_bug.cgi?id=1996492

shtrom:
Unfortunately, it seems all commits in application-services are signed, so we'll have to make sure to strip the signatures.

image

cgsheeh (changes requested):
Looking very good, I think once we have some more test coverage and docs for the `ALLOWED_MERGE_REMOTE_REPOS` setting this will be good to go. 👍

jcristau:
Thanks for the reviews.  Out of curiosity, is there a reference for the issue with cinnabar and signed commits?

PR #535, opened 2025-09-18
uplift: query Lando `Repo` objects instead of Phab repo endpoint (Bug 1989305)

cgsheeh:
Switch the `UpliftRequestForm` to query Lando's local `Repo`s
with the `approval_required=True` flag to gather the list of
valid uplift repositories. This removes an API call to Phabricator
to gather information which Lando already contains locally.

### DIFF
--- a/src/lando/ui/legacy/forms.py
+++ b/src/lando/ui/legacy/forms.py
@@ -1,7 +1,6 @@
 from django import forms
 from django.forms.widgets import RadioSelect
 
-from lando.api.legacy.uplift import get_uplift_repositories
 from lando.main.models import Repo
 from lando.main.models.uplift import (
     UpliftAssessment,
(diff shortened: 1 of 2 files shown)


src/lando/ui/legacy/forms.py
```diff
@@ -85,8 +84,10 @@ class UpliftRequestForm(UpliftAssessmentForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        uplift_repos = get_uplift_repositories()
-        self.fields["repository"].choices = [(repo, repo) for repo in uplift_repos]
+        uplift_repos = Repo.objects.filter(approval_required=True).all()
+        self.fields["repository"].choices = [
+            (repo.name, repo.name) for repo in uplift_repos
```
zzzeid, line 89: nit: Maybe we should have a better display name, since we could probably parse the name and branch now from the repo details.

zzzeid (approved):

src/lando/ui/legacy/forms.py
cgsheeh: Agreed, it would be nice to have a readable name field on the `Repo` class. For now I think staying with the same format is acceptable.

PR #536, opened 2025-09-19
github: add GitHubAPI, GitHubAPIClient classes (bug 1989637)

zzzeid:
WIP DO NOT MERGE

- add GitHubAPI class
- add GitHubAPIClient class
- move "get token" functionality to GitHubAPI class

### DIFF
--- a/src/lando/main/models/repo.py
+++ b/src/lando/main/models/repo.py
@@ -279,6 +279,11 @@ class Repo(BaseModel):
             return self.url.removesuffix(".git")
 
     @property
+    def _github_repo_org(self) -> str | None:
+        if self.is_github:
+            return self._github_repo_url.split("/")[-2]
+
+    @property
     def git_repo_name(self) -> str:
         """Provide the bare name of the Git repo."""
         if self.scm_type != SCM_TYPE_GIT:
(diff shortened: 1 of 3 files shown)


src/lando/utils/github.py
```diff
@@ -0,0 +1,83 @@
...
+
+from lando.main.models.repo import Repo
+
+logger = logging.getLogger(__name__)
+
+
+class GitHubAPI:
+    GITHUB_BASE_URL = "https://api.github.com"
+
+    def __init__(self, repo: Repo):
```
shtrom, line 16: Using `Repo` here is introducing a strong coupling that we can probably avoid.

We could either extract the owner and name in the runner, and pass those two arguments to the constructor, or maybe just allow the `__init__` here to just get an unparsed `Repo.url`, and let it extract the components from the URL.

This way we'd have all the GitHub logic nicely tucked in one place.

src/lando/main/models/repo.py
```diff
@@ -275,6 +275,11 @@ def _github_repo_url(self) -> str | None:
         if self.is_github:
             return self.url.removesuffix(".git")
 
+    @property
+    def _github_repo_org(self) -> str | None:
```
shtrom: This would no longer be needed with #604.

PR #546, opened 2025-09-23
ui: add BaseJobView (bug 1990139)

shtrom:


src/lando/api/legacy/api/landing_jobs.py
```diff
@@ -53,22 +53,22 @@ def put(request: WSGIRequest, landing_job_id: int) -> JsonResponse:
         }
         return JsonResponse(data, status=400)
 
-    landing_job_id = form.cleaned_data["landing_job_id"]
+    job_id = form.cleaned_data["landing_job_id"]
```
shtrom, line 56: Hum, ok, maybe we could have done the renaming, too 🤔

PR #547, opened 2025-09-23
urls: implement GET /landing_jobs/<id> (bug 1983218)

shtrom:


src/lando/api/tests/test_transplants.py
```diff
@@ -962,6 +962,25 @@ def test_integrated_transplant_records_approvers_peers_and_owners(
...
+    ), f"Invalid status code from GET /landing_jobs/{job.id}"
+    assert (
+        response.json["id"] == job.id
+    ), f"Invalid id in GET /landing_jobs/{job.id} response"
+    assert (
+        response.json["status"] == JobStatus.LANDED
+    ), f"Invalid status in GET /landing_jobs/{job.id} response"
+    assert (
+        response.json["commit_id"] == job.landed_commit_id
+    ), f"Invalid commit_id in GET /landing_jobs/{job.id} response"
```
zzzeid, line 982: I don't think we should be adding new views that utilize the legacy "proxy" client -- it only adds more tech debt.

src/lando/main/models/jobs.py
```diff
@@ -275,14 +275,20 @@ def to_dict(self) -> dict[str, Any]:
...
-            "error": self.error,
+            "error": None,
             "id": self.id,
             "requester": self.requester_email,
             "status": self.status,
             "updated_at": self.updated_at,
         }
 
+        if self.status in [JobStatus.FAILED, JobStatus.DEFERRED]:
+            job_dict["error"] = self.error
```
zzzeid, line 286: Why not add the `error` field anyway, to make the response body more consistent?

src/lando/main/models/jobs.py
```diff
@@ -275,14 +275,20 @@ def to_dict(self) -> dict[str, Any]:
...
+
         if self.target_repo:
             job_dict["repository"] = self.target_repo.short_name
 
+        if self.landed_commit_id:
+            job_dict["commit_id"] = self.landed_commit_id
```
zzzeid, line 292: Same here.

zzzeid (dismissed):

shtrom:
Hum, looks like the `CVB.put` is broken. Marking as draft, I'll revisit tomorrow.

src/lando/api/tests/test_transplants.py
```diff
@@ -962,6 +963,26 @@ def test_integrated_transplant_records_approvers_peers_and_owners(
     assert approved_by == [[101], [102]]
 
     assert hg_landing_worker.run_job(job)
+    assert job.landed_commit_id
+
+    # Fetch Job data.
+    response = authenticated_client.get(
```
shtrom: This tests real HTTP interactions, rather than calling internal methods, so we use the normal test client, rather than the `proxy_client`

zzzeid (approved):
lgtm!

PR #551, opened 2025-09-25
worker: refactor `apply_patch` with job status transitions to base (Bug 1990959)

cgsheeh:


shtrom (approved):
lgtm!

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -189,21 +189,20 @@ def apply_and_push(
         """
         self.update_repo(repo, job, scm, job.target_commit_hash)
 
+        def apply_patch(revision: Revision):
```
zzzeid: nit: this would stand out better if it's at the top of the definition, before `self.update_repo`.

PR #552, opened 2025-09-26
Repo/LandingWorker: add and honour selectable landing hooks (bug 1980351)

shtrom:
This PR adds an individually-configurable list of hooks to run for landings for each repo.

It also moves all the hooks logic from the `main.scm.helpers` into a dedicated `utils.landing_checks`, and introduces a `LandingChecks` util which provides services to both the LandingWorker and AutomationWorker.

~NOTE: As the hooks information information was not available for each repo previously, the default of everything will be set for all repos. Most notably, this means that the `BugReferencesCheck`, previously only used for Try, will be enabled for all repos. This may lead to some failed landings until the hook is manually disabled for the affected repos.~
 
`BugReferencesCheck` is now specifically excluded from the default checks, as it's the more common path.

### DIFF
--- a/src/lando/main/admin.py
+++ b/src/lando/main/admin.py
@@ -280,6 +280,13 @@ class RevisionAdmin(admin.ModelAdmin):
 
 class RepoAdmin(admin.ModelAdmin):
     class Media:
+        # By default, labels in full-width views are limited to a fixed width left
+        # column.
+        # When rendering multiple checkboxes, like we do for the hooks selection,
+        # however, the label for each is in the right column. By default, the
+        # fixed-width will still apply, forcing unecessary line breaks that end up using
+        # more vertical space.
+        # This override returns the width to auto in this situation.
         css = {
             "all": ("css/custom_admin.css",),
         }


src/lando/main/models/repo.py
```diff
@@ -46,6 +47,18 @@ class RepoError(Exception):
...
+# Dict of hook names to docstring, suitable as human-friendly choices.
+HOOKS_CHOICES = {chk.__name__: chk.__doc__ or chk.__name__ for chk in ALL_CHECKS}
+
+
+def get_default_hooks() -> list[str]:
+    """Returns a list of all known hook names, suitable as a default value.
+
+    Assuming a normal repository, we enable everything but the BugReferencesCheck, which
+    is only relevant for Try-type repos."""
+    return [hook for hook in HOOKS_CHOICES if hook != BugReferencesCheck]
```
cgsheeh: It looks like `hook` would be the string name of the hook, and thus `hook != BugReferencesCheck` will be `True` even when `hook == "BugReferencesCheck"`. I think we need to compare to `BugReferencesCheck.__name__`.

src/lando/utils/landing_checks.py
```diff
@@ -0,0 +1,499 @@
...
+    """Utility class to run landing checks (a.k.a. hooks) on a list of commits."""
+
+    requester_email: str
+
+    def __init__(self, requester_email: str):  # noqa: ANN001
+        self.requester_email = requester_email
+
+    def run(
+        self,
+        hooks: list[str] | list[type[PatchCheck] | type[PatchCollectionCheck]],
```
cgsheeh: We accept either a string hook name or the concrete object when running hooks here, but we only ever call this method with the concrete objects in a test. I think we should remove the ability to pass `list[type[PatchCheck] | type[PatchCollectionCheck]]` and only accept `list[str]` to keep things simple.

src/lando/main/admin.py
```diff
@@ -2,6 +2,10 @@
 from typing import Callable, Self
 
 from django.contrib import admin
+from django.db.models import Field as DbField
+from django.db.models import Field as FormField
```
cgsheeh: Both of these imports are aliasing `django.db.models.Field`, seems the latter should be aliasing `django.forms.Field` instead.

cgsheeh (changes requested):

src/lando/main/models/repo.py
```diff
@@ -154,6 +167,20 @@ def path(self) -> str:
     # Use this field to enable/disable pre-landing hooks for a repo.
     hooks_enabled = models.BooleanField(default=True)
 
+    # Bring the HOOKS_CHOICES into this namespace for external reference.
+    HOOKS_CHOICES = HOOKS_CHOICES
+
+    # WARNING: Due to this binding, whenever checks are added or removed, a database
+    # migration is needed.
```
zzzeid: I think if HOOKS_CHOICES is not linked to docstrings, we don't really need this comment as there's no "magic" or anything potentially unexpected that will happen here.

src/lando/main/models/repo.py
```diff
@@ -154,6 +167,20 @@ def path(self) -> str:
     # Use this field to enable/disable pre-landing hooks for a repo.
     hooks_enabled = models.BooleanField(default=True)
 
+    # Bring the HOOKS_CHOICES into this namespace for external reference.
+    HOOKS_CHOICES = HOOKS_CHOICES
+
+    # WARNING: Due to this binding, whenever checks are added or removed, a database
```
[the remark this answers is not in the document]
zzzeid: We could use `lando makemigrations --check` to do this, either as a test, or in CI as a command (in case we don't want to impose this test during development, when schema migrations may not be finalized).

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -124,9 +101,17 @@ def run_job(self, job: AutomationJob) -> bool:
...
+                try:
+                    check_errors = landing_checks.run(new_commits)
+                except Exception as exc:
+                    message = "Unexpected error while performing landing checks."
+                    logger.exception(message)
+                    job.transition_status(
+                        JobAction.FAIL,
+                        message=f"{message}\n{exc}",
+                    )
+                    return True  # Do not try again, this is a permanent failure.
```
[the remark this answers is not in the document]
shtrom: Well, it's an unexpected error, so we don't know whether it's transient or not. If it is, a retry would see it through without user intervention.

OTOH, it if it not transient, this means it will fail forever... Yeah, ok, probably safer to keep it Permanent, unless we add a cap on the number of retries.

shtrom:
I cherry-picked 7d55a99d62bf7d3b289bed02b91ec4f485d45095 so I could update the migration that was causing an check failure without a merge of `main`.

src/lando/utils/cache.py
shtrom: Cherry-pick from 56a371718b40af36fec7f56463e0f40c2d36532f.

shtrom:
@cgsheeh @zzzeid ok, this one should be good to go. I merged `main` in, resolved conflicts, and re-tested. It all seems green to me.

One caveat that I discovered when testing, though, is that as the code to apply the hooks in is the workers, they need to refresh the repo from the database before they can honour a change in the configured hooks.

While this can go out as is, I'm wondering whether we should also add something where the workers refresh the repos from the DB before processing a job for a given repo.

What do you both think the best approach should be?

src/lando/main/models/repo.py
```diff
@@ -47,23 +44,55 @@ class RepoError(Exception):
...
+    ]
 
 
 class Repo(BaseModel):
     """Represents the configuration of a particular repo."""
 
     _scm: AbstractSCM | None = None
 
+    class HooksChoices(models.TextChoices):
+        """Allowable statuses of a tree."""
```
cgsheeh: Can this be moved to the top-level? Also, the docstring looks copied from the `TreeStatus` class.

cgsheeh (changes requested):
LGTM, but we should update the docstring on the `HooksChoices` class before landing.

src/lando/utils/landing_checks.py
```diff
@@ -84,6 +112,16 @@ def result(self) -> str | None:
 class TryTaskConfigCheck(PatchCheck):
     """Check for `try_task_config.json` introduced in the diff."""
 
+    @override
+    @classmethod
+    def name(cls) -> str:
+        return "TryTaskConfigCheck"
```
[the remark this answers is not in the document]
shtrom: It's not a requirement. This method was added so we can decouple the name used in the models from the class name. I just continued using the same string as it's still a fitting string, but it is now independent of what we name the class.

src/lando/main/tests/test_models.py
```diff
@@ -342,3 +345,35 @@ def test__models__CommitMap__fetch_push_data_invalid_response(commit_maps, monke
         "Number of hg changesets does not match number of git changesets: 2 vs 1",
     )
     assert CommitMap.objects.all().count() == previous_commit_map_count
+
+
+def test_hook_choices_all_checks():
+    """Ensure that all existing checks are available as configurable hooks."""
+    # If some checks should not be available as hook choices, list their name here.
```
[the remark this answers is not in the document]
shtrom: I figured it would be easier to already have the logic handy, so it's just a matter of adding to the list rather than thinking about how to achieve it.

cgsheeh (approved):
LGTM.

cgsheeh (approved):

PR #564, opened 2025-10-03
dockerfile: install `rust` toolchain before copying code (Bug 1992510)

cgsheeh:


Dockerfile
```diff
@@ -30,6 +21,15 @@ RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
 # See: rust-lang.org/tools/install (Configuring the PATH environment variable).
 ENV PATH="/root/.cargo/bin:${PATH}"
 
+# Copy code into the container.
+RUN mkdir /code
+COPY ./ /code
```
shtrom, line 26: I'd suggest moving this one lien (

shtrom, line 26: I'd suggest moving all this even further down, after line 31 (install pip and setuptools), as none of the lines before depend on the code.

shtrom (changes requested):
We can move `/code` even further down.

Dockerfile
```diff
@@ -30,6 +21,15 @@ RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
...
+
+RUN mkdir -p /code/.ruff_cache
+RUN chown -R app /code/.ruff_cache
+
+RUN pip install --upgrade pip setuptools
+
```
shtrom, line 32:
```suggestion
RUN mkdir -p /code/.ruff_cache
RUN chown -R app /code/.ruff_cache

RUN pip install --upgrade pip setuptools

RUN mkdir /code
COPY ./ /code

```

PR #582, opened 2025-10-06
github: add pull request endpoint (bug 1991125)

zzzeid:
WIP DO NOT PUSH

### DIFF
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -9,13 +9,18 @@ from django.utils.decorators import method_decorator
 from django.views import View
 from django.views.decorators.csrf import csrf_exempt
 
-from lando.main.models import CommitMap
+from lando.main.models import CommitMap, Repo
 from lando.main.models.revision import DiffWarning, DiffWarningStatus
 from lando.main.scm import (
     SCM_TYPE_GIT,
     SCM_TYPE_HG,
 )
+from lando.utils.github import GitHubAPIClient, PullRequest
 from lando.utils.phabricator import get_phabricator_client
+
+
+class APIView(View):
+    pass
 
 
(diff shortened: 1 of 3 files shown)


src/lando/urls.py
```diff
@@ -83,6 +84,14 @@
     ),
 ]
 
+urlpatterns += [
+    path(
+        "api/pulls/<str:repo_name>/<int:number>",
```
shtrom, line 89: Github uses `pulls` rather than `pull`. I think we should match it to avoid confusion [though I think lando uses plurals in general otherwise, so it's a matter of deliberately picking what to be consistent with].

shtrom: Ok, the GitHub API uses plural. Let's stick to that!

zzzeid: I *think* that's how I had it already/originally? 🤷

src/lando/api/views.py
```diff
@@ -148,3 +155,22 @@ class hg2gitCommitMapView(CommitMapBaseView):
...
+        pull_request = PullRequest(client.get_pull_request(number))
+        return JsonResponse(pull_request.serialize(), status=200)
+
+
+class LandingJobAPIView(View):
+    """Handle landing jobs in the API."""
+
+    def post(self, request: WSGIRequest, *args, **kwargs):
+        """Placeholder for creating new landing jobs."""
+        pass
```
shtrom: Oh! I discovered recently that if something empty has a docstring, it doesn't need a pass. The docstring is enough to satisfy the parser that there is a body.

The `pass` doesn't hurt, of course, and I still forget about it, but heh (:

shtrom (approved):

PR #1044, opened 2026-03-31
headless_api: gate request on new Repo.automation_permission (bug 1971103)

shtrom:
* repo: add `automation_permission` and `user_can_use_automation()`
 * headless_api: gate API request to user having the correct repo's `required_automation_permission`
 * test: test headless_api gating on repo `required_automation_permission`
 * repo: fail closed on empty permissions strings
 * utils: add stub data migration command (bug 2013991)


cgsheeh (dismissed):
One more thing that isn't clear to me from this is whether we intend to use scm levels as the `automation_permission`. I guess these will be the `SCM_ALLOW_DIRECT_PUSH` permission in most cases? I think we could make an argument that we could manage these permissions without SCM levels, for example our own custom `create_automation_firefox`. This would allow us to use Django groups to manage access to teams (sheriffs, thunderbird, etc) without having to use the manual checks we have for SCM levels.

shtrom:
> One more thing that isn't clear to me from this is whether we intend to use scm levels as the `automation_permission`. I guess these will be the `SCM_ALLOW_DIRECT_PUSH` permission in most cases? I think we could make an argument that we could manage these permissions without SCM levels, for example our own custom `create_automation_firefox`. This would allow us to use Django groups to manage access to teams (sheriffs, thunderbird, etc) without having to use the manual checks we have for SCM levels.

With the change, we're aiming to continue have that link to SCM levels in LDAP. My reasoning is that doing so lines up with current access practices and processes, which means we don't need to manage it in an ad hoc manner in Lando.

With the coming comm and nss migration, this will likely require the creation of dedicated LDAP groups similar to `SCM_ALLOW_DIRECT_PUSH` for those projects, but this is a one-off change.

src/lando/main/models/repo.py
```diff
@@ -212,6 +212,10 @@ def path(self) -> str:
 
     # Use this field to enable/disable access to this repo via the automation API.
     automation_enabled = models.BooleanField(default=False)
+    automation_permission = models.CharField(
+        default="main.scm_direct_push",
```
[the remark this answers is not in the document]
shtrom: None by default would be good, but this means that firefox repos will find themselves without direct push access when we deploy that change. Not a big deal, but something to remember.

We can also add a dedicated data-migration step that would set it correctly for fx repos on migrate, but we don't have a very clean way to do this at the moment https://bugzilla.mozilla.org/show_bug.cgi?id=2013991 @zzzeid maybe it's an opportunity to work on this pattern?

src/lando/main/models/repo.py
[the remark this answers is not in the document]
shtrom: Fixed and improved test coverage.

zzzeid (approved):

src/lando/main/models/repo.py
```diff
@@ -395,28 +399,47 @@ def phab_identifier(self) -> str | None:
...
 
-            app_label, codename = self.required_permission.split(".", maxsplit=1)
+            app_label, codename = permission.split(".", maxsplit=1)
             return user.user_permissions.filter(
                 content_type__app_label=app_label, codename=codename
             ).exists()
 
         # If the user is not a superuser, we can skip the DB round-trip.
-        return self.required_permission in user.get_user_permissions()
+        return permission in user.get_user_permissions()
```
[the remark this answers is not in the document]
shtrom: Yep. This is just extending on the previous fix, as I'm reusing some of the pre-existing machinery (from the temporary fix 😅)

src/lando/utils/management/commands/migrate_data.py
shtrom, line 1: @zzzeid here's a very stubby data-migration script built for immediate purpose, and we can iterate on it when working on bug 2013991 for real.

src/lando/main/models/repo.py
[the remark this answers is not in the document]
shtrom: 🌮 I added a stub migration script, that we can run manually during the upgrade.

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -0,0 +1,99 @@
...
+
+from lando.main.models.repo import Repo
+
+
+class Command(BaseCommand):
+    help = """This stub commands allows to run data migrations.
+
+    It should be fleshed out as part of bug 2013991.
+
+    TODO: record already-applied migrations.
```
zzzeid: nit: not sure this is needed in the help text, maybe just as a comment.

src/lando/utils/management/commands/migrate_data.py
zzzeid: Sounds good. I have some comments but I can save them to a different discussion when we are working on bug 2013991 for real.

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -0,0 +1,99 @@
...
+            if input().lower() != "y":
+                raise CommandError("User interruption")
+
+            self.stdout.write("Progress: ", ending="")
+
+    #
+    # MIGRATION METHODS
+    #
+
+    def migrate_bug1971103_add_firefox_required_automation_permissions(
```
zzzeid: nit: maybe OK to actually implement a numbering system now (e.g., call this `0001_migrate_...`). Though I think in the long term we should put migrations in their own management commands (maybe even add a `lando adddatamigration` command) and use call_command instead.

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -0,0 +1,99 @@
...
+            raise SystemExit()
+
+        fx_repo_names = ", ".join(repo.name for repo in fx_repos)
+        self._get_confirmation(
+            ask_confirm, "Updating required_automation_permission for: ", fx_repo_names
+        )
+
+        for repo in fx_repos:
+            if not repo.required_automation_permission:
+                repo.required_automation_permission = "main.scm_allow_direct_push"
```
zzzeid: nit: we could use models.profile.SCM_ALLOW_DIRECT_PUSH.

src/lando/conftest.py
```diff
@@ -1054,10 +1056,25 @@ def headless_permission():
 
 
 @pytest.fixture
-def headless_user(user, headless_permission):
+def direct_push_permission():
+    content_type = ContentType.objects.get_for_model(Profile)
+    perm, created = Permission.objects.get_or_create(
+        codename="scm_allow_direct_push", content_type=content_type
+    )
```
zzzeid, line 1063: This should be created as part of migrations running in tests.

src/lando/main/models/repo.py
```diff
@@ -395,28 +400,49 @@ def phab_identifier(self) -> str | None:
...
-            User to check permmission.
+            User to check permission.
+
+        permissions: str
+            Permission string to look for. If empty, fail closed denying any access.
 
         Returns:
             bool: whether the user has the permission
         """
+        # Fail closed if the permission is empty.
```
zzzeid, line 433: Fail closed? Typo?

src/lando/main/models/repo.py
```diff
@@ -395,28 +400,49 @@ def phab_identifier(self) -> str | None:
...
+        Test that the user has a specific permission directly rather than via a role.
 
         If the user is a superuser, this checks that the user was given the permissions
         directly, rather than transitively by virtue of being an admin.
 
         Parameters:
 
         user: User
-            User to check permmission.
+            User to check permission.
```
zzzeid: nit:
```suggestion
            User for whom to check permission.
```

src/lando/main/models/repo.py
```diff
@@ -394,16 +398,25 @@ def phab_identifier(self) -> str | None:
...
     def user_can_push(self, user: User) -> bool:
+        return self._user_has_direct_permission(user, self.required_permission)
+
+    def user_can_use_automation(self, user: User) -> bool:
+        return self._user_has_direct_permission(user, self.automation_permission)
+
+    def _user_has_direct_permission(self, user: User, required_permission: Permission):
         """
-        Test that the user has permission to push to this repo.
+        Test that the user has a specific permission directly rather than via a role.
```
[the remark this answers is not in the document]
zzzeid: FWIW I think this will be resolved in a much cleaner way once we stop using superusers in Lando.

zzzeid (approved):

src/lando/main/models/repo.py
shtrom: In case of unexpected things, we break-out and disallow. https://authzed.com/blog/fail-open

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if options.get("list"):
+            self.stdout.write("Available data migrations:\n")
+            for migration in [
+                method.removeprefix("migrate_")
+                for method in dir(self)
+                if method.startswith("migrate_")
+            ]:
+                self.stdout.write(f"* {migration}")
+                raise SystemExit()
```
cgsheeh: This should be de-dented 1 level, if the intention is to print all the available migrations.

src/lando/conftest.py
shtrom: Do we already have a set of test-only migrations? I looked around but couldn't identify any.

shtrom: Ok, it was already created, so I just used `get` instead of `get_or_create`.

cgsheeh (approved):

src/lando/utils/management/commands/migrate_data.py
shtrom: oh! good catch

shtrom:
Rebase and force-push to work around landing problems https://lando.moz.tools/landings/35783/

shtrom:
Rebase and squash fixups... https://lando.moz.tools/landings/35790/

shtrom:
Squish squash 🙄 https://lando.moz.tools/landings/35793/

PR #1067, opened 2026-04-08
phab: fix operator precedence logic in `get_phabricator_client` (Bug 2030455)

cgsheeh:
Fix the operator precedence logic in `get_phabricator_client`
so we only consider the app-wide Phab keys when an `api_key`
is not passed. Add a test that fails before the fix is
applied to confirm the behaviour and fix.

Add a `backfill_phids` management command which populates
the PHID field on all existing Phabricator API keys to resolve
this issue for existing profiles.

### DIFF
--- a/src/lando/main/management/commands/backfill_phids.py
+++ b/src/lando/main/management/commands/backfill_phids.py
@@ -1,0 +1,93 @@
+import argparse
+import logging
+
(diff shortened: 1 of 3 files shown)


zzzeid (approved):
Probably the backfill script should be deleted after it is run.

PR #1070, opened 2026-04-10
pull_request: improve warnings and blockers UI (bug 2024930)

zzzeid:
- add icons beside warnings/blockers to make them more prominent
- and check mark when there are no warnings/blockers
- add spinners when loading warnings/blockers
- other small html tweaks

### DIFF
--- a/src/lando/static_src/legacy/js/components/Stack.js
+++ b/src/lando/static_src/legacy/js/components/Stack.js
@@ -154,20 +154,20 @@ $.fn.stack = function() {
                             var has_warnings = warnings.length !== 0;
 
                             if (!has_blockers) {
-                                $("#blockers").html("None found.");
+                                $("#blockers").html(`<li><span class="fa-li has-text-success"><i class="fa fa-check"></i></span>None found.</li>`);
                             } else {
                                 $("#blockers").html("");
                                 for (var blocker of blockers) {
-                                    $("#blockers").append(`<li>${blocker}</li>`);
+                                    $("#blockers").append(`<li><span class="fa-li has-text-danger"><i class="fa fa-ban"></i></span>${blocker}</li>`);
                                 }
                             }
 
                             if (!has_warnings) {
-                                $("#warnings").html("None found.");
+                                $("#warnings").html(`<li><span class="fa-li has-text-success"><i class="fa fa-check"></i></span>None found.</li>`);
                             } else {
                                 $("#warnings").html("");
                                 for (var warning of warnings) {
-                                    $("#warnings").append(`<li>${warning}</li>`);
+                                    $("#warnings").append(`<li><span class="fa-li has-text-warning"><i class="fa fa-warning"></i></span>${warning}</li>`);
                                 }
                             }
 
--- a/src/lando/ui/jinja2/stack/pull_request.html
+++ b/src/lando/ui/jinja2/stack/pull_request.html
@@ -15,31 +15,31 @@
                     data-csrf-token="{{ csrf_token }}">Loading</button>
         </p>
     </p>
-    <p class="acknowledge-warnings-section">
-        <label class="checkbox">
-            <input type="checkbox" id="acknowledge-warnings" autocomplete="off" />
-            Acknowledge all warnings below to continue.
-        </label>
-    </p>
     <div class="StackPage-stack">
+        <p class="acknowledge-warnings-section">
+            <label class="checkbox">
+                <input type="checkbox" id="acknowledge-warnings" autocomplete="off" />
+                Acknowledge all warnings below to continue.
+            </label>
+        </p>
         <table>
             <tr>
                 <th>Warnings</th>
                 <td>
-                    <ul id="warnings">
+                    <ul id="warnings" class="fa-ul">
                         <li>
-                            Checking...
-                        </li />
+                            <span class="fa-li"><i class="fa fa-spinner fa-pulse"></i></span>Checking...
+                        </li>
                     </ul>
                 </td>
             </tr>
             <tr>
                 <th>Blockers</th>
                 <td>
-                    <ul id="blockers">
+                    <ul id="blockers" class="fa-ul">
                         <li>
-                            Checking...
-                        </li />
+                            <span class="fa-li"><i class="fa fa-spinner fa-pulse"></i></span>Checking...
+                        </li>
                     </ul>
                 </td>
             </tr>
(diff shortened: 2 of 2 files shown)


src/lando/ui/jinja2/stack/pull_request.html
```diff
@@ -15,52 +15,44 @@ <h1>
...
+                        <button class="button is-link is-small is-info post-try-job"
+                                data-pull-number="{{ pull_request.number }}"
+                                data-repo-name="{{ target_repo.name }}"
+                                data-csrf-token="{{ csrf_token }}">Request a try push</button>
+                        {% if last_try_job %}
+                            {{ last_try_job }}
+                            {{ treeherder_link(last_try_job.landed_treeherder_revision, last_try_job.target_repo, "Treeherder") | safe }}
+                        {% else %}
+                            No try jobs found.
+                        {% endif %}
```
zzzeid, line 59: This was needlessly in a <ul> block.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -154,20 +154,20 @@ $.fn.stack = function() {
                             var has_warnings = warnings.length !== 0;
 
                             if (!has_blockers) {
-                                $("#blockers").html("None found.");
+                                $("#blockers").html(`<li><span class="fa-li has-text-success"><i class="fa fa-check"></i></span>None found.</li>`);
                             } else {
                                 $("#blockers").html("");
                                 for (var blocker of blockers) {
-                                    $("#blockers").append(`<li>${blocker}</li>`);
+                                    $("#blockers").append(`<li><span class="fa-li has-text-danger"><i class="fa fa-ban"></i></span>${blocker}</li>`);
```
cgsheeh, line 161: If a blocker/warning ever returns a user-controlled value (PR title, commit message, etc) we could become vulnerable to XSS. We should escape the values before adding them to the HTML, or create an element and use `.text()` to insert the text.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -154,20 +154,20 @@ $.fn.stack = function() {
...
                             }
 
                             if (!has_warnings) {
-                                $("#warnings").html("None found.");
+                                $("#warnings").html(`<li><span class="fa-li has-text-success"><i class="fa fa-check"></i></span>None found.</li>`);
                             } else {
                                 $("#warnings").html("");
                                 for (var warning of warnings) {
-                                    $("#warnings").append(`<li>${warning}</li>`);
+                                    $("#warnings").append(`<li><span class="fa-li has-text-warning"><i class="fa fa-warning"></i></span>${warning}</li>`);
```
cgsheeh, line 170: Same comment re: escaping the inputs.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -154,20 +154,20 @@ $.fn.stack = function() {
...
                                 $("#blockers").html("");
                                 for (var blocker of blockers) {
-                                    $("#blockers").append(`<li>${blocker}</li>`);
+                                    $("#blockers").append(`<li><span class="fa-li has-text-danger"><i class="fa fa-ban"></i></span>${blocker}</li>`);
                                 }
                             }
 
                             if (!has_warnings) {
-                                $("#warnings").html("None found.");
+                                $("#warnings").html(`<li><span class="fa-li has-text-success"><i class="fa fa-check"></i></span>None found.</li>`);
```
cgsheeh, line 166: There's room for de-duplication here as this is the same as the `blockers` "none found" state.

zzzeid:
Screenshot:

image

There are also some loading spinners that were added to warnings/blockers when they are being loaded.

cgsheeh (changes requested):
~~Can you provide screenshots, or a link to a revision where we can see this?~~ Edit: posted at the same time you posted the screenshots, thanks!

The commit message just says "improve", it should be expanded to include details of what is actually changing after this PR.

src/lando/static_src/legacy/js/components/Stack.js
shtrom: They do: they may return strings from github comments.

src/lando/static_src/legacy/js/components/Stack.js
shtrom: Yep, we could make a little function that takes the `div` and renders a given list into it.

shtrom (commented):
LGTM, with agreement on Connor's comments.

PS: looks great!

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -152,22 +152,23 @@ $.fn.stack = function() {
...
+                            var success_placeholder = `<li><span class="fa-li has-text-success"><i class="fa fa-check"></i></span>None found.</li>`;
 
                             if (!has_blockers) {
-                                $("#blockers").html("None found.");
+                                $("#blockers").html(success_placeholder);
                             } else {
                                 $("#blockers").html("");
                                 for (var blocker of blockers) {
-                                    $("#blockers").append(`<li>${blocker}</li>`);
+                                    $("#blockers").append(`<li><span class="fa-li has-text-danger"><i class="fa fa-ban"></i></span>${blocker}</li>`);
```
cgsheeh: We should be escaping on the client-side. The API returns JSON, and it's the client's job to know where a value is being rendered and escape accordingly. `escape()` works here because the warnings/blockers end up in HTML text, but if we ever want to put the warnings in a different element we may become vulnerable to XSS again. Escaping belongs at the render site where the context is known.

cgsheeh, line 162: Escaping server-side is probably fine in this case, however thinking longer term it would make more sense to escape on the client-side instead. We're essentially trusting the input from the server to be sanitized for this use case. If we ever change the endpoint used, or put the response in a different HTML element, we could become vulnerable to XSS again.

src/lando/api/views.py
```diff
@@ -79,6 +80,11 @@ def generate_warnings_and_blockers(
     blockers += pr_checks.run(pr_blockers, pull_request)
     pr_warnings = [chk.name() for chk in ALL_PULL_REQUEST_WARNINGS]
     warnings = pr_checks.run(pr_warnings, pull_request)
+
+    # Sanitize blockers and warnings as they may be rendered in a page.
+    warnings = [escape(warning) for warning in warnings]
```
cgsheeh: If we ever want to display the warnings somewhere else (lando-cli, as a warning consumed by some other CI task, etc) we will need to work around the fact this endpoint returns HTML escaped text.

Ideally we would write a test for this so we don't regress in the future.

zzzeid: >we will need to work around the fact this endpoint returns HTML escaped text

Hmm, this is a good point. There's always a chance that one of the escaped characters would end up in the warning/blocker... Good compromise maybe is to put an `escape=True` argument so at least this remains exposed in the future.

PR #1071, opened 2026-04-11
uplifts: allow uplift requests before landing on autoland (Bug 2030756)

cgsheeh:
Previously, requesting an uplift required the patch to have already
landed on autoland so that Lando had a `Revision` record with patch
data. This blocked sec-related uplifts where developers had to wait
for sec-approval before landing on autoland, and only then could
request the uplift.

Now, when an uplift request is submitted, Lando fetches the revision
and diff data directly from Phabricator for any revisions that don't
already exist in the database.

The `Revision` creation logic from the landing POST endpoint has been
extracted into reusable helpers that both the landing and uplift
flows now share.

### DIFF
--- a/src/lando/api/legacy/revisions.py
+++ b/src/lando/api/legacy/revisions.py
@@ -1,11 +1,19 @@
 import logging
 from collections import Counter
+from datetime import datetime
 from typing import (
     Any,
     NamedTuple,
     Optional,
 )
 
+from django.db import transaction
+
+from lando.api.legacy.stacks import (
+    get_diffs_by_phid,
+    get_revisions_by_id,
+)
+from lando.main.models import Revision
 from lando.utils.phabricator import (
     PhabricatorClient,
     PhabricatorRevisionStatus,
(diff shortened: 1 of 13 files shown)


src/lando/api/legacy/revisions.py
```diff
@@ -239,3 +247,124 @@ def find_title_and_summary_for_landing(
...
+
+    patch_data = {
+        "author_name": author_name or "",
+        "author_email": author_email or "",
+        "commit_message": commit_message,
+        "timestamp": str(int(datetime.now().timestamp())),
+    }
+
+    with transaction.atomic():
+        revision, created = Revision.objects.update_or_create(
```
shtrom, line 283: `created` is not used. Should it be? The previous code wasn't using the same path, so perhaps it doesn't matter to know if the `Revision` is new here (I feel like it doesn't)

src/lando/api/legacy/stacks.py
```diff
@@ -77,15 +70,86 @@ def request_extended_revision_data(
...
+
+    diffs = phab.call_conduit_collated(
+        "differential.diff.search",
+        constraints={"phids": diff_phids},
+        attachments={"commits": True},
+    )
+    return result_list_to_phid_dict(phab.expect(diffs, "data"))
+
+
+def request_extended_revision_data(
```
shtrom: nit: I'd keep the entry-function at the top (step-down style). I think it would also make the diff easier to read, showing that bits are replaced by functions whose body is moved further down.

src/lando/api/legacy/stacks.py
```diff
@@ -77,15 +70,86 @@ def request_extended_revision_data(
...
-    revs = result_list_to_phid_dict(phab.expect(revs, "data"))
+    return result_list_to_phid_dict(phab.expect(revs, "data"))
+
+
+def get_revisions_by_id(
+    phab: PhabricatorClient, revision_ids: list[int]
+) -> dict[str, dict]:
+    """Fetch revisions from Phabricator by integer ID.
 
+    Returns a PHID-keyed dict of revision data with reviewer and project
```
shtrom, line 118: PHID or ID-keyed?

src/lando/api/tests/test_uplift.py
```diff
@@ -195,38 +195,78 @@ def test_uplift_creation_fails_when_revisions_missing(
...
+    url = reverse("uplift-page")
+    form_data = {
+        "source_revisions": [999999],
+        "repositories": ["firefox-beta"],
+    }
+    form_data |= CREATE_FORM_DATA
+
+    response = authenticated_client.post(url, data=form_data, HTTP_REFERER="/D999999")
+
+    assert response.status_code == 302, "Failed seeding should redirect."
```
shtrom, line 262: nit: Where should it redirect? Should we check that, too?

shtrom (approved):
r+ with a few minor things

src/lando/api/legacy/revisions.py
cgsheeh: The previous code was doing [this](https://github.com/mozilla-conduit/lando/blob/9e90472ccb3b9c297dbf7720762cd8b65f2935de/src/lando/api/legacy/api/transplants.py#L308), which is the same as "get or create" followed by updating the `diff_id`, which we just roll into one operation here. We don't need to check the `created` flag in this case.

src/lando/api/legacy/stacks.py
cgsheeh: PHID keyed for the return type. That's why the function signature is `dict[str, dict]` and not `dict[int, dict]`. See also `result_list_to_phid_dict` which is called as the return value.

PR #1073, opened 2026-04-13
api.views: add pull request description webhook (bug 2024944)

zzzeid:
- add view to be called by GitHub to update pull request with data from Lando
- generate pull request description containing auto-generated portion and user-inputted portion
- separate auto-generated and user-inputted portions by special delimiter
- add pr_description.md jinja2 template
- add patch methods in GitHub API client

### DIFF
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -5,15 +5,18 @@ from functools import wraps
 from typing import Callable
 
 from django import forms
+from django.conf import settings
 from django.core.handlers.wsgi import WSGIRequest
 from django.http import Http404, HttpRequest, HttpResponse, JsonResponse
+from django.template.loader import render_to_string
+from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.utils.html import escape
 from django.views import View
 from django.views.decorators.csrf import csrf_exempt
 from requests import HTTPError
 
-from lando.api.legacy.commit_message import replace_reviewers
+from lando.api.legacy.commit_message import parse_bugs, replace_reviewers
 from lando.main.auth import PrivateRepoPermissionMixin, require_authenticated_user
 from lando.main.models import (
     CommitMap,
@@ -26,7 +29,12 @@ from lando.main.models import (
 from lando.main.models.landing_job import get_jobs_for_pull
 from lando.main.models.revision import DiffWarning, DiffWarningStatus
 from lando.main.scm import SCMType
-from lando.utils.github import GitHubAPIClient, PullRequest, PullRequestPatchHelper
+from lando.utils.github import (
+    PR_DELIMITER,
+    GitHubAPIClient,
+    PullRequest,
+    PullRequestPatchHelper,
+)
 from lando.utils.github_checks import (
     ALL_PULL_REQUEST_BLOCKERS,
     ALL_PULL_REQUEST_WARNINGS,
@@ -368,3 +376,42 @@ class PullRequestContentAPIView(PullRequestAPIView):
             form.cleaned_data["title"],
         )
         return HttpResponse(status=204)
+
+
+@method_decorator(csrf_exempt, name="dispatch")
+class PullRequestUpdateWebhook(PullRequestAPIView):
+    """API method called by GitHub to update Lando data in pull request."""
+
+    def generate_context(self, request: HttpRequest) -> dict[str, str | list[str]]:
+        """Generate various context variables used in rendering the template."""
+        context = generate_warnings_and_blockers(
+            self.target_repo, self.pull_request, request
+        )
+
+        path = reverse(
+            "pull-request",
+            kwargs={
+                "repo_name": self.target_repo.name,
+                "number": self.pull_request.number,
+            },
+        )
+
+        context["lando_url"] = f"{settings.SITE_URL}{path}"
+        context["pr_delimiter"] = PR_DELIMITER
+        bugs = parse_bugs(self.pull_request.title)
+        context["bugs"] = bugs
+        context["title"] = self.pull_request.title
+
+        # This commit body refers to the portion of the description below the
+        # delimiter (i.e., user-inputted value).
+        context["commit_body"] = self.pull_request.commit_body
+        return context
+
+    def post(
+        self, request: WSGIRequest, repo_name: str, pull_number: int
+    ) -> JsonResponse:
+        """Generate content and update PR description."""
+        context = self.generate_context(request)
+        rendered = render_to_string("pr_description.md", context)
+        self.client.update_pull_request_content(pull_number, rendered)
+        return JsonResponse({"status": "success"})
--- a/src/lando/ui/jinja2/pr_description.md
+++ b/src/lando/ui/jinja2/pr_description.md
@@ -1,0 +1,33 @@
+{# Helpers for generating warnings and blockers #}
+{% macro _warning(content) %}:warning: {{ content }}{% endmacro %}
+{% macro _blocker(content) %}:no_entry_sign: {{ content }}{% endmacro %}
+{% macro _column(content) %}|{{ content }}|{% endmacro %}
+{% macro _header() %}|---------|{% endmacro %}
+{% macro _table(title, entries, macro) %}
+{{ _column(title) }}
+{{ _header() }}
+{% for entry in entries %}
+{{ macro(entry) }}
+{% endfor %}
+{% endmacro %}
+{% block upper %}
+{{ commit_body|safe }}
+{% endblock %}
+{{ pr_delimiter|safe }}
+---
+{% block lower %}
+Lando: [link]({{ lando_url }})
+{% if bugs %}Bugzilla: {% for bug in bugs %}[bug {{ bug }}]({{ bug|bug_url }}){% if not loop.last %}, {% endif %}{% endfor %}{% endif %}
+
+{% if not warnings and not blockers %}
+:white_check_mark: All Lando checks passed
+{%- endif %}
+{% if warnings %}
+
+{{ _table("Warnings", warnings, _warning) }}
+{%- endif %}
+{% if blockers %}
+
+{{ _table("Blockers", blockers, _blocker) }}
+{%- endif %}
+{% endblock %}
--- a/src/lando/urls.py
+++ b/src/lando/urls.py
@@ -26,6 +26,7 @@ from lando.api.views import (
     LegacyDiffWarningView,
     PullRequestChecksAPIView,
     PullRequestContentAPIView,
+    PullRequestUpdateWebhook,
     git2hgCommitMapView,
     hg2gitCommitMapView,
 )
@@ -130,6 +131,11 @@ urlpatterns += [
         "api/pulls/<str:repo_name>/<int:pull_number>",
         PullRequestContentAPIView.as_view(),
         name="api-pull-request-update-content",
+    ),
+    path(
+        "api/pulls/<str:repo_name>/<int:pull_number>/webhook",
+        PullRequestUpdateWebhook.as_view(),
+        name="api-pull-request-description",
     ),
 ]
 
--- a/src/lando/utils/github.py
+++ b/src/lando/utils/github.py
@@ -22,6 +22,11 @@ from lando.utils.const import URL_USERINFO_RE
 logger = logging.getLogger(__name__)
 
 
+PR_DELIMITER = (
+    "<!--/ -+-+- DO NOT MODIFY THIS LINE - ENTER COMMIT MESSAGE ABOVE -+-+- /-->"
+)
+
+
 class GitHub:
     """Work with authentication to GitHub repositories."""
 
@@ -242,11 +247,11 @@ class GitHubAPIClient:
         elif content_type == "application/vnd.github.diff; charset=utf-8":
             return result.text
 
-    def _post(self, path: str, *args, **kwargs):
+    def _post(self, path: str, *args, **kwargs) -> dict:
         result = self._api.post(path, *args, **kwargs)
         return result.json()
 
-    def _patch(self, path: str, *args, **kwargs):
+    def _patch(self, path: str, *args, **kwargs) -> dict:
         result = self._api.patch(path, *args, **kwargs)
         return result.json()
 
@@ -391,12 +396,15 @@ class GitHubAPIClient:
         )
 
     def update_pull_request_content(
-        self, pull_number: int, body: str, title: str
+        self, pull_number: int, body: str, title: str | None = None
     ) -> dict:
-        """Update the pull request description with provided body and title."""
+        """Update the pull request description with provided body and optional title."""
+        params = {"body": body}
+        if title is not None:
+            params["title"] = title
         return self._patch(
             f"{self.repo_base_url}/issues/{pull_number}",
-            json={"body": body, "title": title},
+            json=params,
         )
 
     @classmethod
@@ -466,6 +474,16 @@ class PullRequest:
     def __repr__(self) -> str:
         return f"Pull request #{self.number} ({self.head_repo_git_url})"
 
+    @staticmethod
+    def _parse_body_segments(body: str) -> tuple[str, str]:
+        """Return the commit body, delimited by the PR_DELIMITER or an empty string when not possible."""
+        if not body:
+            return ""
+        parts = body.split(PR_DELIMITER)
+
+        # Return the user-controlled portion.
+        return parts[0].strip()
+
     def __init__(self, client: GitHubAPIClient, data: dict):
         self.client = client
 
@@ -484,6 +502,7 @@ class PullRequest:
         self.diff_url = data["diff_url"]
         self.patch_url = data["patch_url"]
         self.body = data["body"] or ""  # description
+        self.commit_body = self._parse_body_segments(self.body)
         self.is_draft = data["draft"]
         self.comments_url = data["comments_url"]
         self.commits_url = data["commits_url"]
@@ -641,8 +660,8 @@ class PullRequest:
 
         lines = [self.title, ""]
 
-        if self.body:
-            lines += [self.body, ""]
+        if self.commit_body:
+            lines += [self.commit_body, ""]
 
         lines.append(f"Pull request: {self.html_url}")
 
--- a/src/lando/api/tests/test_views.py
+++ b/src/lando/api/tests/test_views.py
@@ -4,6 +4,7 @@ import pytest
 from django.test import Client
 
 from lando.main.models import Repo, SCMType
+from lando.utils.github import PullRequest
 
 
 @pytest.fixture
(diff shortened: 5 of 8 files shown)


Magnolia-Liu (approved):

src/lando/api/views.py
```diff
@@ -326,3 +334,52 @@ def get(
...
+        return context
+
+    @method_decorator(require_authenticated_user)
+    def post(
+        self, request: WSGIRequest, repo_name: str, pull_number: int
+    ) -> JsonResponse:
+        """Generate content and update PR description."""
+
+        if not request.user.has_perm("main.can_change_landing_job"):
+            raise PermissionError()
```
cgsheeh: This is the OS-level `PermissionError`, but you want to use the Django-level `PermissionDenied`. In the current state this becomes a 500 instead of a 403.

src/lando/utils/github.py
```diff
@@ -22,6 +22,9 @@
 logger = logging.getLogger(__name__)
 
 
+SPECIAL_DELIMITER = "-" * 9
```
cgsheeh: This variable name could be more descriptive, something like `PR_SUMMARY_DELIMITER` would make it clearer why this is "special".

This delimiter is also a very common pattern in markdown, there's a real risk of detecting this string in user-input instead of the Lando-specific value. Could we use something specific to Lando to avoid potentially triggering the delimit behaviour erroneously? Maybe something like `<!--LANDO-DESCRIPTION-DELIMITER-->`?

src/lando/jinja.py
```diff
@@ -450,6 +450,7 @@ def message_type_to_notification_class(flash_message_category: str) -> str:
 
 
 def environment(**options) -> Environment:
+    options.update({"trim_blocks": True, "lstrip_blocks": True})
```
cgsheeh, line 454: What do we need this for? Is it really required at the global level, or could we apply it to just the template rendering introduced in this PR? A comment would make this clearer.

src/lando/api/views.py
```diff
@@ -326,3 +334,52 @@ def get(
...
+            },
+        )
+
+        context["lando_url"] = f"{settings.SITE_URL}{path}"
+        context["special_delimiter"] = SPECIAL_DELIMITER
+        bugs = parse_bugs(self.pull_request.title)
+        context["bugs"] = []
+
+        for bug in bugs:
+            context["bugs"].append((bug, f"{settings.BUGZILLA_URL}/{bug}"))
```
cgsheeh: Could we use `linkify_bug_numbers` or `bug_url` here instead? Also, this can be a list comprehension.

src/lando/utils/github.py
```diff
@@ -451,6 +470,9 @@ def __init__(self, client: GitHubAPIClient, data: dict):
         self.diff_url = data["diff_url"]
         self.patch_url = data["patch_url"]
         self.body = data["body"]  # description
+        self.parsed_body = (
+            self.body.split(SPECIAL_DELIMITER)[-1].strip() if self.body else ""
```
cgsheeh: Perhaps instead of just taking -1 here, we should collect and join everything after the first split (ie `[1:]`). If the user has the special delimiter in their description for some reason, we risk wiping away their commit message by accident. We should probably also split by lines and assume the delimiter is on a dedicated line.

src/lando/ui/jinja2/pr_description.md
```diff
@@ -0,0 +1,37 @@
...
+{% macro _header() %}|---------|{% endmacro %}
+{% macro _table(title, entries, macro) %}
+{{ _column(title) }}
+{{ _header() }}
+{% for entry in entries %}
+{{ macro(entry) }}
+{% endfor %}
+{% endmacro %}
+
+{% block upper %}
```
cgsheeh, line 13: Question: is the plan for something to extend this template in the future? We could drop the upper/lower blocks until then, seems like unnecessary complexity right now.

src/lando/api/views.py
```diff
@@ -326,3 +334,52 @@ def get(
             # The StaleMetadataException error message is safe for user consumption.
             return JsonResponse({"errors": [str(exc)]}, status=500)
         return JsonResponse(warnings_and_blockers)
+
+
+@method_decorator(csrf_exempt, name="dispatch")
+class PullRequestUpdateWebhook(PullRequestAPIView):
+    """API method called by GitHub to update Lando data in pull request."""
+
+    def generate_context(self, request: HttpRequest) -> dict[str, str]:
```
cgsheeh: The values are a mix of different types, not just `str`:

```suggestion
    def generate_context(self, request: HttpRequest) -> dict[str, Any]:
```

src/lando/api/views.py
```diff
@@ -326,3 +334,52 @@ def get(
...
+        """Generate various context variables used in rendering the template."""
+        context = {}
+        context.update(
+            generate_warnings_and_blockers(self.target_repo, self.pull_request, request)
```
cgsheeh: Question: should we catch `StaleMetadataException` here as we do in `PullRequestChecksAPIView`?

src/lando/ui/jinja2/pr_description.md
```diff
@@ -0,0 +1,37 @@
...
+{{ _table("Blockers", blockers, _blocker) }}
+{% endif %}
+
+{% endblock %}
+
+<!--/ DO NOT REMOVE THE LINE BELOW /-->
+{{ special_delimiter }}
+
+{% block lower %}
+{{ body|e }}
```
cgsheeh: As this is a markdown file, sending the body through `escape` will escape it as though it were HTML, and this has the opportunity to compound if the user updates their description multiple times. For example, the text `Fixes a < b && c logic error` -> `a &lt; b &amp;&amp; c` -> `a &amp;lt; b &amp;amp;&amp;amp; c`, etc.

cgsheeh (changes requested):

zzzeid:
@cgsheeh thanks for the early feedback, this PR was not ready for review I think it was accidentally marked as such and then reverted.

src/lando/ui/jinja2/pr_description.md
zzzeid: Though this is a markdown file, there's the possibility of unescaped html ending up in this particular field, which ultimately ends up on the GitHub PR description (which accepts some HTML, e.g., the `<strong>` tag). This is a common scenario in commit message (e.g., referencing certain tags, etc).

Good point though -- I'll make sure this scenario is handled.

src/lando/utils/github.py
zzzeid: This was a temporary placeholder, a more complex delimiter with a lower possibility of collision will be added before this PR is ready for review.

src/lando/jinja.py
zzzeid: I could file a bug for this. In my opinion [this behaviour](https://jinja.palletsprojects.com/en/stable/templates/#whitespace-control) is more desirable in our case than the default behaviour, across all templates.

src/lando/ui/jinja2/pr_description.md
zzzeid: Yes, see https://github.com/mozilla-conduit/lando/pull/1088/changes#diff-2f0f1e5dce6ab1f09c9afda9f39c85872df5b213360f6433244019403127abba (still WIP).

src/lando/api/views.py
zzzeid: I don't think this is necessary. This hook will be called upon every update of the pull request on GitHub. It's also meant to be informative, the final warnings and blockers are generated on the Lando page.

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=2000955.

src/lando/jinja.py
zzzeid: See #1181.

src/lando/ui/jinja2/pr_description.md
zzzeid: To get this out the door, I'm going to [file a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2044799) for this, and instead of using `|e`, I'll use `|striptags` which is the safest approach here. The downside is that legitimate tags (e.g., part of a commit message referencing a tag) will be stripped from the description. I'll add a note about this, but we can do a follow up fix.

FWIW having tags in the commit message already causes some problems in our Phabricator setup.

EDIT: the original case that you referenced will be handled already, i.e., stray characters are not escaped, as they are not actual tags.

src/lando/ui/jinja2/pr_description.md
```diff
@@ -0,0 +1,36 @@
...
+{% endif %}
+
+{% endblock %}
+
+---
+{{ pr_delimiter|safe }}
+
+{% block lower %}
+{{ commit_body|striptags }}
+{% endblock %}
```
shtrom: nit: I'd expect the original text towards the top, with dynamic updates at the bottom.

src/lando/urls.py
```diff
@@ -125,6 +126,11 @@
         PullRequestChecksAPIView.as_view(),
         name="api-pull-request-checks",
     ),
+    path(
+        "api/pulls/<str:repo_name>/<int:pull_number>/webhook",
```
shtrom, line 136: What payload do we expect GitHub to send here?

src/lando/api/views.py
```diff
@@ -334,3 +342,44 @@ def get(
             # The StaleMetadataException error message is safe for user consumption.
             return JsonResponse({"errors": [str(exc)]}, status=500)
         return JsonResponse(warnings_and_blockers)
+
+
+@method_decorator(csrf_exempt, name="dispatch")
+class PullRequestUpdateWebhook(PullRequestAPIView):
+    """API method called by GitHub to update Lando data in pull request."""
```
shtrom, line 383: Is this docstring accurate? This doesn't look like a webhook to be called by GitHub to me. It looks like it's called by an authenticated lando user to update a GitHub PR description, so Lando->GitHub, rather than the GitHub->Lando flow I'd expect for a GitHub webhook.

zzzeid: Yes it is accurate. `@method_decorator(require_authenticated_user)` was added as a fail-safe, it is removed in #1219 in favour of signature validation.

GitHub calls this webhook, this webhook calls GitHub back and updates the PR with all the necessary changes based on the current state in Lando.

src/lando/ui/jinja2/pr_description.md
zzzeid: I did test this layout out, and it seemed more logical to put status-related things above the commit message, rather than below, to grab the user's attention. Here are the two examples:

Current state:
Image


Commit message on top:

Image

src/lando/urls.py
zzzeid: There is some information about this in the [docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads#example-webhook-delivery). With this current state we are ignoring the payload, but in future iterations we can leverage that payload for more specific functionality (e.g., ignoring requests when it's unnecessary to take any action, etc...)

src/lando/api/views.py
zzzeid: Removed this, since all these PRs will likely land at the same time.

src/lando/ui/jinja2/pr_description.md
zzzeid: @cgsheeh @Magnolia-Liu @marco-c any thoughts here?

marco-c: I also would have expected the warnings and blockers to be below the PR description, not sure exactly why my brain expects it that way.
It also follows the usual GitHub UI where you have PR description and then status checks at the bottom (if we consider these to be similar in scope to status checks, and in fact some of them could maybe become status checks in the future).

I do see the point of catching user's attention, though I guess it'll still be super visible because descriptions are usually short enough and warnings and blockers are very prominent (given they are tables and there are yellow and red icons).

shtrom (approved):

src/lando/ui/jinja2/pr_description.md
zzzeid: Will flip the order.

zzzeid: Order flipped and tests updated.

src/lando/api/views.py
```diff
@@ -334,3 +342,43 @@ def get(
...
+        bugs = parse_bugs(self.pull_request.title)
+        context["bugs"] = bugs
+        context["title"] = self.pull_request.title
+
+        # This commit body refers to the portion of the description below the
+        # delimiter (i.e., user-inputted value).
+        context["commit_body"] = self.pull_request.commit_body
+        return context
+
+    def post(
```
cgsheeh, line 410: How does this work for private/sec-repos? I think we'll hit `raise_404_if_needed` from the base class on every webhook call. Guess it's unsupported for now?

src/lando/utils/github.py
```diff
@@ -241,6 +249,10 @@ def _post(self, path: str, *args, **kwargs):
         result = self._api.post(path, *args, **kwargs)
         return result.json()
 
+    def _patch(self, path: str, *args, **kwargs):
```
cgsheeh: Missing return type `-> dict` here.

src/lando/utils/github.py
```diff
@@ -444,6 +463,19 @@ class Review(str, Enum):
     def __repr__(self) -> str:
         return f"Pull request #{self.number} ({self.head_repo_git_url})"
 
+    @staticmethod
+    def _parse_body_segments(body: str) -> tuple[str]:
```
cgsheeh:
```suggestion
    def _parse_body_segments(body: str) -> tuple[str, str]:
```

Also, a docstring describing the output format would be useful.

src/lando/ui/jinja2/pr_description.md
cgsheeh: Do we still need the upper/lower blocks?

src/lando/ui/jinja2/pr_description.md
```diff
@@ -0,0 +1,34 @@
...
+{% for entry in entries %}
+{{ macro(entry) }}
+{% endfor %}
+{% endmacro %}
+{% block upper %}
+{{ commit_body|striptags }}
+{% endblock %}
+
+---
+{{ pr_delimiter|safe }}
```
cgsheeh: Seems like these lines should be swapped, otherwise we will append `---` on every update.
```suggestion
{{ pr_delimiter|safe }}
---
```

Can we add test coverage to assert that the content is stable across repeated updates?

src/lando/api/views.py
```diff
@@ -334,3 +342,43 @@ def get(
...
+        self, request: WSGIRequest, repo_name: str, pull_number: int
+    ) -> JsonResponse:
+        """Generate content and update PR description."""
+        context = self.generate_context(request)
+        rendered = render_to_string("pr_description.md", context)
+        self.client.update_pull_request_body(pull_number, rendered)
```
cgsheeh: This makes an API call to Github, should we wrap it in a `try/except` or retry mechanism?

src/lando/api/views.py
```diff
@@ -334,3 +342,43 @@ def get(
...
+@method_decorator(csrf_exempt, name="dispatch")
+class PullRequestUpdateWebhook(PullRequestAPIView):
+    """API method called by GitHub to update Lando data in pull request."""
+
+    def generate_context(self, request: HttpRequest) -> dict[str, str | list[str]]:
+        """Generate various context variables used in rendering the template."""
+        context = {}
+        context.update(
+            generate_warnings_and_blockers(self.target_repo, self.pull_request, request)
+        )
```
cgsheeh: Nit: you can just initialize `context` directly from the warnings here:
```suggestion
        context = generate_warnings_and_blockers(self.target_repo, self.pull_request, request)

```

src/lando/ui/jinja2/pr_description.md
```diff
@@ -0,0 +1,34 @@
...
+{% macro _header() %}|---------|{% endmacro %}
+{% macro _table(title, entries, macro) %}
+{{ _column(title) }}
+{{ _header() }}
+{% for entry in entries %}
+{{ macro(entry) }}
+{% endfor %}
+{% endmacro %}
+{% block upper %}
+{{ commit_body|striptags }}
```
cgsheeh: This still strips content out of the user-controlled portion of the PR description. We might want to add a test to confirm this is doing what we expect, I think devs will be unhappy if we end up editing content out of their PR descriptions.

cgsheeh (changes requested):

src/lando/api/views.py
zzzeid: Yes it's unsupported for now. When we add support, we can add the webhooks to private repos too on GitHub. This likely will be part of the larger private repos story since the existing ones are barely used.

src/lando/api/views.py
zzzeid: I'll file a bug for this, we should perhaps do that in a more abstracted way for all calls to github (including adding comments, etc.)

zzzeid: https://bugzilla.mozilla.org/show_bug.cgi?id=2046546

src/lando/utils/github.py
zzzeid: We currently have an exception for that rule in our config. Not an issue though, can add it though to this and the other missing ones.

src/lando/ui/jinja2/pr_description.md
zzzeid: I filed a bug for this previously here: https://bugzilla.mozilla.org/show_bug.cgi?id=2044799. It will be fixed in a follow up, but this is a bit of an uncharted territory, since GitHub allows some html in the PR description but when transformed to a commit message, it may look quite different. It's not an issue we have in Phabricator currently.

I can update the tests to account for this.

src/lando/ui/jinja2/pr_description.md
zzzeid: Nice catch, I'll update this.

src/lando/ui/jinja2/pr_description.md
zzzeid: Yes, the PR I linked above was just moved to a different one: https://github.com/mozilla-conduit/lando/pull/1182.

src/lando/ui/jinja2/pr_description.md
zzzeid: >Can we add test coverage to assert that the content is stable across repeated updates?

This is not trivial in the current state. We need a more robust GitHub mock which can act as an intermediary and keep a concept of a PR (get/update etc.) persistent within the scope of the client session. The tests right now basically test that the GitHub API is being sent the expected payload. The issue pointed out here would be caught if we were also re-fetching a new payload from GitHub (mock) to update the state of the pull request.

I think this could fall under [bug 2005686](https://bugzilla.mozilla.org/show_bug.cgi?id=2005686) given the above scope (CC @shtrom).

Thoughts?

src/lando/ui/jinja2/pr_description.md
zzzeid: Ended up fixing this, since usage of backticks seems to make this behave exactly how we want it on GitHub.

src/lando/utils/github.py
```diff
@@ -444,6 +465,28 @@ class Review(str, Enum):
     def __repr__(self) -> str:
         return f"Pull request #{self.number} ({self.head_repo_git_url})"
 
+    @staticmethod
+    def _parse_body_segments(body: str) -> tuple[str, str]:
+        """Return a tuple delimited by the PR_DELIMITER or empty strings when not possible."""
+        if body:
```
cgsheeh: Nit: flipping this to `if not body:` would allow de-denting the entire function for readability.

src/lando/utils/github.py
```diff
@@ -444,6 +465,28 @@ class Review(str, Enum):
...
+                .strip()
+                .removesuffix("```")
+                .strip()
+            )
+            if len(parts) > 2:
+                # Something went wrong, the delimiter should only be present once.
+                raise ValueError("Multiple delimiters detected in PR description")
+            elif len(parts) == 2:
+                return tuple(parts)
+            # Assume this is the original PR description, i.e., there was no delimiter.
```
cgsheeh: Seems like we wouldn't want to strip the triple-backticks out if this is the case. We could refactor the above `elif` to check `elif len(parts) != 2` and return this default, then do the parsing only when `len(parts) == 2`.

src/lando/utils/github.py
```diff
@@ -462,6 +505,8 @@ def __init__(self, client: GitHubAPIClient, data: dict):
         self.diff_url = data["diff_url"]
         self.patch_url = data["patch_url"]
         self.body = data["body"]  # description
+
+        self.commit_body, self.lando_body = self._parse_body_segments(self.body)
```
cgsheeh: I think this should be handled more gracefully, in the current state we would throw a `ValueError` on multiple delimiters, which would break any attempt to load the PR in Lando.

src/lando/utils/github.py
```diff
@@ -444,6 +465,28 @@ class Review(str, Enum):
...
+    def _parse_body_segments(body: str) -> tuple[str, str]:
+        """Return a tuple delimited by the PR_DELIMITER or empty strings when not possible."""
+        if body:
+            parts = body.split(PR_DELIMITER)
+            parts[0] = (
+                parts[0]
+                .strip()
+                .removeprefix("```text")
+                .strip()
+                .removesuffix("```")
```
cgsheeh: Using this `text` hack causes Github to display the user-controlled portion of the PR description as if it were a monospaced code-block. It also means users can't add their own markdown to PR descriptions. I think that's enough breakage that it's worth re-considering this approach, but let me know if I'm misunderstanding what this does.

Image

Can we simply allow the user's content to pass-through untouched instead? I don't think this opens any injection risk, we don't evaluate the commit body as a template in Jinja, and the content will end up in Github's description box where `<script>` and other tags are stripped. In addressing bug 2044799, users can wrap HTML tags in backticks to display them as code when necessary. WDYT?

cgsheeh (changes requested):

src/lando/utils/github.py
zzzeid: >Using this text hack causes Github to display the user-controlled portion of the PR description as if it were a monospaced code-block

This is part of the tradeoff. Part of the problem is that GitHub renders some HTML tags but not others. On one hand we can have some rendering of markdown but various tags either partially interpreted/passed through, or interpreted (e.g., `<script>` causes strange behaviour, `<strong>` is interpreted, `<h1>` is interpreted) on the other hand we can have all the text kept as is without any rendering (markdown or otherwise) and it will match the commit message. With the `text` backticks, this problem is solved.

> the content will end up in Github's description box where <script> and other tags are stripped

That's part of the issue, we would still be losing that parity with Phabricator when it comes to what can be part of the commit message. Some tags also break the description. For example, if a `<script>` tag is referenced in a legitimate commit message (without backticks), it causes the rest of the text in the PR description to no longer be interpreted as HTML, thus breaking the display.

I don't think there's going to be a single perfect solution at this time, but keeping in mind that we will be allowing users to edit the commit message from Lando before landing, we won't have any rendering of HTML or markdown there, and whatever a user inputs there should be saved and shown as-is on GitHub without unexpected behaviour.

Do we have any data as to how often markdown is used in Phabricator descriptions? Keeping in mind those are destined to become plaintext commit messages. Other than whatever is coming in from the plain text commit message? The main thing we'll be losing here is the rendering on the GitHub PR page.

src/lando/utils/github.py
zzzeid: >Seems like we wouldn't want to strip the triple-backticks out if this is the case

I don't think this would have any impact. If it's the original description it still has to pass through the pr_description template which needs any outside backticks removed, if they happen to have been added by someone intentionally/unintentionally. There are some edge cases we are not handling (e.g., if there are backticks in the content, which may cause some strange rendering on github but I think that's out of scope.)

But upon further thinking, I think we can simply return the first part and ignore the rest in all cases.

src/lando/utils/github.py
zzzeid: I agree, raising a `ValueError` is not ideal. I think the best we can do here is either clear the whole user-controlled portion (which may cause some data loss) or assume the first part is the user-controlled portion. This should only happen if someone intentionally or unintentionally copy/pastes the delimiter.

src/lando/utils/github.py
zzzeid: I think since the current approach at least has consistent behaviour (i.e., nothing breaks in the GitHub description, no matter what the commit message content is) we should keep it as it is, gather some feedback, and change this based on feedback.

zzzeid: After some offline discussion, we'll go with keeping the commit message as-is, with the caveat that unescaped html tags (without backticks, e.g., `<script>`) may break the GitHub PR description, get rendered (e.g., <strong>test</strong>) or have unexpected other consequences. They may also break the lando generated portion in the PR description.

zzzeid: We can investigate alternate approaches or possible other discussion in this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2047204

src/lando/utils/github.py
```diff
@@ -444,6 +465,19 @@ class Review(str, Enum):
...
         return f"Pull request #{self.number} ({self.head_repo_git_url})"
 
+    @staticmethod
+    def _parse_body_segments(body: str) -> tuple[str, str]:
+        """Return the commit body, delimited by the PR_DELIMITER or an empty string when not possible."""
+        if not body:
+            return ""
+        parts = body.split(PR_DELIMITER)
+        parts[0] = (
+            parts[0].strip().removeprefix("```text").strip().removesuffix("```").strip()
```
cgsheeh: We need to remove the `remove{prefix,suffix}` lines here.

cgsheeh (changes requested):

cgsheeh (approved):

PR #1087, opened 2026-04-14
api.views: create base pull request view (bug 2002628)

zzzeid:
A few different views use common attributes (e.g., target_repo and pull_request). Those can be automatically set in a base view to avoid repetition.

### DIFF
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -203,7 +203,23 @@ class hg2gitCommitMapView(CommitMapBaseView):
     scm = SCMType.HG
 
 
-class LandingJobPullRequestAPIView(View):
+class PullRequestAPIView(View):
+    """Set various common attributes for views that extend this one."""
+
+    target_repo: Repo
+    client: GitHubAPIClient
+    pull_request: PullRequest
+
+    def dispatch(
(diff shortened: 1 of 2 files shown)


src/lando/api/views.py
```diff
@@ -203,16 +203,31 @@ class hg2gitCommitMapView(CommitMapBaseView):
     scm = SCMType.HG
 
 
-class LandingJobPullRequestAPIView(View):
+class PullRequestAPIView(View):
```
shtrom, line 206: nit: We have a basic `LandoView` in https://github.com/mozilla-conduit/lando/blob/main/src/lando/ui/views.py#L4-L5

It's a noop at the moment, but I think we should use it as the base of all our view, so we have a single base place where we control them all.

shtrom (approved):

src/lando/api/views.py
zzzeid: I agree with having a base view for all views, though I think it should go in `lando.main` (e.g., `BaseLandoView`). Then `LandoView` can be renamed to `LandoUIView`, and we can have `LandoAPIView`) for these two. I'll file a bug for this.

PR #1089, opened 2026-04-15
hg: read rejects file contents in-memory (Bug 2031725)

cgsheeh:
Use of the hard-coded `/tmp/rejects_path` directory to store
`.rej` file contents causes intermittent test failures during
parallelized test runs. Tests will attempt to clear the
directory during `clean_repo` and the later read file contents
in `process_merge_conflicts`, causing several different
intermittent failures from this single issue.

Change `HgSCM` to read the rejects content into an in-memory
instance variable, which is initialized empty and cleared at
the same time the previous implementation would `rmdir` the
directory. Move logic to an instance method and add a unit test
for the parsing logic. Expand tests to ensure parsing cannot fail on
invalid UTF-8, and that the in-memory rejects are properly
parsed on `PatchConflict`.

### DIFF
--- a/src/lando/main/scm/hg.py
+++ b/src/lando/main/scm/hg.py
@@ -5,7 +5,6 @@ import os
 import posixpath
 import re
 import shlex
-import shutil
 import subprocess
 import tempfile
 import uuid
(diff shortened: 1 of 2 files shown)


src/lando/main/scm/hg.py
```diff
@@ -172,6 +171,7 @@ class HgSCM(AbstractSCM):
 
     def __init__(self, path: str, config: dict | None = None, **kwargs):
         self.config = copy.copy(self.DEFAULT_CONFIGS)
+        self.rejects_content: dict[str, str] = {}
```
shtrom, line 175: nit: Could we forward-declare this attribute at the top of the class definition?

shtrom (approved):
r+ w/ nit

PR #1090, opened 2026-04-15
testing: monkeypatch over `settings.REPO_ROOT` instead of over-writing (Bug 2031725)

cgsheeh:
We previously overwrote the `settings.REPO_ROOT` variable in
`test__models__Repo__pulse_routing_key_validator`, which would
then cause failures in `test__models__Repo__system_path_validator`,
since the variable we used at test collection time.

Instead, `monkeypatch` over the variable in the Pulse test to
avoid polluting other tests. Also create the path variable inside
the test, instead of at module collection time, to avoid regressing
this issue in the future.

### DIFF
--- a/src/lando/main/tests/test_models.py
+++ b/src/lando/main/tests/test_models.py
@@ -80,17 +80,18 @@ def test__models__Repo__scm_not_calculated_when_preset(
 
 
 @pytest.mark.parametrize(
-    "path, expected_exception",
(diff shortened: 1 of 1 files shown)


shtrom:
> We previously overwrote the settings.REPO_ROOT variable in test__models__Repo__pulse_routing_key_validator

👍🏼 😬 👍🏼

shtrom (approved):

PR #1091, opened 2026-04-16
landing_checks: block commits with REPO- flag mismatch (bug 2032267)

shtrom:
We use the Lando `repo.name`, e.g., firefox-autoland. This won't match the name that the VCT hook uses, e.g., autoland. This means we will need to disable the equivalent HgMO check after this has landed.

 * landing_checks: add support for `repo_name` to `LandingChecks`
 * landing_checks: block commits with `REPO-` flag mismatch  (bug 2032267)
 * tests: add `ids` to `test_check_commit_message_invalid_message`
 * tests: DRY `HG_PATCH_AUTHOR_COMMIT_MESSAGE_TEMPLATE` in `landing_checks`
 * tests: port `repolocked` tests from v-c-t

### DIFF
--- a/src/lando/api/legacy/workers/automation_worker.py
+++ b/src/lando/api/legacy/workers/automation_worker.py
@@ -102,7 +102,7 @@ class AutomationWorker(Worker):
 
             if not self.skip_checks(job, new_commits) and repo.hooks_enabled:
                 patch_helpers = repo.scm.get_patch_helpers_for_commits(new_commits)
-                landing_checks = LandingChecks(job.requester_email)
+                landing_checks = LandingChecks(job.requester_email, repo.name)
                 try:
                     check_errors = landing_checks.run(repo.hooks, patch_helpers)
                 except Exception as exc:
--- a/src/lando/utils/landing_checks.py
+++ b/src/lando/utils/landing_checks.py
@@ -446,6 +446,7 @@ class PatchCollectionCheck(Check, ABC):
     """
 
     push_user_email: str | None = None
+    repo_name: str | None = None
 
     @abstractmethod
     def next_diff(self, patch_helper: PatchHelper):
@@ -672,6 +673,7 @@ class PatchCollectionAssessor:
 
     patch_helpers: Iterable[PatchHelper]
     push_user_email: str | None = None
+    repo_name: str | None = None
 
     def run_patch_collection_checks(
         self,
@@ -686,7 +688,10 @@ class PatchCollectionAssessor:
         """
         issues = []
 
-        checks = [check(self.push_user_email) for check in patch_collection_checks]
+        checks = [
+            check(self.push_user_email, self.repo_name)
+            for check in patch_collection_checks
+        ]
 
         for patch_helper in self.patch_helpers:
             # Pass the patch information into the push-wide check.
@@ -724,9 +729,11 @@ class LandingChecks:
     """Utility class to run landing checks (a.k.a. hooks) on a list of commits."""
 
     requester_email: str
+    repo_name: str
 
-    def __init__(self, requester_email: str):
+    def __init__(self, requester_email: str, repo_name: str):
         self.requester_email = requester_email
+        self.repo_name = repo_name
 
     def run(
         self,
@@ -754,7 +761,9 @@ class LandingChecks:
         stack_checks = [chk for chk in ALL_STACK_CHECKS if chk.name() in hook_names]
 
         assessor = PatchCollectionAssessor(
-            patches, push_user_email=self.requester_email
+            patches,
+            push_user_email=self.requester_email,
+            repo_name=self.repo_name,
         )
         return assessor.run_patch_collection_checks(
             patch_collection_checks=stack_checks, patch_checks=commit_checks
--- a/src/lando/utils/landing_checks.py
+++ b/src/lando/utils/landing_checks.py
@@ -20,6 +20,8 @@ from lando.api.legacy.commit_message import (
 )
 from lando.main.scm.helpers import PatchHelper
 
+REPO_FLAG_RE = re.compile(r"[\s.;]REPO-(?P<repo>[-a-zA-Z0-9]+)(?:\w|$)")
+
 
 def wrap_filenames(filenames: list[str]) -> str:
     """Convert a list of filenames to a string with names wrapped in backticks."""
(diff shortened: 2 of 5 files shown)


src/lando/utils/landing_checks.py
```diff
@@ -510,6 +513,13 @@ def next_diff(self, patch_helper: PatchHelper):
             )
             return
 
+        if match := REPO_FLAG_RE.search(firstline):
```
shtrom: Note: this is the simplest implementation, added to the existing `CommitMessagesCheck`. This means we won't be able to selectively disable only the `REPO-` check.

I think it should be sufficient for our purpose, though.

src/lando/utils/landing_checks.py
```diff
@@ -510,6 +513,13 @@ def next_diff(self, patch_helper: PatchHelper):
...
+            if match["repo"] != self.repo_name:
+                self.commit_message_issues.append(
+                    f"Commit locked to another repo than {self.repo_name}: {commit_message}"
+                )
+                return
```
shtrom, line 526: Question: By `return`ing early here, and in other checks below, we only give a single error at a time per commit message. Should we instead not `return` and present all errors at once?

cgsheeh: If we can detect and warn about more issues that sounds like a good idea. Since we return a single string as the result of each check, it becomes a question of how the string is displayed in the UI.

I'd say this is acceptable for now, and in the future we can improve the checks so something more dynamic is returned from each warning/blocker.

src/lando/utils/tests/test_landing_checks.py
```diff
@@ -215,7 +215,23 @@ def test_check_commit_message_valid_message(commit_message: str, error_message:
             ),
             "`git-format-patch` cruft should result in a failed check.",
         ),
+        (
```
cgsheeh, line 218: Could we expand this to cover [more of the test cases from the hg hook](https://hg-edge.mozilla.org/hgcustom/version-control-tools/file/f21e253da4211b0b470fbced977be9522087b2d8/hghooks/tests/test-prevent-repolocked.t)?

src/lando/utils/landing_checks.py
cgsheeh: Agreed, I think this is okay to move here. :)

cgsheeh (dismissed):

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -102,7 +102,7 @@ def run_job(self, job: AutomationJob) -> bool:
 
             if not self.skip_checks(job, new_commits) and repo.hooks_enabled:
                 patch_helpers = repo.scm.get_patch_helpers_for_commits(new_commits)
-                landing_checks = LandingChecks(job.requester_email)
+                landing_checks = LandingChecks(job.requester_email, repo.name)
```
shtrom, line 105: Using `repo.name`  here may be problematic, as the `repo.name` is the name in lando, e.g., `firefox-autoland`, which doesn't match the name that the VCT hook uses, e.g., `autoland`. This means either one or the other check will choke...

src/lando/utils/landing_checks.py
shtrom: We return nothing here. We just add one error message with the commit message, and give up. We could add all the errors regarding every commit message in one go.

Given the way we currently render the errors, though, this would be trading off cognitive load for repeated partial fixes.

src/lando/utils/tests/test_landing_checks.py
```diff
@@ -215,34 +215,199 @@ def test_check_commit_message_valid_message(commit_message: str, error_message:
...
+        # (
+        #     "Bug 1 - update file REPO-bar",
+        #     "Commit locked to a repo other than foo: ",
+        #     "Commits are disallowed when REPO-bar is in any commit in the push.",
+        # ),
+        # (
+        #     "Bug 1 - update file REPO-bar",
+        #     "",
+        #     "Commits locked to bar (with REPO-bar in the title) should be allowed on subdir/bar.",
+        # ),
```
shtrom, line 306: Test case from v-c-t not ported.
```suggestion
```

src/lando/utils/tests/test_landing_checks.py
```diff
@@ -215,34 +215,199 @@ def test_check_commit_message_valid_message(commit_message: str, error_message:
...
+        (
+            "Bug 1 - update file REPO-subdir/bar",
+            "Push contains commits intended to be locked to subdir/bar but the repo name is badly formatted. '/' is not allowed: ",
+            "Commits locked to subdir/bar (with REPO-subdir/bar in the title) should not be allowed on foo.",
+        ),
+        # (
+        #     "Bug 1 - update file REPO-subdir",
+        #     "",
+        #     "Commits locked to subdir (with REPO-subdir in the title) should not be allowed on subdir/bar.",
+        # ),
```
shtrom, line 321: No ported.
```suggestion
```

src/lando/utils/tests/test_landing_checks.py
```diff
@@ -215,34 +215,199 @@ def test_check_commit_message_valid_message(commit_message: str, error_message:
...
+        (
+            "Bug 1 - update file REPO-bar REPO-foo",
+            "",
+            "Commits locked to bar and foo should be allowed on foo.",
+        ),
+        # (
+        #     "Bug 1 - update file\n\nREPO-foo",
+        #     "",
+        #     "Commits locked to foo (with REPO-foo in the title) should be allowed on try.",
+        # ),
```
shtrom, line 336: Not ported.
```suggestion
```

src/lando/utils/tests/test_landing_checks.py
shtrom: Done. That caught a few issues!

I added them all in the parametrise test cases, but some of them weren't relevant (as matching on HgMO repo paths). I left them commented out for review, with a suggestion to remove before merging.

src/lando/utils/landing_checks.py
cgsheeh: Ah, I mis-read the code. In that case it does seem like we should let the hook continue looking for errors, but it's not a blocker.

cgsheeh (approved):
LGTM, please remove the commented out test cases before landing. :)

src/lando/utils/landing_checks.py
shtrom: Yep. I filed a separate bug https://bugzilla.mozilla.org/show_bug.cgi?id=2033352

src/lando/api/legacy/workers/automation_worker.py
shtrom: We'll disable the Hg-side hook https://bugzilla.mozilla.org/show_bug.cgi?id=2032267

PR #1092, opened 2026-04-16
landing_checks: add blocker for signed Git commits (bug 1996492)

shtrom:
Git cinnabar doesn't support signed commits. If we allow any signed commit to
land into a repo synced to HgMO with git-hg-sync, this will immediately create a
divergence which will be painful to resolve.

For backward compatibility with existing deployments, we set
`gpg.ssh.allowedSignersFile` in the Dockerfile. This allows SSH-signature
detection to be functional on pre-existing clones. New clones will have the
config explicitly set on creation.

 * scm.helpers: add PatchHelperMetadata
 * GitScm: populate signature in PatchHelperMetadata
 * tests: add support for SSH-signing commits in test fixtures
 * test: add tests for commit signature detection
 * PatchHelper: add get_commit_summary to return first line of commit
 * landing_checks: add blocker for signed Git commits (bug 1996492)
 * repo: add PreventSignedCommitsCheck to HooksChoices
 * scm: type-hint PatchHelper factory methods as Self


src/lando/main/scm/helpers.py
```diff
@@ -117,14 +135,36 @@ class PatchHelper(ABC):
     # - subject
     headers: dict[str, str]
 
+    metadata: PatchHelperMetadata
+
     @classmethod
     @abstractmethod
-    def from_string_io(cls, string_io: io.StringIO) -> "PatchHelper":
+    def from_string_io(cls, string_io: io.StringIO, **kwargs) -> "PatchHelper":
```
cgsheeh, line 142: I think this should be `-> typing.Self:`

src/lando/main/scm/helpers.py
```diff
@@ -330,19 +379,22 @@ class GitPatchHelper(PatchHelper):
 
     @classmethod
     @override
-    def from_string_io(cls, string_io: io.StringIO) -> "GitPatchHelper":
+    def from_string_io(cls, string_io: io.StringIO, **kwargs) -> "GitPatchHelper":
```
cgsheeh, line 382: `-> Self:`, same below.

src/lando/main/scm/helpers.py
```diff
@@ -108,6 +109,23 @@ def get_timestamp_from_hg_date_header(date_header: str) -> str:
...
+    ExpiredKey = "Y"
+    RevokedKey = "R"
+    Error = "E"
+    NoSignature = "N"
+
+
+@dataclass
+class PatchHelperMetadata:
+
+    signature: SignatureValidity | None = None
```
cgsheeh, line 126: Seems like this should just be a concrete `SignatureValidity` instead of `Optional`, and the `None` case becomes `NoSignature` instead.

src/lando/main/scm/helpers.py
```diff
@@ -108,6 +109,23 @@ def get_timestamp_from_hg_date_header(date_header: str) -> str:
     return date_header.split(" ")[0]
 
 
+class SignatureValidity(enum.Enum):
```
cgsheeh, line 112: AFAICT we don't actually surface the signature state anywhere in the check/warning. If we plan to use this for something in the future we can keep it, but otherwise I think we could simplify this down to a `True`/`False` "does the commit have a signature".

src/lando/utils/tests/test_landing_checks.py
```diff
@@ -243,6 +245,46 @@ def test_check_commit_message_invalid_message(
...
+@pytest.mark.parametrize(
+    "signature_validity,expected_allowed",
+    (
+        (SignatureValidity.Good, False),
+        (SignatureValidity.Bad, False),
+        (SignatureValidity.UnknownValidity, False),
+        (SignatureValidity.ExpiredSignature, False),
+        (SignatureValidity.ExpiredKey, False),
+        (SignatureValidity.RevokedKey, False),
+        (SignatureValidity.Error, False),
```
cgsheeh, line 257: If we reduce the check to yes/no "is there a signed commit" we remove all these unnecessary tests.

src/lando/utils/landing_checks.py
```diff
@@ -545,6 +545,37 @@ def result(self) -> str | None:
...
+            patch_helper.metadata.signature
+            and patch_helper.metadata.signature != SignatureValidity.NoSignature
+        ):
+            self.signed_commits.append(patch_helper.get_commit_summary())
+
+    def result(self) -> str | None:
+        if self.signed_commits:
+            return "Patch introduces one or more signed commits: " + ", ".join(
+                self.signed_commits
+            )
```
cgsheeh, line 576: We can use an f-string here.

src/lando/utils/landing_checks.py
```diff
@@ -545,6 +545,37 @@ def result(self) -> str | None:
...
+@dataclass
+class PreventSignedCommitsCheck(PatchCollectionCheck):
+    """Prevent patches from introducing signed commits."""
+
+    signed_commits: list[str] = field(default_factory=list)
+
+    @override
+    @classmethod
+    def name(cls) -> str:
+        return "PreventSignedCommitCheck"
```
cgsheeh, line 557: The name is `"PreventSignedCommitsCheck"` (plural on `Commits`) in the migration and enum, so this will fail to run in production.

src/lando/conftest.py
```diff
@@ -894,32 +894,69 @@ def hg_server(hg_test_bundle: pathlib.Path, tmpdir: os.PathLike, worker_id: str)
...
+        if signed:
+            subprocess.run(
+                ("git", "config", "gpg.format", "ssh"),
+                cwd=str(clone_path),
+                check=True,
+            )
+
+            # This is mandatory for SSH signature validation to work.
+            subprocess.run(
+                ("git", "config", "gpg.ssh.allowedSignersFile", "/dev/null"),
```
cgsheeh, line 951: We already do this in the Dockerfile, do we need to do it again here?

src/lando/main/scm/helpers.py
```diff
@@ -148,6 +188,14 @@ def set_header(self, name: bytes | str, value: str):
 
         self.headers[name.lower()] = value
 
+    def get_commit_summary(self) -> str:
```
cgsheeh, line 191: This should be `get_commit_title` to align with Phabricator and Github terminology.

src/lando/main/scm/helpers.py
```diff
@@ -148,6 +188,14 @@ def set_header(self, name: bytes | str, value: str):
...
+        """Returns the commit summary (first line)."""
+        desc = self.get_commit_description()
+        if desc:
+            return desc.split("\n")[0]
+        # This should not happen, but if it does `grep` will lead us here.
+        return "(empty commit description)"
```
cgsheeh, line 197: Should we throw here?

src/lando/main/scm/git.py
```diff
@@ -95,12 +95,16 @@ def clone(self, source: str):
...
         self._git_run("checkout", self.default_branch, cwd=self.path)
-        self._git_setup_user()
+        self._git_repo_config()
 
-    def _git_setup_user(self):
+    def _git_repo_config(self):
         """Configure the git user locally to repo_dir so as not to mess with the real user's configuration."""
         self._git_run("config", "user.name", LANDO_USER_NAME, cwd=self.path)
         self._git_run("config", "user.email", LANDO_USER_EMAIL, cwd=self.path)
+        # This is mandatory for SSH signature validation to work.
```
cgsheeh, line 104: We set this here, in the Dockerfile and in the tests. I'd prefer we reduce down to one location if possible.

cgsheeh (dismissed):

src/lando/conftest.py
shtrom: I think we should keep it (but not necessarily in the tests here):

While the Dockerfile does it at the deployment level (and, crucially, for existing repositories without having to manually update their config), this ensures than any new repository is explicitely configured as needed, regardless of deployment.

src/lando/main/scm/helpers.py
shtrom: I preferred to err on the side of returning something to the user, rather than an error page.

src/lando/main/scm/git.py
shtrom: Ah, my comment about this on the tests was really indented for here. We could remove it from the tests, and rely on the Dockerfile, or just let make sure the repository gets setup correctly by this line of code here. I'll try this last approach.

But, yes, the Dockerfile is there for backward-compat with existing clones.

shtrom: Removing the set on the test fixture, and relying on the `_git_repo_config` to set the repo correctly worked! 👌🏼

So I'd say:
* keep in `_git_repo_config` as it's the right place to have it
* remove from the test fixture and rely on the repo getting configured correctly
* keep in `Dockerfile` for backward compat with existing repos without needing a manual migration step

src/lando/utils/landing_checks.py
shtrom: Not sure it makes it more legible, but ok 😆

src/lando/main/scm/helpers.py
```diff
@@ -108,6 +110,12 @@ def get_timestamp_from_hg_date_header(date_header: str) -> str:
     return date_header.split(" ")[0]
 
 
+@dataclass
+class PatchHelperMetadata:
+
+    signature: bool = False
```
shtrom, line 116: Question: do we need this? We could just add the signature at the `PatchHelper` level. I felt like the metadata was more commit-specific, than just a textual diff/patch, but perhaps this is not relevant?

Thoughts?

src/lando/main/tests/test_scm_helpers.py
```diff
@@ -506,6 +506,75 @@ def test_scm_get_patch_helpers_for_commits(
...
+
+    if sign_new:
+        # Only new commits from the merge should be flagged.
+        assert any(
+            ph.metadata.signature for ph in patch_helpers
+        ), "The signed commit was not detected"
+    else:
+        assert not any(
+            ph.metadata.signature for ph in patch_helpers
+        ), "A previously signed commit was erroneously detected"
```
cgsheeh, line 576: Nit: this is technically the same as `any(ph.metadata.signature for ph in patch_helpers) == sign_new`

src/lando/utils/landing_checks.py
```diff
@@ -545,6 +545,34 @@ def result(self) -> str | None:
...
+    @classmethod
+    def name(cls) -> str:
+        return "PreventSignedCommitsCheck"
+
+    @override
+    @classmethod
+    def description(cls) -> str:
+        return "Prevent patches from introducing signed commits."
+
+    @override
```
cgsheeh, line 564: Nit: do we need to `@override` here? It doesn't appear we do this on other `next_diff` methods.

src/lando/main/scm/helpers.py
cgsheeh: It could go either way. The boundary between what Lando considers a patch and a commit is thin. :)

Do you think we will be adding any other commit-only metadata in the near future? For decisions like this my default is to go with the simpler solution, until we have a concrete need for extra abstraction.

src/lando/utils/landing_checks.py
cgsheeh: My instinct was that string concatenation is usually the wrong way to do string formatting. Looking again, I think this case was an exception. :)

cgsheeh (approved):

src/lando/main/scm/git.py
```diff
@@ -260,7 +264,40 @@ def get_patch(self, revision_id: str) -> str | None:
...
+        From git-log(1):
+
+             show  "G"  for  a  good (valid) signature, "B" for a bad signature, "U" for
+             a good signature with unknown validity, "X" for a good signature that has
+             expired, "Y" for a good signature made by an expired key, "R" for a good
+             signature made by a revoked key, "E" if the signature cannot be checked
+             (e.g. missing key) and "N" for no signature
+
+        To properly detect SSH signatures, even if invalid, the system must have
+        `gpg.ssh.allowedSignersFile` to an existing file, even if a dummy.
```
zzzeid, line 286:
```suggestion
        `gpg.ssh.allowedSignersFile` point to any existing file.
```

src/lando/main/tests/test_scm_helpers.py
```diff
@@ -506,6 +506,75 @@ def test_scm_get_patch_helpers_for_commits(
...
+
+        new_commits = scm.describe_local_changes()
+        assert len(new_commits) == 3, "Unexpected number of commits"
+
+        # Flatten the filter() Iterable, so we can count the number of elements.
+        patch_helpers = list(scm.get_patch_helpers_for_commits(new_commits))
+
+        assert (
+            len(patch_helpers) == 2
+        ), "Unexpected number of PatchHelpers: there shouldn't be one for the merge commit."
```
zzzeid, line 566: nit: the assert message is a little misleading given the check being performed, since in theory the length could be 0 or 3 or anything other than 2 and it could throw the same message.

zzzeid (approved):

src/lando/main/scm/helpers.py
shtrom: Can't think of any at this point. I'll remove.

shtrom: Actually, I'll leave it, because it's a nice dataclass, that gets populated by kwargs, which I think is better than the attributes we add and need to support manually in the PatchHelper itself.

src/lando/main/tests/test_scm_helpers.py
shtrom: Ah, good point. But the error message is slightly different.

src/lando/utils/landing_checks.py
shtrom: I think we'd need the `@override` everywhere. IIRC it's a noop, but it allows the LSP to warn if your method doesn't correctly implement the overidden method's interface.

We should probably add it to the others, too!

shtrom:
Squashed due to https://lando.moz.tools/landings/37963/

shtrom:
Squash rebase on main for https://lando.moz.tools/landings/37990/

PR #1093, opened 2026-04-16
api, main, ui: remove try handover (bug 2031011)

zzzeid:


src/lando/main/models/landing_job.py
```diff
@@ -48,6 +48,7 @@ class LandingJob(BaseJob):
 
     is_pull_request_job = models.BooleanField(default=False, blank=True)
 
+    # NOTE: The handover fields are legacy and are kept for reference only.
```
shtrom, line 51: nit: can we mark them as `@deprecated`?

src/lando/urls.py
```diff
@@ -121,11 +120,6 @@
         LandingJobPullRequestAPIView.as_view(),
         name="api-landing-job-pull-request",
     ),
-    path(
-        "api/pulls/<str:repo_name>/<int:pull_number>/try_jobs",
```
shtrom, line 125: nit: we could return a 410 GONE here. Though it's a very bespoke endpoint, so I doubt anyone would benefit.

shtrom (approved):
lgtm.

For completeness, I assume that keeping the `handover_repo` references in the admin and ETL are deliberate?

It makes sense to me as we keep the fields, and we don't want to remove them from the existing ETL data.

```
$ git grep -i handover
src/lando/main/admin.py:        "handover_repo__name",
src/lando/main/admin.py:        "handover_repo",
src/lando/main/migrations/0042_landingjob_handover_repo_landingjob_is_handed_over.py:            name="handover_repo",
src/lando/main/migrations/0042_landingjob_handover_repo_landingjob_is_handed_over.py:                related_name="handover_jobs",
src/lando/main/migrations/0043_repo_treeherder_name.py:        ("main", "0042_landingjob_handover_repo_landingjob_is_handed_over"),
src/lando/main/models/landing_job.py:    # NOTE: The handover fields are legacy and are kept for reference only.
src/lando/main/models/landing_job.py:    # Reference to the handover repo. A handover repo is a repo that a landing
src/lando/main/models/landing_job.py:    handover_repo = models.ForeignKey(
src/lando/main/models/landing_job.py:        related_name="handover_jobs",
src/lando/main/models/landing_job.py:                Q(target_repo__in=repositories) | Q(handover_repo__in=repositories)
src/lando/utils/management/commands/etl.py:        "handover_repo_id",
src/lando/utils/tests/test_etl.py:        result["handover_repo_id"] is None
src/lando/utils/tests/test_etl.py:    ), "`handover_repo_id` should exist and be `None`.
```

src/lando/main/models/landing_job.py
cgsheeh: What do we need the references for in the main DB? I think we should remove them from here but keep them in the ETL for reference (though just as an empty value for now).

cgsheeh (changes requested):

src/lando/main/models/landing_job.py
zzzeid: We can remove them, but I don't think they should be removed in this PR or immediately after deployment. We can treat them as deprecated, and remove them in a future PR once Lando is running with the new setup for a while.

The values will be accessible anyway via DB backups if needed. They fields are kept here for gradual deprecation which we generally do with these fields (e.g., in case we need to check the admin or historical jobs for a while after this is deprecated.)

I'll file a bug to remove these fields in the near future.

src/lando/urls.py
zzzeid: True, the endpoint is no longer used (was used by Lando, and by code-review-bot only during testing, so technically this endpoint was never externally used in production by things other than Lando). I think we could safely do a `404` here given these facts.

src/lando/main/models/landing_job.py
cgsheeh: Gradual removal with a bug pre-filed sounds good to me! :+1:

cgsheeh (approved):

src/lando/main/models/landing_job.py
zzzeid: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2036801.

zzzeid:
> For completeness, I assume that keeping the `handover_repo` references in the admin and ETL are deliberate?
> 
> It makes sense to me as we keep the fields, and we don't want to remove them from the existing ETL data.

Yes, deliberately kept in the admin and ETL until these fields are fully removed. I can add "DEPRECATED" to the help text in the admin, if that helps?

PR #1100, opened 2026-04-21
landing_jobs: expose revision info in Job JSON data (bug 2033359)

shtrom:
Exposing more revision information allows consumers of the Job status JSON endpoint (at `/landing_jobs/<id>`) to display it before the stack has fully landed.

This is particularly useful for TreeHerder, and would allow users to confirm earlier that links they have received for others are correct.

### DIFF
--- a/src/lando/api/tests/test_transplants.py
+++ b/src/lando/api/tests/test_transplants.py
@@ -874,9 +874,9 @@ def test_integrated_transplant_updated_diff_id_reflected_in_landed_phabricator_r
         result["status"] == JobStatus.SUBMITTED
     ), f"Invalid status in GET /landing_jobs/{job.id} response"
 
-    assert result[
-        "revisions"
-    ], f"Missing or empty revisions list in GET /landing_jobs/{job.id} response"
+    assert (
+        len(result["revisions"]) == 1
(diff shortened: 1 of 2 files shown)


cgsheeh (approved):

src/lando/api/tests/test_transplants.py
```diff
@@ -874,6 +874,25 @@ def test_integrated_transplant_updated_diff_id_reflected_in_landed_phabricator_r
...
+    assert (
+        result["revisions"][0]["author_email"] == db_revision.author_email
+    ), "Missing revision author email in GET /landing_jobs/{job.id} response"
+    assert (
+        result["revisions"][0]["author_name"] == db_revision.author_name
+    ), "Missing revision author name in GET /landing_jobs/{job.id} response"
+    assert (
+        result["revisions"][0]["commit_message"] == db_revision.commit_message
+    ), "Missing revision commitmessage in GET /landing_jobs/{job.id} response"
+
```
zzzeid: nit: I think the assert message is inaccurate here, since the assert is not checking if something is present or not, but rather that it matches the expected value -- e.g., "Mismatch in returned vs. actual value."


 Also, if you wanted to be a little more prudent, you could check the length of `result["revisions"]`.

zzzeid (approved):

PR #1101, opened 2026-04-22
api: fix missing repo name for LandingChecks (bug 2032267)

shtrom:


src/lando/api/views.py
```diff
@@ -73,7 +73,7 @@ def generate_warnings_and_blockers(
     # mind for the future.
     patch_helper = PullRequestPatchHelper(pull_request)
     author_email = pull_request.author[1]
-    landing_checks = LandingChecks(author_email)
+    landing_checks = LandingChecks(author_email, str(target_repo.name))
```
cgsheeh: Why do we need to wrap in a `str()` here? `target_repo.name` is a `CharField` which should be a `str` already?

cgsheeh (approved):

src/lando/api/views.py
zzzeid: +1 to this (as a nit).

zzzeid (approved):

src/lando/api/views.py
shtrom: LSP complained (: But yeah...

```suggestion
    landing_checks = LandingChecks(author_email, target_repo.name)
```

PR #1102, opened 2026-04-22
try_api: run landing checks on submission (bug 2033883)

shtrom:
* models: move `TRY_HOOKS` constant from `create_environment_repos` to `Repo`
 * models: disable NSPR and WPT checks in `TRY_HOOKS`
 * tests: set proper hooks on `mocked_repo_config_try` fixture
 * tests: add `diff_to_git_patch` fixture
 * try_api: run landing checks on submission (bug 2033883)

### DIFF
--- a/src/lando/conftest.py
+++ b/src/lando/conftest.py
@@ -36,6 +36,7 @@ from lando.main.models import (
 )
 from lando.main.models.landing_job import LandingJob, add_job_with_revisions
 from lando.main.models.profile import SCM_ALLOW_DIRECT_PUSH
+from lando.main.models.repo import TRY_HOOKS
 from lando.main.models.revision import Revision
 from lando.main.scm import SCMType
 from lando.main.scm.commit import CommitData
@@ -829,6 +830,7 @@ def mocked_repo_config_try(mock_repo_config):
         short_name="try",
         is_phabricator_repo=False,
         is_try=True,
+        hooks=TRY_HOOKS,
         force_push=True,
     )
 
--- a/src/lando/conftest.py
+++ b/src/lando/conftest.py
@@ -236,9 +236,8 @@ Subject: {commit_description}
 """.strip()  # noqa: W291, `git` adds a trailing whitespace after `--`.
 
 
-    {diff}
-    -- 
-    """.rstrip()).strip()  # noqa: W291, `git` adds a trailing whitespace after `--`.
+@pytest.fixture
+def diff_to_git_patch() -> Callable:
 
     def _diff_to_git_patch(
         diff: str,
--- a/src/lando/conftest.py
+++ b/src/lando/conftest.py
@@ -9,6 +9,7 @@ import uuid
 from collections.abc import Callable, Iterable
 from datetime import datetime, timezone
 from pathlib import Path
+from textwrap import dedent
 
 import pytest
 import requests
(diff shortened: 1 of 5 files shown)


src/lando/main/models/repo.py
```diff
@@ -446,3 +452,17 @@ def _user_has_direct_permission(self, user: User, permission: str):
...
+# Try repos require a slightly different set of hooks from the defaults for normal
+# repos. This needs to be below the definition of Repo, so we can use it via
+# get_default_hooks.
+TRY_HOOKS = list(
+    # Set difference takes precedence over set union.
+    set(get_default_hooks())
+    - {CommitMessagesCheck.name()}
+    - {PreventNSPRNSSCheck.name()}
+    - {TryTaskConfigCheck.name()}
+    - {WPTSyncCheck.name()}
```
shtrom, line 466: Note: I added (the removal of) the WPT and NSPR checks, as I don't think they should apply for Try. Please correct me if I'm wrong.

cgsheeh: That sounds correct to me.

AIUI we will need to update the `repo.hooks` entry via the admin panel to do this in prod. This list only updates the state in the initial creation.

src/lando/try_api/api.py
```diff
@@ -202,19 +195,45 @@ def patches(
...
+        patch_helpers.append(ph)
+
+    landing_checks = LandingChecks(request.user.email, repo.name)
+    errors = landing_checks.run(
+        repo.hooks,
+        patch_helpers,
+    )
+
+    if any(errors):
+        bulleted_errors = "\n  - ".join(errors)
```
cgsheeh: Nit: this could be `if errors:` if we're trying to check for an empty list. If we are using `any(errors)` because we may have a `[None, None]` etc, we would need to filter those out in the `.join()`.

src/lando/try_api/api.py
```diff
@@ -202,19 +195,45 @@ def patches(
...
+        raise BadRequestProblemException(
+            title="Errors found in pre-submission patch checks.",
+            detail=error_message,
+        )
+
+    try_job = LandingJob.objects.create(
+        target_repo=repo,
+        requester_email=request.user.email,
+        target_commit_hash=target_commit_hash,
+        status=JobStatus.CREATED,
```
cgsheeh: Nit: now that we're creating all the DB entries at the end of the request, we could use a `with transaction.atomic():` and set this to `SUBMITTED` right away. This way we avoid creating an orphan `LandingJob` if we fail before the `Revision` objects are created and linked, and the job is updated to `Submitted`.

src/lando/conftest.py
```diff
@@ -222,6 +224,33 @@ def _patch(number=0):
     return _patch
 
 
+@pytest.fixture
+def diff_to_git_patch() -> Callable:
+    GIT_PATCH_TEMPLATE = dedent("""
```
cgsheeh: Nit: personally I would define this outside the fixture to avoid the `dedent()`.

cgsheeh (approved):

src/lando/try_api/api.py
zzzeid: I agree with this, I think this is not a nit since it could be confusing or lead to unexpected results in the future. In the latter case (e.g., if there's `[None, None]`) this will raise an exception. It would make more sense for there not to be any `None` in `errors` since that breaks the paradigm (i.e., `errors` would become a non-empty list, implying there are errors).

zzzeid, line 219: What does "bulleted" mean in this context?

zzzeid (dismissed):

src/lando/main/models/repo.py
shtrom: We could add a migration, but this one is easy to do at any time.

Like right now.

Done for `try` and `try-comm-central` in prod.

src/lando/try_api/api.py
shtrom: No, that should be an empty list. Fixed.

shtrom: That's copied from old lando-api, but that's

- something
- like
- this

but with leading `-` rather than `*`, matching what old lando did.

This is in the plain-text error response from the API.

src/lando/try_api/api.py
```diff
@@ -202,35 +196,62 @@ def patches(
...
+                },
+            )
+            revisions.append(revision)
+
+        add_revisions_to_job(revisions, try_job)
 
-    try_job.status = JobStatus.SUBMITTED
-    try_job.save()
+        try_job.status = JobStatus.SUBMITTED
+        try_job.save()
```
cgsheeh: Nit: now that this code is in a transaction, you can set the status to `SUBMITTED` in the `LandingJob.objects.create()` call above, and remove the `save()`.

cgsheeh (approved):

src/lando/try_api/api.py
shtrom: Gah! Forgot 😅

PR #1106, opened 2026-04-23
remote_settings: compressor fixes (bug 2034880)

zzzeid:
- update storage to use custom class
- add db cache backend for compressor
- update settings to enable offline compression

NOTE: upon first deployment after this is merged, it's necessary to run:
```
lando createcachetable
```

### DIFF
--- a/src/lando/remote_settings.py
+++ b/src/lando/remote_settings.py
@@ -3,9 +3,19 @@ import os
 from lando.main.logging import MozLogFormatter
 from lando.settings import *  # noqa: F403
 
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    },
+    "db": {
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):

shtrom (approved):

zzzeid:
Re-requesting review since these updates have now been tested on latest DEV setup, and they appear to fix all unexpected issues.

src/lando/remote_settings.py
```diff
@@ -3,9 +3,21 @@
 from lando.main.logging import MozLogFormatter
 from lando.settings import *  # noqa: F403
 
+
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        "LOCATION": "unique-snowflake",
+    },
```
zzzeid, line 9: Note: this is the default currently, it is just not explicitly set as such in remote settings.

cgsheeh (approved):

PR #1107, opened 2026-04-24
ui: show recent jobs on homepage when logged in (bug 1652787)

shtrom:
This requires a tiny tweak to the TimeLine CSS to force `margin-bottom` to 0 in `-itemDetails` (otherwise it gets overriden by Bulma's `content`).

### DIFF
--- a/src/lando/ui/jinja2/home.html
+++ b/src/lando/ui/jinja2/home.html
@@ -1,7 +1,7 @@
 {% extends "partials/layout.html" %}
 {% block main %}
     <main class="container content">
-        <div id="home">
+        <section id="home">
             <h1>Welcome to Lando</h1>
             <p class="bordered">
                 Lando is an automated code lander for Mozilla.
@@ -22,6 +22,28 @@
                 For more information on Lando, consult the <a href="http://moz-conduit.readthedocs.io/en/latest/lando-user.html">user
                 guide</a>.
             </p>
-        </div>
+        </section>
+        {% if user_is_authenticated and (pending_jobs|default(False) or final_jobs|default(False)) %}
+            <section class="StackPage">
+                <h1>Your recent jobs</h1>
+                {% set embedded = True %}
+                {% if pending_jobs %}
+                    <h2>Pending</h2>
+                    <div class="StackPage-landingQueue StackPage-timeline">
+                        {% for job in pending_jobs %}
+                            {% include "partials/job.html" %}
+                        {% endfor %}
+                    </div>
+                {% endif %}
+                {% if final_jobs %}
+                    <h2>Finished (last 7 days)</h2>
+                    <div class="StackPage-landingQueue StackPage-timeline">
+                        {% for job in final_jobs %}
+                            {% include "partials/job.html" %}
+                        {% endfor %}
+                    </div>
+                {% endif %}
+            </section>
+        {% endif %}
     </main>
 {% endblock %}
--- a/src/lando/ui/jinja2/home.html
+++ b/src/lando/ui/jinja2/home.html
@@ -27,6 +27,7 @@
             <section class="StackPage">
                 <h1>Your recent jobs</h1>
                 {% set embedded = True %}
+                <h2>Pending</h2>
                 {% if pending_jobs %}
                     <h2>Pending</h2>
                     <div class="StackPage-landingQueue StackPage-timeline">
--- a/src/lando/ui/jinja2/home.html
+++ b/src/lando/ui/jinja2/home.html
@@ -37,6 +37,7 @@
                 {% else %}
                     <p>You don't have any pending job currently.</p>
                 {% endif %}
+                <h2>Finished (last {{ MAX_JOBS_HISTORY }})</h2>
                 {% if final_jobs %}
                     <h2>Finished (last 7 days)</h2>
                     <div class="StackPage-landingQueue StackPage-timeline">
--- a/src/lando/ui/jinja2/home.html
+++ b/src/lando/ui/jinja2/home.html
@@ -23,7 +23,7 @@
                 guide</a>.
             </p>
         </section>
-        {% if user_is_authenticated and (pending_jobs|default(False) or final_jobs|default(False)) %}
+        {% if user_is_authenticated and (pending_jobs|default([]) or final_jobs|default([])) %}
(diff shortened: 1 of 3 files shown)


src/lando/ui/jinja2/home.html
```diff
@@ -22,6 +22,28 @@ <h1>Welcome to Lando</h1>
...
+                {% if pending_jobs %}
+                    <h2>Pending</h2>
+                    <div class="StackPage-landingQueue StackPage-timeline">
+                        {% for job in pending_jobs %}
+                            {% include "partials/job.html" %}
+                        {% endfor %}
+                    </div>
+                {% endif %}
+                {% if final_jobs %}
+                    <h2>Finished (last 7 days)</h2>
```
cgsheeh: We should use the `FINAL_JOBS_MAX_DAYS` constant instead of hard-coding `7` here, to avoid confusion if we change the maximum.

src/lando/ui/jinja2/home.html
```diff
@@ -22,6 +22,28 @@ <h1>Welcome to Lando</h1>
                 For more information on Lando, consult the <a href="http://moz-conduit.readthedocs.io/en/latest/lando-user.html">user
                 guide</a>.
             </p>
-        </div>
+        </section>
+        {% if user_is_authenticated and (pending_jobs|default(False) or final_jobs|default(False)) %}
+            <section class="StackPage">
+                <h1>Your recent jobs</h1>
+                {% set embedded = True %}
```
cgsheeh: Ideally we would do `{% with embedded = True %}...{% endwith %}` to avoid this variable leaking out and causing weird rendering elsewhere.

As a larger change, we should really put our template partials into macros instead, so we can render elements like method calls.

cgsheeh (approved):
Ship it!

src/lando/ui/jinja2/home.html
zzzeid: Might make more sense to show last N jobs here, instead of base it on time, at least for this quick-and-dirty version. This way would also make the UI a little more consistent. Likely in the future this would go in an async table that has some dynamic filters the user would tweak.

src/lando/ui/jinja2/home.html
```diff
@@ -22,6 +22,28 @@ <h1>Welcome to Lando</h1>
                 For more information on Lando, consult the <a href="http://moz-conduit.readthedocs.io/en/latest/lando-user.html">user
                 guide</a>.
             </p>
-        </div>
+        </section>
+        {% if user_is_authenticated and (pending_jobs|default(False) or final_jobs|default(False)) %}
```
zzzeid: I think `pending_jobs` and `final_jobs` should not default to a boolean here, since their name implies that they're an iterable. The context variables *should* default to a "falsey" value if the querset is empty. We can also convert the queryset into a list for the purposes of the context variable.

```suggestion
        {% if user_is_authenticated and (pending_jobs or final_jobs) %}
```

src/lando/ui/jinja2/home.html
```diff
@@ -22,6 +22,28 @@ <h1>Welcome to Lando</h1>
...
+            <section class="StackPage">
+                <h1>Your recent jobs</h1>
+                {% set embedded = True %}
+                {% if pending_jobs %}
```
zzzeid: I don't think we should hide the section if there are no jobs. Instead, we should show the heading with something that says "You do not have any submitted jobs.". Also this section should probably be named "active" or something a little more encompassing of deferred and in-progress jobs.

src/lando/ui/jinja2/home.html
```diff
@@ -22,6 +22,28 @@ <h1>Welcome to Lando</h1>
...
+                    <h2>Pending</h2>
+                    <div class="StackPage-landingQueue StackPage-timeline">
+                        {% for job in pending_jobs %}
+                            {% include "partials/job.html" %}
+                        {% endfor %}
+                    </div>
+                {% endif %}
+                {% if final_jobs %}
```
zzzeid: Same here. Show the section with a default message, instead of hiding it, if there are no jobs.

zzzeid (dismissed):

src/lando/ui/jinja2/home.html
shtrom: > As a larger change, we should really put our template partials into macros instead, so we can render elements like method calls.

Very much agree with this!

src/lando/ui/jinja2/home.html
shtrom: The default is just in that particular test, in case either variable is _completely_ missing from the context (which would lead to a 500). We can make the default `[]`, no worries, but it makes no functional difference.

src/lando/ui/jinja2/home.html
shtrom: > Ideally we would do `{% with embedded = True %}...{% endwith %}` to avoid this variable leaking out and causing weird rendering elsewhere.

TIL.

src/lando/ui/legacy/pages.py
```diff
@@ -12,19 +11,17 @@
...
             context["pending_jobs"] = LandingJob.objects.filter(
                 requester_email=request.user.email, status__in=JobStatus.pending()
             )
             context["final_jobs"] = LandingJob.objects.filter(
                 requester_email=request.user.email,
                 status__in=JobStatus.final(),
-                updated_at__gt=datetime.datetime.now()
-                - datetime.timedelta(days=self.FINAL_JOBS_MAX_DAYS),
-            )
+            ).order_by("-updated_at")[: self.MAX_JOBS_HISTORY]
```
zzzeid, line 25: nit: this could result in jobs jumping around in the history. Might be better to fix the sorting to `created_at` to avoid unexpected/confusing jumping around.

zzzeid (approved):

src/lando/ui/legacy/pages.py
shtrom: In a way, I think it's better if you get the most-recent jobs first, rather than having to dig down when looking for old jobs that completed recently.

The trade-off is if we go in the admin UI and change jobs, they will pop back up in the user's queue. 

I'm tempted to leave it as is for now, but we can see how users feel about it, and how often we change jobs and create confusion, and change to created if that's too much.

PR #1108, opened 2026-04-24
ui: don't redirect pages for jobs without revisions to canonical url (bug 2034715)

shtrom:


src/lando/ui/jobs.py
```diff
@@ -66,9 +66,13 @@ def get(
...
-            not revision_id
-            or (not landing_job.revisions.filter(revision_id=revision_id))
-        ) and (revision_id := landing_job.revisions[0].revision_id):
+            (
+                not revision_id
+                or (not landing_job.revisions.filter(revision_id=revision_id))
+            )
+            and landing_job.revisions.exists()
+            and (revision_id := landing_job.revisions[0].revision_id)
+        ):
```
cgsheeh: I think this would be easier to parse in a separate helper method, it's hard to parse in this multi-line conditional.

cgsheeh (changes requested):

src/lando/ui/jobs.py
zzzeid: I think either a separate helper method, or re-writing this so it's easier to read. It's a little messy as it is now.

shtrom: Yeah, it did require some better thinking to make it more legible 👍🏼 😅

cgsheeh (approved):

PR #1123, opened 2026-05-04
scm: send patch content to a file instead of `stdout` (Bug 2036740)

cgsheeh:


zzzeid (commented):
Curious what problems this has already presented in the logs? Seems like something we can handle when parsing logs or via the logs UI / query. We could be losing some info here, and setting an arbitrary value for length doesn't seem like an improvement.

cgsheeh:
> Curious what problems this has already presented in the logs? Seems like something we can handle when parsing logs or via the logs UI / query. We could be losing some info here, and setting an arbitrary value for length doesn't seem like an improvement.

@zzzeid we're logging the entire patch content for each commit when running `hg export`. This is making the logs excessively long for larger patches/stacks, and makes logs hard to parse in ArgoCD view (which while imperfect is still convenient and useful). There's no reason we need to view this content in the logs, and if we need to debug we can easily reproduce the command in the container shell.

Alternatively we could add a flag to control the behaviour. We could either enable/disable the truncation, or enable/disable logging the command output altogether where appropriate. I figured this approach was a reasonable middle ground for now.

zzzeid:
> > Curious what problems this has already presented in the logs? Seems like something we can handle when parsing logs or via the logs UI / query. We could be losing some info here, and setting an arbitrary value for length doesn't seem like an improvement.
> 
> @zzzeid we're logging the entire patch content for each commit when running `hg export`. This is making the logs excessively long for larger patches/stacks, and makes logs hard to parse in ArgoCD view (which while imperfect is still convenient and useful). There's no reason we need to view this content in the logs, and if we need to debug we can easily reproduce the command in the container shell.
> 
> Alternatively we could add a flag to control the behaviour. We could either enable/disable the truncation, or enable/disable logging the command output altogether where appropriate. I figured this approach was a reasonable middle ground for now.

It sounds like `hg export` is the main command that is causing this excessive output? Is there a way to change the logging behaviour for that particular command (e.g., reduce verbosity or change the output via flags), instead of applying this blanket truncation to everything? As a last resort, maybe we can implement a filter for that particular command's output to exclude patch content.

Lastly if there's an error with that command, we should probably still log the output as is since even the patch content may reveal something important.

src/lando/main/scm/git.py
zzzeid, line 1: In your original description you mentioned this issue is with the `hg export` command. Do we need this to apply to git commands as well?

zzzeid (changes requested):
Having looked at this implementation, it would actually be more suitable as something that we define in the logging framework. Maybe adding a filter or special formatter in `main.logging` and making it applicable only to certain modules/commands that way, and only enabling it for say remote logging (or even prod) in settings.

This would avoid making any changes to `scm.git` or `scm.hg`.

src/lando/main/scm/git.py
cgsheeh: Yes, `git diff` content doesn't need to be emitted in full to the logs, hence why I added `truncate_log_output` below.

cgsheeh:
> Having looked at this implementation, it would actually be more suitable as something that we define in the logging framework. Maybe adding a filter or special formatter in `main.logging` and making it applicable only to certain modules/commands that way, and only enabling it for say remote logging (or even prod) in settings.
> 
> This would avoid making any changes to `scm.git` or `scm.hg`.

How would you suggest we define this in the logging framework? We need to truncate the log output on a per-callsite basis, not a per-module or app-wide basis. We would either need to match commands by string in the log message (brittle), or have the call site add flags to opt them into the truncated logging (via `extra=` or otherwise), at which point we're still updating `{git,hg}.py` anyways.

src/lando/main/scm/git.py
zzzeid: Hmm, for both the git cases, I think we could simply remove `--stdout` or use `--output` to prevent the unnecessary output in the first place (which may be a legacy implementation anyway). I wonder if this is something we can also do in `hg` commands? I think if we can avoid truncating output it would be best.

zzzeid:
> How would you suggest we define this in the logging framework? We need to truncate the log output on a per-callsite basis, not a per-module or app-wide basis. We would either need to match commands by string in the log message (brittle), or have the call site add flags to opt them into the truncated logging (via `extra=` or otherwise), at which point we're still updating `{git,hg}.py` anyways.

I guess with our existing implementation there wouldn't be a way to do it without adding _something_ to the `scm` modules (e.g., to be able to use `funcName` below, maybe by adding an extra parameter), but the actual truncation would be better placed in the formatter IMO, that way you can easily disable it if needed. You would use something like `module` and `funcName` somewhere in the formatter to do this. E.g.`,

```python
if record.module == "hg" and record.funcName == "_hg_run":
    ...
```

But if possible we should solve this problem by not using stdout in the first place (as I mentioned in the other comment).

zzzeid (commented):
Oops my other comment did not get submitted, here it is.

src/lando/main/scm/git.py
shtrom: In a way, if the output somehow ends up being incorrect, it's probably best to have _some_ of it in the logs, for quicker identification, than none at all. Truncating would allow to retain that.

shtrom: That said, yes, not using `--stdout` may be a good idea (likely with `--output-directory` and `--numbered-files` to get predictable names).

cgsheeh:
Updated to instead send patch output to a temp file and read from there. Errors should still surface via `stderr` on command failure.

src/lando/main/scm/git.py
```diff
@@ -237,26 +237,46 @@ def add_diff_from_patches(self, patches: str) -> str:
...
+                    cwd=self.path,
+                )
+                diff_bytes = Path(diff_file.name).read_bytes()
+
+        try:
+            diff = diff_bytes.decode("utf-8")
+        except UnicodeDecodeError:
+            diff = diff_bytes.decode("latin-1")
+
+        return diff.lstrip()
```
zzzeid, line 257: We're using `lstrip` here, but `strip` in `get_patch`. Any particular reason for the disparity? Maybe add a comment as to why.

If they both can use the same flow, we could factor out the try/except and strip into a method since it's repeated below.

src/lando/main/scm/git.py
```diff
@@ -237,26 +237,46 @@ def add_diff_from_patches(self, patches: str) -> str:
...
 
     @override
     def get_patch(self, revision_id: str) -> str | None:
-        """Return a complete patch for the given revision, in the git extended diff format.
+        """Return a complete patch for the given revision, in the git extended diff format."""
+        # Write to a temp file instead of `--stdout` so the patch stays out of command logs.
+        with tempfile.NamedTemporaryFile(suffix=".patch") as patch_file:
```
zzzeid, line 263: I think it would be a good idea to log the name of the temporary file, in case we need to dig around to do some debugging.

Edit: maybe that's already logged in the command output, so disregard if so.

zzzeid (approved):
lgtm! Much cleaner approach.

Couple of nits/questions.

src/lando/main/scm/git.py
cgsheeh: Yes, it's logged in command output. However the file is also removed when exiting the context manager, so it's a bit of a moot point anyway. :)

src/lando/main/scm/git.py
cgsheeh: This is because `_git_run` used `rstrip=False` here, so we would run `lstrip()` but not `rstrip()`. In `get_patch` we have `rstrip=True`, so we would do both `lstrip()` and `rstrip()`, which is the same as `strip()`.

shtrom (approved):

src/lando/main/scm/git.py
zzzeid: Ah yeah, we could set `delete=False` and `delete_on_close=False` if we wanted to retain those for debugging.

PR #1134, opened 2026-05-05
format: replaced black with ruff for code formatting (Bug 2036413)

Magnolia-Liu:
Remove black references from format, tests and documentation.


Makefile
```diff
@@ -21,7 +21,7 @@ help:
 	@echo
 	@echo "target is one of:"
 	@echo "    help                 show this message and exit"
-	@echo "    format               run ruff and black on source code"
+	@echo "    format               run ruff on source code"
```
cgsheeh: Technically, we also run `djLint` with this command. It would be good to update that now. :)

cgsheeh (changes requested):

zzzeid (changes requested):
Couple of questions: is there any way to tweak the setting so that migration files are not affected? This would be ideal for two reasons (1) it would be good to avoid editing old migration files in general and (2) every new migration file will run into this problem, since these files are generated by Django (those are the files under `migrations` subdirectories).

ruff.toml
```diff
@@ -19,6 +19,8 @@ exclude = [
     "src/lando/version/__init__.py",
 ]
 
+extend-exclude = ["src/lando/version/__init__.py", "src/lando/*/migrations/*"]
```
shtrom, line 22: nit: Do we need this both in the `ruff.toml` and `pyproject.toml`? Ideally, we should have as much as the config in `pyproject.toml`, and hopefully none in `ruff.toml`.

OTOH, if that doesn't work, I think it would be preferable to have the things in `ruff.toml`, and no duplication.

Magnolia-Liu: Keeping the exclude in only pyproject.toml didn't work so instead I'm keeping it in ruff.toml

cgsheeh (approved):

zzzeid (approved):

pyproject.toml
shtrom, line 90: Nit: it may be worth leaving a comment here to remind us that `ruff` is configured in `ruff.toml`.

```suggestion
extend_exclude = "src/lando/static_src/legacy/vendor/.*|htmlcov/.*"

# Ruff settings are in ruff.toml.
```

shtrom (approved):

shtrom:
It may be worth running one last `make format` after a last merge of main just before landing, to make sure everything is as it should be moving forwards.

cgsheeh:
> It may be worth running one last `make format` after a last merge of main just before landing, to make sure everything is as it should be moving forwards.

`test_formatting.py` runs the formatter, so as long as it passes we are good to go. :)

cgsheeh (approved):

PR #1135, opened 2026-05-05
worker: run repo maintenance during idle time (Bug 2037216)

cgsheeh:
Lando's workers typically run repo-cleaning commands at
the beginning of each job processing time. In hg workers,
we run `hg strip` at the start of each job to remove
previously-created stale commits, despite those commits
not interfering with the job completion. In Git, we take
the opposite approach and simply ignore the temporary
work branch - it is never cleaned up.

Add a new "repo maintenance" step that runs while the
worker is being throttled due to no jobs remaining in
the queue. To avoid running excessive maintenance, the
runtime of the last maintenance run for each repo is
recorded in the worker, and maintenance is skipped if
it has been completed within the threshold.

For Mercurial, move the `hg strip` command into this maintenance
task, which should save us about 8s for each push to try.
For Git, add a cleanup of the stale working branches,
so we no longer have thousands of temp branches in our
worker repos.

After this change, each `HgSCM.clean_repo` call sites
always pass `strip_non_public_commits=False`, while
`GitSCM.clean_repo` call sites always pass `True`.
Remove the kwarg and make each behaviour the default.


src/lando/main/scm/git.py
```diff
@@ -488,18 +488,38 @@ def update_repo(
...
+        ).splitlines()
+        if not branches:
+            return
+
+        # `git branch -D` refuses to delete the currently checked-out branch,
+        # so move off any `lando-*` branch first.
+        if self.get_current_branch().startswith("lando-"):
+            self._git_run("checkout", "--force", self.default_branch, cwd=self.path)
+
+        self._git_run("branch", "-D", *branches, cwd=self.path)
```
zzzeid: This should go in its own method that is called within`maintenance`, so that we can add other maintenance methods as needed, if needed.

src/lando/main/scm/git.py
```diff
@@ -488,18 +488,38 @@ def update_repo(
...
         *,
-        strip_non_public_commits: bool = True,
         attributes_override: str | None = None,
     ):
         """Reset the local repository to the origin"""
-        if strip_non_public_commits:
-            self._git_run(
-                "reset", "--hard", f"origin/{self.default_branch}", cwd=self.path
-            )
+        self._git_run("reset", "--hard", f"origin/{self.default_branch}", cwd=self.path)
```
zzzeid, line 553: This change seems outside the scope of this PR.

src/lando/main/scm/hg.py
```diff
@@ -539,7 +539,8 @@ def update_repo(
         if not target_cset:
             target_cset = self._get_remote_head(source)
 
-        # Strip any lingering changes.
+        # Clean working directory; idle-time `maintenance()` is responsible
+        # for stripping draft commits, so we don't pay that cost here.
```
zzzeid: I would elaborate a little more perhaps on what is being done here, but at this point it's irrelevant what another method is doing.


```suggestion
        # Clean working directory.
```

src/lando/main/scm/hg.py
```diff
@@ -748,12 +749,12 @@ def read_rejects_files(self) -> dict[str, str]:
...
         self,
         *,
-        strip_non_public_commits: bool = True,
         attributes_override: str | None = None,
     ):
         """Clean the local working copy from all extraneous files.
 
-        `attributes_override` is ignored.
+        `attributes_override` is ignored. Stripping draft commits happens in
+        `maintenance()` during worker idle time, not here.
```
zzzeid: Same here, it's irrelevant what happens elsewhere.

If attributes_overrides is no longer used, we should maybe remove it or raise an error if a value is passed.

src/lando/main/scm/hg.py
```diff
@@ -768,12 +769,26 @@ def clean_repo(
...
+        except hglib.error.ServerError as exc:
+            raise SCMException(
+                "Failed to open hg server for maintenance.", "", str(exc)
+            ) from exc
+        try:
+            self.run_hg(["strip", "--no-backup", "-r", "not public()"])
+        except HgException:
+            pass
+        finally:
+            self.hg_repo.close()
```
zzzeid, line 795: Same comment here, this should go in a separate method that is called within `maintenance`.

src/lando/api/legacy/workers/base.py
```diff
@@ -77,6 +78,9 @@ def bootstrap_repos(self):
 
     last_job_finished: bool | None = None
 
+    # Minimum seconds between idle-time `scm.maintenance()` runs per repo.
```
zzzeid: Nit (here and elsewhere similar)

```suggestion
    # Minimum seconds between `scm.maintenance` calls per repo when idle.
```

src/lando/api/legacy/workers/base.py
```diff
@@ -284,6 +291,40 @@ def refresh_active_repos(self):
         ]
         logger.info(f"{len(self.active_repos)} enabled repos: {self.active_repos}")
 
+    def run_idle_maintenance(self):
+        """Run `scm.maintenance()` on each enabled repo, throttled per repo.
```
zzzeid: nit

```suggestion
        """Call `scm.maintenance` on each enabled repo, throttled per repo.
```

("Run" or call implies the `()`)

src/lando/api/legacy/workers/base.py
```diff
@@ -77,6 +78,9 @@ def bootstrap_repos(self):
 
     last_job_finished: bool | None = None
 
+    # Minimum seconds between idle-time `scm.maintenance()` runs per repo.
+    MAINTENANCE_INTERVAL_SECONDS = 300
```
zzzeid: This should go in a configuration variable like other duration variables.

src/lando/api/legacy/workers/base.py
```diff
@@ -284,6 +291,40 @@ def refresh_active_repos(self):
         ]
         logger.info(f"{len(self.active_repos)} enabled repos: {self.active_repos}")
 
+    def run_idle_maintenance(self):
+        """Run `scm.maintenance()` on each enabled repo, throttled per repo.
+
+        Called when no job is available. Each repo is maintained at most once
+        per `MAINTENANCE_INTERVAL_SECONDS` so this branch stays cheap on
+        consecutive idle loops.
```
zzzeid: I think "this branch" is ambiguous here. Maybe reword it to avoid confusing/mixing up a repo branch with a "code path".

zzzeid (changes requested):
Few small comments (and note failing tests).

src/lando/main/scm/git.py
cgsheeh: This is explained in the commit message:

> After this change, each `HgSCM.clean_repo` call sites
> always pass `strip_non_public_commits=False`, while
> `GitSCM.clean_repo` call sites always pass `True`.
> Remove the kwarg and make each behaviour the default.

src/lando/main/scm/git.py
cgsheeh: I'd rather hold off on this until we have a second maintenance task that justifies the split. Extracting the body into a helper would leave `maintenance()` as a one-line wrapper, which isn't really useful. We're assuming we will know what the next maintenance step might look like, and if we're wrong it just creates more work for the next person who adds it.

src/lando/main/scm/hg.py
cgsheeh: See my comment on the Git version of `maintenance`.

src/lando/api/tests/test_hg.py
```diff
@@ -43,37 +43,32 @@ def test_integrated_hgrepo_clean_repo(hg_clone):
...
+        # Dirty the directory again before exiting the context manager.
+        new_file.write("extra data", mode="a")
         assert repo.run_hg_cmds([["status"]])
 
-        # Cleaning should remove commit and clean working directory.
-        repo.clean_repo()
-        with pytest.raises(HgCommandError, match="no changes found"):
-            repo.run_hg_cmds([["outgoing"]])
-        assert not repo.run_hg_cmds([["status"]])
+    with repo.for_pull(), hg_clone.as_cwd():
```
shtrom: What is the purpose of the `hg_clone.as_cwd()` here?

src/lando/api/tests/test_hg.py
```diff
@@ -31,7 +31,7 @@ def test_integrated_hgrepo_clean_repo(hg_clone):
     repo = HgSCM(hg_clone.strpath)
```
shtrom: nit: Could we rename this variable to `scm` while we're at it?

src/lando/api/tests/test_worker.py
```diff
@@ -29,3 +30,73 @@ def test_Worker__no_SSH_PRIVATE_KEY(
 
     # It should complain, but continue.
     assert LandingWorker.SSH_PRIVATE_KEY_ENV_KEY in caplog.text
+
+
+@pytest.fixture
+def mocked_enabled_repos(hg_landing_worker):
```
shtrom, line 37: Could we use the git_landing_worker instead? This may exist for longer than the hg one.

src/lando/api/tests/test_worker.py
```diff
@@ -29,3 +30,73 @@ def test_Worker__no_SSH_PRIVATE_KEY(
...
+def test_Worker_run_idle_maintenance_runs_again_after_interval(
+    hg_landing_worker, mocked_enabled_repos
+):
+    hg_landing_worker.run_idle_maintenance()
+
+    # Pretend the previous run happened beyond the throttle window.
+    interval = hg_landing_worker.worker_instance.maintenance_interval_seconds
+    for repo in mocked_enabled_repos:
+        hg_landing_worker.last_maintenance_at[repo.id] -= interval + 1
+    hg_landing_worker.run_idle_maintenance()
```
shtrom, line 89: nit


```suggestion

    hg_landing_worker.run_idle_maintenance()
```

src/lando/api/tests/test_worker.py
```diff
@@ -29,3 +30,73 @@ def test_Worker__no_SSH_PRIVATE_KEY(
...
+
+
+@pytest.mark.django_db
+def test_Worker_run_idle_maintenance_isolates_failures(
+    caplog, hg_landing_worker, mocked_enabled_repos
+):
+    assert len(mocked_enabled_repos) >= 2, "Test requires at least two enabled repos."
+
+    failing_repo, *healthy_repos = mocked_enabled_repos
+    failing_repo._scm.maintenance.side_effect = SCMException("boom", "", "")
```
shtrom, line 138: 💥

shtrom (approved):

src/lando/main/scm/git.py
```diff
@@ -488,18 +488,38 @@ def update_repo(
...
+            "for-each-ref",
+            "--format=%(refname:short)",
+            "refs/heads/lando-*",
+            cwd=self.path,
+        ).splitlines()
+        if not branches:
+            return
+
+        # `git branch -D` refuses to delete the currently checked-out branch,
+        # so move off any `lando-*` branch first.
```
zzzeid: Interesting bit here. In the future it might make more sense to ensure we are back on the default branch after a job is finished.

src/lando/main/scm/git.py
```diff
@@ -488,18 +488,38 @@ def update_repo(
...
+        if self.get_current_branch().startswith("lando-"):
```
zzzeid: nit: with our current setup, this is a little redundant. Seems that deterministically, this if statement will always be `True` with the exception of perhaps first start.

nit: `self.get_current_branch().startswith("lando-")` (or even `"lando-"`) should probably be in their own property/method/variable. E.g., `self.is_on_job_branch` or something similar.

src/lando/main/scm/git.py
zzzeid: Could hold off, but at that point I would suggest either modifying the docstring (to be more centered around what the maintenance concept is about, and adding more detail in the summary), or renaming the method to `delete_job_branches`.

> would leave maintenance() as a one-line wrapper

We do that elsewhere, nothing wrong with that (see `run_code_formatters`).

src/lando/main/scm/hg.py
zzzeid: See my reply as well in that thread.

src/lando/api/legacy/workers/base.py
```diff
@@ -284,6 +288,40 @@ def refresh_active_repos(self):
...
         logger.info(f"{len(self.active_repos)} enabled repos: {self.active_repos}")
 
+    def run_idle_maintenance(self):
+        """Call `scm.maintenance` on each enabled repo, throttled per repo.
+
+        Called when no job is available. Each repo is maintained at most once
+        per `worker_instance.maintenance_interval_seconds` to avoid unnecessary
+        cleanup.
+        """
+        now = monotonic()
```
zzzeid: Interesting use here of `monotonic` (we don't use that anywhere else in the repo). The convention is to use `datetime.now`. I'd suggest sticking with the convention, or updating other instances to use monotonic so that we don't have multiple ways of doing the same thing.

zzzeid (approved):
Few additional comments but looks good otherwise.

src/lando/api/tests/test_hg.py
cgsheeh: `repo = HgSCM()` is the convention in this file, oddly enough. I updated this instance, but we should fix the others in a follow-up. :)

src/lando/api/legacy/workers/base.py
cgsheeh: Good point. Technically for this purpose `monotonic` is more correct, but it's not worth breaking the convention over. :)

src/lando/api/tests/test_worker.py
cgsheeh: I tried switching it over but hit test failures, filed https://bugzilla.mozilla.org/show_bug.cgi?id=2037553 to track the underlying issue.

src/lando/api/tests/test_hg.py
shtrom: It pre-dates the SCM split (;

zzzeid (commented):
It occurred to me that there may be some unexpected behaviour (not mentioned here) as to what would happen (1) on first deploy (i.e., the first maintenance run) but also (2) when the landing worker is busy for a long time (which is typical) and things accumulate in such a way that the maintenance run may take longer than expected (or possibly take more resources than expected). Worth testing those scenarios before deploying.

src/lando/main/scm/git.py
cgsheeh: > We do that elsewhere, nothing wrong with that (see `run_code_formatters`).

This is structurally different from `run_code_formatters`, as that method binds specific arguments to the generic `run_mach_command` helper that we use elsewhere in the workers. Splitting the branch deletion out of `maintenance` wouldn't do anything analogous - it would just forward from one no-arg method to another with no re-used code.

As an alternative proposal, I could rename both `maintenance` commands to `delete_work_branches` or similar, since that's the scope of both `maintenance` commands at the moment. Then we can expand the concept to a more general `maintenance` wrapper if we want to perform other tasks here.

shtrom: > As an alternative proposal, I could rename both maintenance commands to delete_work_branches or similar, since that's the scope of both maintenance commands at the moment. Then we can expand the concept to a more general maintenance wrapper if we want to perform other tasks here.

From the perspective of the outside callsites, I think it's probably best to keep “maintenance”, rather than expose lower-level logic of what the maintenance entails.

zzzeid: I think these two things would be satisfactory in my opinion:
> modifying the docstring (to be more centered around what the maintenance concept is about ...
```python
def maintenance (...):
    """
    Perform various maintenance tasks while the worker is idling.

    ... Currently this method cleans up work branches ...
    """
    # rest of the code here as it exists.
```

> This should go in its own method that is called within maintenance
```python
def maintenance(...):
    """Call various maintenance methods while the worker is idling."""
    _cleanup_work_branches(...)
```

src/lando/api/tests/test_hg.py
cgsheeh: I copied it from the other uses in this test, but I removed it from the newly added test code and it still passes.

src/lando/main/scm/git.py
cgsheeh: Good point, the check is redundant. I've removed it and we now move to the default branch for the repo unconditionally. :+1:

cgsheeh:
> It occurred to me that there may be some unexpected behaviour (not mentioned here) as to what would happen (1) on first deploy (i.e., the first maintenance run) but also (2) when the landing worker is busy for a long time (which is typical) and things accumulate in such a way that the maintenance run may take longer than expected (or possibly take more resources than expected). Worth testing those scenarios before deploying.

I'm going to throw this up on dev and test it out before merging/deploying. :+1: 

To handle 2), we should check that we haven't exceeded the throttle time between each repo maintenance call, so we only clean up a few repos at a time before looking for another landing job. We could also sort the repos by time since the last maintenance to decide which one to process, that way each repo is guaranteed to have maintenance run on it eventually, and we don't accidentally spend several minutes running maintenance tasks instead of processing landing jobs.

PR #1138, opened 2026-05-08
scm.git: add support for cloning private repo (bug 2037928)

zzzeid:
- add entry to clone test repo locally
- update scm.git to use authenticated URL
- move existing path authentication functionality to reusable method

### DIFF
--- a/src/lando/main/scm/git.py
+++ b/src/lando/main/scm/git.py
@@ -77,6 +77,12 @@ class GitSCM(AbstractSCM):
         self.default_branch = default_branch
         super().__init__(path)
 
+    def authenticate_path_if_possible(self, path: str) -> str:
+        """Return authenticated URL if supported."""
+        if GitHub.is_supported_url(path):
+            return GitHub(path).authenticated_url
+        return path
+
     @classmethod
     @override
     def scm_type(cls) -> SCMType:
@@ -90,10 +96,12 @@ class GitSCM(AbstractSCM):
         return "Git"
 
     @override
-    def clone(self, source: str):
-        """Clone a repository from a source."""
+    def clone(self, pull_path: str):
+        """Clone a repository from a source (pull_path)."""
+        pull_path = self.authenticate_path_if_possible(pull_path)
+
         # When cloning, self.path doesn't exist yet, so we need to use another CWD.
-        self._git_run("clone", source, self.path, cwd="/")
+        self._git_run("clone", pull_path, self.path, cwd="/")
         self._git_run("checkout", self.default_branch, cwd=self.path)
         self._git_repo_config()
 
@@ -121,8 +129,7 @@ class GitSCM(AbstractSCM):
         if force_push:
             push_command += ["--force"]
 
-        if GitHub.is_supported_url(push_path):
-            push_path = GitHub(push_path).authenticated_url
+        push_path = self.authenticate_path_if_possible(push_path)
 
         push_command += [push_path]
 
@@ -466,6 +473,9 @@ class GitSCM(AbstractSCM):
 
         self.clean_repo(attributes_override=attributes_override)
         # Fetch all refs at the given pull_path, and overwrite the `origin` references.
+
+        pull_path = self.authenticate_path_if_possible(pull_path)
+
         self._git_run(
             "fetch",
             "--prune",
--- a/src/lando/main/scm/git.py
+++ b/src/lando/main/scm/git.py
@@ -77,11 +77,12 @@ class GitSCM(AbstractSCM):
         self.default_branch = default_branch
         super().__init__(path)
 
-    def authenticate_path_if_possible(self, path: str) -> str:
-        """Return authenticated URL if supported."""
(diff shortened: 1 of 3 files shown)


src/lando/main/scm/git.py
```diff
@@ -90,10 +96,12 @@ def scm_name(cls) -> str:
         return "Git"
 
     @override
-    def clone(self, source: str):
-        """Clone a repository from a source."""
+    def clone(self, pull_path: str):
```
cgsheeh: This changes the argument name from `source` to `pull_path`, which doesn't match the `AbstractSCM` implementation. We should either update all instances to use `pull_path` or keep the variable name as `source`.

src/lando/main/scm/git.py
```diff
@@ -77,6 +77,12 @@ def __init__(self, path: str, default_branch: str = "main", **kwargs):
         self.default_branch = default_branch
         super().__init__(path)
 
+    def authenticate_path_if_possible(self, path: str) -> str:
```
cgsheeh: The docstring and implementation use `url`, whereas the method signature here uses `path`. We should use `url` as well as it aligns with other usages, and to avoid confusing with a filesystem path.

Nit: this can be a static/class method, since it doesn't rely on `self`.

src/lando/main/scm/git.py
```diff
@@ -77,6 +77,12 @@ def __init__(self, path: str, default_branch: str = "main", **kwargs):
...
+        """Return authenticated URL if supported."""
```
cgsheeh:
```suggestion
        """Return authenticated URL if supported by Github."""
```

Also, what does "supported" mean in this context?

cgsheeh (changes requested):
This PR has no test coverage. We should add tests for `authenticate_path_if_possible` and it's call sites.

src/lando/main/scm/git.py
zzzeid: This was moved to a reusable method from the original implementation on lines 124-125. It refers to this method: `if GitHub.is_supported_url(push_path):` (i.e., if it returns `True` or `False`). It doesn't actually refer to support by GitHub, but rather if it's a GitHub URL to start with. Left this generic so that we don't duplicate the docstrings in `GitHub.is_supported_url` and `GitHub.parse_url` but I can update the docstring to give a little more info.

zzzeid:
> This PR has no test coverage. We should add tests for `authenticate_path_if_possible` and it's call sites.

For clarification, the core functionality already has coverage in `src/lando/utils/tests/test_github.py` (it was pre-existing code). The new method is to avoid duplication of multiple uses of the same 2 lines, but can add a simple test for this.

src/lando/main/scm/git.py
zzzeid: Unfortunately we also use `path` in all the repo fields (e.g., `pull_path`, `push_path`, etc.) along with `url`. At some point it may be good to update all of those to refer to urls (or uris?). In theory we can use system paths in those values in tests so it's not super clear cut.

zzzeid:
> > This PR has no test coverage. We should add tests for `authenticate_path_if_possible` and it's call sites.
> 
> For clarification, the core functionality already has coverage in `src/lando/utils/tests/test_github.py` (it was pre-existing code). The new method is to avoid duplication of multiple uses of the same 2 lines, but can add a simple test for this.

Added test to make sure the new method is covered.

shtrom (approved):

src/lando/main/tests/test_scm_helpers.py
```diff
@@ -730,3 +731,21 @@ def test_strip_git_version_info_lines():
...
+@pytest.mark.parametrize(
+    "url,is_supported_url,result",
+    (
+        ("some_supported_url", True, "authenticated_url"),
+        ("some_unsupported_url", False, "some_unsupported_url"),
+    ),
+)
+@mock.patch("lando.main.scm.git.GitHub")
+def test_GitSCM__authenticate_path_if_possible(github, is_supported_url, url, result):
+    mock_github = mock.MagicMock()
```
cgsheeh, line 767: This test seems to be only testing the mocks. If I remove the `if GitHub.is_supported_url(url)` check from `authenticate_path_if_possible` entirely, both cases still pass. I think this is because in the `False` case the `url` and `result` are the same value, so the assertion can't tell if the short-circuit fired or if the mocked `authenticated_url` was returned, which I would expect the test to catch.

cgsheeh (changes requested):

cgsheeh (approved):

src/lando/main/tests/test_scm_helpers.py
zzzeid: It is meant to be a unit test (i.e., only testing the mocks) so there's coverage for the new method. `GitHub.is_supported_url` is tested elsewhere (in `test_github`).

I just added assertions to ensure that `GitHub.is_supported_url` is called the expected times with the expected values so that it can catch regressions.

cgsheeh (approved):

PR #1158, opened 2026-05-13
test: let test_ruff_format do its assertion (bug 2039175)

shtrom:
`check_*` methods raise `subprocess.CalledProcessError` exceptions on non-0 status codes. This prevent the rest of the logic of the test from doing its assertions, and showing a more palatable error message to the user.

### DIFF
--- a/src/lando/tests/test_formatting.py
+++ b/src/lando/tests/test_formatting.py
@@ -9,8 +9,9 @@ from lando.settings import LINT_PATHS
 
 def test_ruff_format():
     cmd = ("ruff", "format", "--diff")
-    output = subprocess.check_output(cmd + LINT_PATHS)
-    assert not output, "The python code does not adhere to the project style."
+    output = subprocess.run(cmd + LINT_PATHS, stdout=subprocess.PIPE)
+    assert not output.stdout, "The python code does not adhere to the project style."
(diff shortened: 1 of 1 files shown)


src/lando/tests/test_formatting.py
```diff
@@ -9,8 +9,9 @@
 
 def test_ruff_format():
     cmd = ("ruff", "format", "--diff")
-    output = subprocess.check_output(cmd + LINT_PATHS)
-    assert not output, "The python code does not adhere to the project style."
+    output = subprocess.run(cmd + LINT_PATHS, stdout=subprocess.PIPE)
```
Magnolia-Liu, line 12: Question: should similar changes be made to the other formatting tests?

shtrom: Good question! I don't think so. I had a quick check yesterday: both others use the `subprocess.call` method https://docs.python.org/3/library/subprocess.html#subprocess.call, which returns the status code, rather than throw an exception, so I think it's fine as it is (with the CAVEAT that `subprocess.call` is marked as ”Older high-level API“, but there's no indication of it getting deprecated, so I think we're good).

cgsheeh (approved):
LGTM!

PR #1159, opened 2026-05-13
settings: catch IntegrityError and render error in output (Bug 2037991)

cgsheeh:
Catch the `IntegrityError` thrown when trying to add a Phab API
key to a profile where there is already an existing PHID associated
with a user profile.

We add a test for this case, and a test for the happy patch since
it is currently untested. Move the `phabdouble` fixture into the
top-level `conftest.py` so it can be accessed from the `ui` app.

### DIFF
--- a/src/lando/ui/legacy/user_settings.py
+++ b/src/lando/ui/legacy/user_settings.py
@@ -47,7 +47,7 @@ def manage_api_key(request: WSGIRequest) -> JsonResponse:
         try:
             profile.save_phabricator_api_key(api_key, phid=phid)
         except IntegrityError:
-            logger.warning(
+            logger.info(
                 "Phabricator PHID `%s` is already linked to another Lando account.",
                 phid,
             )
--- a/src/lando/ui/legacy/user_settings.py
+++ b/src/lando/ui/legacy/user_settings.py
@@ -5,10 +5,29 @@ from django.db import IntegrityError
 from django.http import HttpResponseNotAllowed, JsonResponse
 
 from lando.main.auth import require_authenticated_user
+from lando.main.models.profile import Profile
 from lando.ui.legacy.forms import UserSettingsForm
 from lando.utils.phabricator import get_phabricator_client
 
 logger = logging.getLogger(__name__)
+
+
+def phid_conflict_response(phid: str) -> JsonResponse:
+    """Return a 400 response indicating the PHID is owned by another account."""
+    logger.info(
+        "Phabricator PHID `%s` is already linked to another Lando account.",
+        phid,
+    )
+    return JsonResponse(
+        {
+            "errors": {
+                "phabricator_api_key": [
+                    "This Phabricator account is already linked to another Lando user."
+                ]
+            }
+        },
+        status=400,
+    )
 
 
 @require_authenticated_user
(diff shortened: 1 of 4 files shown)


src/lando/ui/legacy/user_settings.py
```diff
@@ -43,6 +44,23 @@ def manage_api_key(request: WSGIRequest) -> JsonResponse:
         phid = whoami["phid"]
         logger.debug("Phabricator API key verified for PHID `%s`.", phid)
 
-        profile.save_phabricator_api_key(api_key, phid=phid)
+        try:
+            profile.save_phabricator_api_key(api_key, phid=phid)
+        except IntegrityError:
+            logger.warning(
```
shtrom: If the error message is clearly visible to the user, we may not need it in the logs, or at least not at this level.

shtrom (approved):

Magnolia-Liu (approved):

src/lando/ui/legacy/user_settings.py
```diff
@@ -43,6 +44,23 @@ def manage_api_key(request: WSGIRequest) -> JsonResponse:
         phid = whoami["phid"]
         logger.debug("Phabricator API key verified for PHID `%s`.", phid)
 
-        profile.save_phabricator_api_key(api_key, phid=phid)
+        try:
+            profile.save_phabricator_api_key(api_key, phid=phid)
```
zzzeid, line 74: Can we not check for existing PHIDs here, avoiding the integrity error altogether?

zzzeid (changes requested):

src/lando/ui/legacy/user_settings.py
zzzeid: I agree, this should not be an issue in the backend if we can avoid the error altogether.

src/lando/ui/legacy/user_settings.py
cgsheeh: We could, but it would add another DB query on every API key update, and we'd still want to keep the `IntegrityError` handler to cover the potential race condition (unlikely but technically possible). Happy to change it, this just follows EAFP and is more efficient for the happy path. WDYT?

zzzeid: This endpoint is not used frequently, the extra DB query would be negligible.

The code flow would be more logical if we're explicit about detecting the error state, rather than working around the issue. The integrity error could also (in theory, though very unlikely) occur in a different situation, so here we'd be handling the expected error case that we already discovered.

cgsheeh: I added a check for a duplicate PHID. Personally I don't think it's a meaningful improvement - we're still catching the `IntegrityError` to catch the race condition, so it's more code for the same functionality - but happy to move forward with it.

zzzeid: Hmm what's the point of catching the `IntegrityError`? We don't need to do that as long as we know there is no duplicate value. Otherwise it would be a misleading error since if triggered, it is some other issue that is causing that and not the PHID.

I would suggest removing the try/except altogether.

cgsheeh: > Hmm what's the point of catching the `IntegrityError`? We don't need to do that as long as we know there is no duplicate value. Otherwise it would be a misleading error since if triggered, it is some other issue that is causing that and not the PHID.

The `IntegrityError` handler covers the race between our `.exists()` and `.save()` — without it, two concurrent requests linking the same PHID would 500 instead of returning a clean 400. And since `phabricator_phid` is the only unique field we touch here, an `IntegrityError` can't really come from anywhere else.
                                                                                                                                                                   
Yes, the race is unlikely — but the bug is specifically about an unhandled server error, and switching to just an `exists()` check doesn't actually resolve that; it just makes the window smaller.
                                                                                                                                                                   
This is why I preferred just the `try`/`except` version — it was less code, handled both cases, and required fewer DB lookups on the happy path.

zzzeid: To handle the (very unlikely) race condition we could do the check and the save with a lock, but I think that's overkill. This endpoint is infrequently used, I think the basic issue is to present a meaningful error to the user and avoid having a server issue.

I won't split more hairs here but I think preventing an integrity error is better than handling it when it happens.

zzzeid: TL;DR this should be handled as a validation problem, and not a database integrity problem.

zzzeid (approved):
I won't block this PR on the outstanding details of the implementation, particularly since this endpoint is a legacy one, and requires a revamp/renovation anyway (it should go in `lando.api`, and should use class-based views, and have overall better validation.)

PR #1161, opened 2026-05-14
github: add ability to edit commit message from PR in Lando UI (bug 2003888)

Magnolia-Liu:
- Added a view callable by GitHub to update pull request content with data from Lando
- Added editable commit body and title fields in the PR UI
  - Land button is disabled while fields are being edited to prevent landing stale content
- Added an "Edit" and "Save" commit message button that refreshes the page on submit + cancel changes button
  - Text areas are disabled during save to prevent concurrent edits
- Added error display for invalid input

Demo of UI:

https://github.com/user-attachments/assets/a8e09848-c810-4d01-bf3f-c252a379f418

### DIFF
--- a/src/lando/api/tests/test_views.py
+++ b/src/lando/api/tests/test_views.py
@@ -1,8 +1,25 @@
 from unittest import mock
 
 import pytest
+from django.test import Client
 
 from lando.main.models import Repo, SCMType
+
+
+@pytest.fixture
+def repo_mc_github_api_client(repo_mc):
+    repo_mc(SCMType.GIT, name="git-repo")
+
+    mock_github_api_client = mock.MagicMock()
+    mock_github_api_client.repo_is_private = False
+    return mock_github_api_client
+
+
+@pytest.fixture
+def csrf_client(user, user_plaintext_password):
+    csrf_client = Client(enforce_csrf_checks=True)
+    csrf_client.login(username=user.username, password=user_plaintext_password)
+    return csrf_client
 
 
 @pytest.mark.django_db(transaction=True)
@@ -162,3 +179,108 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
     mock_github_api_client.repo_is_private = False
     test = client.get(f"/api/pulls/{repo.name}/1/landing_jobs")
     assert test.status_code == 200
+
+
+@mock.patch("lando.api.views.GitHubAPIClient")
+@pytest.mark.django_db(transaction=True)
+@pytest.mark.parametrize(
+    "payload, expected_status, expected_response",
+    [
+        (
+            {
+                "title": "Valid New Title",
+                "body": "Valid New Body",
+            },
+            204,
+            b"",
+        ),
+        (
+            {
+                "title": "",
+                "body": "Valid New Body",
+            },
+            400,
+            {
+                "title": ["This field is required."],
+            },
+        ),
+        (
+            {
+                "title": "a" * 300,
+                "body": "Valid New Body",
+            },
+            400,
+            {
+                "title": ["Ensure this value has at most 256 characters (it has 300)."],
+            },
+        ),
+    ],
+)
+def test__views__pull_request_content_api_view(
+    github_api_client,
+    authenticated_client,
+    repo_mc_github_api_client,
+    payload,
+    expected_status,
+    expected_response,
+):
+    """Test PullRequestContentAPIView validation and success responses."""
+
+    github_api_client.return_value = repo_mc_github_api_client
+
+    mock_pull_request = mock.MagicMock()
+
+    repo_mc_github_api_client.build_pull_request.return_value = mock_pull_request
+    repo_mc_github_api_client.update_pull_request_content.return_value = payload
+
+    result = authenticated_client.put(
+        "/api/pulls/git-repo/100",
+        data=payload,
+        content_type="application/json",
+    )
+
+    assert result.status_code == expected_status
+
+    if expected_status == 204:
+        assert result.content == expected_response
+    else:
+        response_json = result.json()
+        for key, value in expected_response.items():
+            assert response_json.get(key) == value
+
+
+@mock.patch("lando.api.views.GitHubAPIClient")
+@pytest.mark.django_db(transaction=True)
+def test__views__pull_request_content_api_view__unauthenticated(
+    github_api_client, client, repo_mc_github_api_client
+):
+    """An anonymous PUT should be rejected by the auth decorator."""
+
+    github_api_client.return_value = repo_mc_github_api_client
+
+    result = client.put(
+        "/api/pulls/git-repo/100",
+        data={"title": "Valid New Title", "body": "Valid New Body"},
+        content_type="application/json",
+    )
+
+    # return 403 instead of 401 due to bug with django auth decorator. See: https://github.com/django/django/blob/main/django/core/handlers/exception.py#L75C51-L75C54
+    assert result.status_code == 403
+    assert b"403 Forbidden" in result.content
+
+
+@mock.patch("lando.api.views.GitHubAPIClient")
+@pytest.mark.django_db(transaction=True)
+def test__views__pull_request_content_api_view__missing_csrf_token(
+    github_api_client, csrf_client, repo_mc_github_api_client
+):
+    """An authenticated PUT without a CSRF token should be rejected."""
+    github_api_client.return_value = repo_mc_github_api_client
+
+    result = csrf_client.put(
+        "/api/pulls/git-repo/100",
+        data={"title": "Valid New Title", "body": "Valid New Body"},
+        content_type="application/json",
+    )
+    assert result.status_code == 403
+    assert b"403 Forbidden" in result.content
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -6,7 +6,7 @@ from typing import Callable
 
 from django import forms
 from django.core.handlers.wsgi import WSGIRequest
-from django.http import Http404, HttpRequest, JsonResponse
+from django.http import Http404, HttpRequest, HttpResponse, JsonResponse
 from django.utils.decorators import method_decorator
 from django.utils.html import escape
 from django.views import View
@@ -214,7 +214,11 @@ class PullRequestAPIView(View, PrivateRepoPermissionMixin):
     def dispatch(
         self, request: WSGIRequest, repo_name: str, pull_number: int, *args, **kwargs
     ) -> JsonResponse:
-        self.target_repo = Repo.objects.get(name=repo_name)
+        try:
+            self.target_repo = Repo.objects.get(name=repo_name)
+        except Repo.DoesNotExist as e:
+            raise Http404 from e
+
         self.client = GitHubAPIClient(self.target_repo.url)
         self.raise_404_if_needed(request, self.client)
 
@@ -334,3 +338,33 @@ class PullRequestChecksAPIView(PullRequestAPIView):
             # The StaleMetadataException error message is safe for user consumption.
             return JsonResponse({"errors": [str(exc)]}, status=500)
         return JsonResponse(warnings_and_blockers)
+
+
+class PullRequestContentAPIView(PullRequestAPIView):
+    """Handle pull request content updates in the API."""
+
+    @method_decorator(require_authenticated_user)
+    def put(
+        self, request: WSGIRequest, repo_name: str, pull_number: int
+    ) -> JsonResponse:
+        """Update pull request content"""
+        if not self.target_repo.user_can_push(request.user):
+            return JsonResponse(
+                {"errors": ["You do not have permission to modify this pull request."]},
+                status=403,
+            )
+
+        class Form(forms.Form):
+            body = forms.CharField(required=False)
+            title = forms.CharField(max_length=256)
+
+        form = Form(json.loads(request.body))
+        if not form.is_valid():
+            return JsonResponse(form.errors, status=400)
+
+        self.client.update_pull_request_content(
+            self.pull_request.number,
+            form.cleaned_data["body"],
+            form.cleaned_data["title"],
+        )
+        return HttpResponse(status=204)
--- a/src/lando/static_src/legacy/css/pages/StackPage.scss
+++ b/src/lando/static_src/legacy/css/pages/StackPage.scss
@@ -31,6 +31,19 @@
   & table {
     margin-bottom: 0;
   }
+
+  #commit-title,
+  #commit-body {
+    font-family: "Fira Mono", monospace;
+  }
+
+  #commit-title {
+    min-height: 50px;
+  }
+
+  #commit-body {
+    min-height: 75px;
+  }
 }
 
 .StackPage-revision {
--- a/src/lando/static_src/legacy/js/components/Stack.js
+++ b/src/lando/static_src/legacy/js/components/Stack.js
@@ -137,7 +137,7 @@ $.fn.stack = function () {
               method: "GET",
             }).then(async (response) => {
               $("#save-edit-pr").prop("disabled", false);
-              if (response.status == 204) {
+              if (response.status == 200) {
                 var result = await response.json();
                 var blockers = result.blockers;
                 var warnings = result.warnings;
@@ -264,7 +264,7 @@ $.fn.stack = function () {
               "X-CSRFToken": csrf_token,
             },
           }).then(async (response) => {
-            if (response.status === 200) {
+            if (response.status === 204) {
               $("#commit-title-error").text("");
               $("#commit-body-error").text("");
               $("#commit-title").removeClass("is-danger");
--- a/src/lando/static_src/legacy/js/components/Stack.js
+++ b/src/lando/static_src/legacy/js/components/Stack.js
@@ -79,8 +79,9 @@ $.fn.stack = function () {
     // This should be cleaned up as part of bug 1995754.
     var is_pull_request_page = Boolean($("button.post-landing-job").length);
     if (is_pull_request_page) {
+      var saved_landing_state = null;
       var pull_request_button = $("button.post-landing-job");
-
+      $("#save-edit-pr").prop("disabled", true);
       $("#acknowledge-warnings").on("click", function () {
         if (this.checked) {
           pull_request_button.prop("disabled", false);
@@ -135,6 +136,7 @@ $.fn.stack = function () {
             fetch(`/api/pulls/${repo_name}/${pull_number}/checks`, {
               method: "GET",
             }).then(async (response) => {
+              $("#save-edit-pr").prop("disabled", false);
               if (response.status == 200) {
                 var result = await response.json();
                 var blockers = result.blockers;
@@ -186,6 +188,15 @@ $.fn.stack = function () {
                     .addClass("is-warning");
                   pull_request_button.html("Acknowledge warnings to continue");
                 }
+                saved_landing_state = {
+                  html: pull_request_button.html(),
+                  disabled: pull_request_button.prop("disabled"),
+                  classes: pull_request_button.attr("class"),
+                  ack_section_visible: $(".acknowledge-warnings-section").is(
+                    ":visible",
+                  ),
+                  ack_checked: $("#acknowledge-warnings").prop("checked"),
+                };
               } else {
                 // TODO: handle this case. See bug 1996000.
               }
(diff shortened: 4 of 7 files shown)


src/lando/utils/github.py
zzzeid, line 1: To reduce chances of this PR getting blocked, we can take out the portions in this PR that were borrowed from #1073 and instead just keep the `_patch` and `_patch` methods. We can always re-add the delimiter functionality after that PR is more robust.

src/lando/api/views.py
```diff
@@ -326,3 +326,21 @@ def get(
             # The StaleMetadataException error message is safe for user consumption.
             return JsonResponse({"errors": [str(exc)]}, status=500)
         return JsonResponse(warnings_and_blockers)
+
+
+class PullRequestContentAPIView(PullRequestAPIView):
+    def put(
```
zzzeid, line 347: Would be good to add docstrings to this method and the class.

zzzeid (commented):
Just a few drive-by comments

src/lando/api/views.py
```diff
@@ -334,3 +334,25 @@ def get(
             # The StaleMetadataException error message is safe for user consumption.
             return JsonResponse({"errors": [str(exc)]}, status=500)
         return JsonResponse(warnings_and_blockers)
+
+
+class PullRequestContentAPIView(PullRequestAPIView):
+    """Handle pull request content updates in the API."""
+
+    def put(
```
shtrom, line 347: This should be limited this to authenticated users. 

```suggestion
    @method_decorator(require_authenticated_user)
    def put(
```

and tested.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -71,7 +71,7 @@ $.fn.stack = function() {
     // This should be cleaned up as part of bug 1995754.
     var is_pull_request_page = Boolean($('button.post-landing-job').length);
     if (is_pull_request_page) {
-        var pull_request_button = $('button.post-landing-job');
+        var pull_request_button = $('button.post-landing-job');
```
shtrom: Be careful about whitespaces changes left behind (:

There's another PR coming up that will put a linter on JS and CSS, so it won't be a problem for long, but best keep an eye out for them.

FWIW, when I prepare commits, I use `git add -p` which allows you to include, modify, or exclude every chunk of change into the index. It's a great way to find stray whitespace changes!

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
...
+                document.getElementById('cancel-edit-pr').classList.remove('is-hidden');
+                document.getElementById('commit-title').focus();
+                document.getElementById("post-landing-job").disabled = true;
+                return;
+            }
+
+            var body = document.getElementById('commit-body').value;
+            var title = document.getElementById('commit-title').value;
+
+                fetch(`/api/pulls/${repo_name}/${pull_number}`, {
```
shtrom: Indent is off here (also a problem that will disappear shortly!)

src/lando/ui/jinja2/stack/pull_request.html
```diff
@@ -61,13 +62,28 @@ <h1>
...
                 <th>Commit Body</th>
-                <td>{{ pull_request.body }}</td>
+                <td>
+                    <textarea readonly id="commit-body" class="textarea">{{ pull_request.body }}</textarea>
+                    <p class="help is-danger" id="commit-body-error"></p>
+                </td>
             </tr>
         </table>
+        <p>
+            <button id="save-pr"
```
shtrom: nit: As this button is modal, it may be clearer for its name to evoke all the possible modes, e.g., `edit-save-pr` or something in this ilk.

src/lando/static_src/legacy/css/pages/StackPage.scss
```diff
@@ -32,6 +32,19 @@
   & table {
     margin-bottom: 0;
   }
+
+  #commit-title,
+  #commit-body {
+    font-family: 'Fira Mono', monospace;
+  }
+
+  #commit-title {
```
shtrom, line 40: nit: we could make those less `textarea`-looking when not in edit mode.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
                 }
             });
         });
-    };
-  });
+
+        $('button.save-pr').on('click', function(e) {
+            var save_pr_button = this;
+
+            if (save_pr_button.dataset.mode === 'edit') {
```
shtrom: With this, the default logic is to try to save, unless we are in `edit` mode.

As saving is an action with side-effects beyond the page, I think it may be safer to flip the logic, or have an explicit default.

Something like

```
if (....dataset.mode === save) {
  ...
  return;
}

// default / edit logic
```

Or a switch-like construct

```
witch(....dataset.mode) {
  case "edit":
    ...
    break;
  case "save":
    ...
    break;
  default:
    // error handling, maybe reset the mode to 'edit'?
} 
```
The `switch` is probably cleaner, but defaulting to `edit` is likely sufficient, and skips one error-handling step that users wouldn't be able to do anything about anyway.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
...
+            var body = document.getElementById('commit-body').value;
+            var title = document.getElementById('commit-title').value;
+
+                fetch(`/api/pulls/${repo_name}/${pull_number}`, {
+                    method: 'PUT',
+                    body: JSON.stringify({"body": body, "title": title}),
+                    headers: {
+                        'Accept': 'application/json',
+                        'Content-Type': 'application/json',
+                        'X-CSRFToken': csrf_token
```
shtrom: note that we should test the API endpoint in case there is no CSRF token.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
                 }
             });
         });
-    };
-  });
+
+        $('button.save-pr').on('click', function(e) {
```
shtrom: I think using `$("#save-pr")` here would be safer.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
...
+            var save_pr_button = this;
+
+            if (save_pr_button.dataset.mode === 'edit') {
+                document.getElementById('commit-title').removeAttribute('readonly');
```
shtrom: I think you could use `$(...)` here, as we do in the rest of the file.

```suggestion
                $('#commit-title').removeAttribute('readonly');
```

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
...
+                        'Accept': 'application/json',
+                        'Content-Type': 'application/json',
+                        'X-CSRFToken': csrf_token
+                    }
+                }).then(async (response) => {
+                console.log(response);
+                    if (response.status === 400) {
+                        var result = await response.json();
+                        if (result.title){ 
+                            document.getElementById("commit-title-error").textContent = result.title;
```
shtrom: `$("#...")`

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
...
+                        window.location.reload();
+                    }
+                    else{
+                        console.error("error updating pull request:", response);
+                    }
+            });
+        });
+
+        $('button.cancel-edit-pr').on('click', function(e) {
+            document.getElementById('commit-title').setAttribute('readonly', true);
```
shtrom: `$("#...")`

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -198,6 +198,80 @@ $.fn.stack = function() {
...
+                        save_pr_button.classList.add("is-loading");
+                        document.getElementById("commit-title-error").textContent = "";
+                        document.getElementById("commit-body-error").textContent = ""; 
+                        document.getElementById("commit-title").classList.remove("is-danger");
+                        document.getElementById("commit-body").classList.remove("is-danger");
+                        document.getElementById("commit-title").disabled = true;
+                        document.getElementById("commit-body").disabled = true;
+                        document.getElementById('cancel-edit-pr').classList.add('is-hidden');
+
+                        window.location.reload();
```
shtrom: What is the purpose of reloading here?

We could provide a smoother experience by updating the DOM and returning the form to static mode.

Maybe rather than doing the mode selection based on the `save-pr` button, you could have it on the whole `form`. This would also allow you to control the look and visibility of child element in CSS rather than having to change them dynamically in JS.

src/lando/ui/jinja2/stack/pull_request.html
```diff
@@ -61,13 +62,28 @@ <h1>
...
             <tr>
                 <th>Commit Body</th>
-                <td>{{ pull_request.body }}</td>
+                <td>
+                    <textarea readonly id="commit-body" class="textarea">{{ pull_request.body }}</textarea>
+                    <p class="help is-danger" id="commit-body-error"></p>
+                </td>
             </tr>
         </table>
+        <p>
```
shtrom, line 78: It may make sense to have title, body and button as a div-like `form` in a single table row.

src/lando/ui/jinja2/stack/pull_request.html
```diff
@@ -6,7 +6,8 @@ <h1>
             <a href="{{ pull_request.html_url }}">{{ pull_request.title }}</a> #{{ pull_request.number }} ({{ target_repo.name }}@{{ target_repo.default_branch }})
         </h1>
         <p>
-            <button class="button is-large is-loading post-landing-job has-text-white"
+            <button id="post-landing-job"
+                    class="button is-large is-loading post-landing-job has-text-white"
```
shtrom, line 10: Having an `id` and a `class` named the same could be a bit of a footgun for our future selves.

Having an `id` is good, though. Perhaps we should just make sure that everything rely on it being an `id`, and remove the class altogether.

This would however force any page to only have a single landing button. I don't think we have it any differently at the moment, but maybe it's not something we want to commit to right here?

If anything, I'd look for ways to avoid needing to make this particular change, if possible.

src/lando/utils/github.py
```diff
@@ -381,6 +390,15 @@ def add_comment_to_pull_request(self, pull_number: int, comment: str) -> dict:
             json={"body": comment},
         )
 
+    def update_pull_request_content(
```
shtrom, line 393:
```suggestion
    def update_pull_request_description(
```

It may also make sense to make the `title` optional, so we can use it to only change the `body`.

shtrom (changes requested):
Mostly stylistic comments, but this is looking good, and the overall logic makes sense.

src/lando/api/tests/test_views.py
```diff
@@ -162,3 +162,80 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
...
+                "body": "Valid New Body",
+            },
+            400,
+            {
+                "title": ["Ensure this value has at most 256 characters (it has 300)."],
+            },
+        ),
+    ],
+)
+def test__views__pull_request_content_api_view(
```
shtrom, line 219: We'll need to cover the CSRF token and authenticated-user logic, too.

src/lando/api/tests/test_views.py
```diff
@@ -162,3 +162,80 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
...
+
+    repo_mc(SCMType.GIT, name="git-repo")
+
+    mock_github_api_client = mock.MagicMock()
+    mock_github_api_client.repo_is_private = False
+    github_api_client.return_value = mock_github_api_client
+
+    mock_pull_request = mock.MagicMock()
+    mock_pull_request.title = "Hello"
+    mock_pull_request.body = "World"
```
zzzeid: Are these values still in use?

src/lando/api/views.py
zzzeid: Yes, moreover, we should check that the user has the correct permission set in `self.target_repo`. This way if they do not have that permission, they will be unable to modify this pull request.

```python
if not self.target_repo.user_can_push(request.user):
    ...
```

This will be the first "write" method triggered on behalf of the request user that is implemented for pull requests, which is why this is not done elsewhere.

src/lando/static_src/legacy/css/pages/StackPage.scss
zzzeid: I think this would be a nice to have too.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -226,6 +226,90 @@ $.fn.stack = function () {
           }
         });
       });
+
+      $("button.save-pr").on("click", function (e) {
```
zzzeid: I noticed in the UI demo that there is no feedback on the save button when it is clicked (unless I missed it?) This would be good for usability (e.g., disable the button temporarily, or change the text, etc...

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -226,6 +226,90 @@ $.fn.stack = function () {
...
+            document
+              .getElementById("commit-body")
+              .classList.remove("is-danger");
+            document.getElementById("commit-title").disabled = true;
+            document.getElementById("commit-body").disabled = true;
+            document
+              .getElementById("cancel-edit-pr")
+              .classList.add("is-hidden");
+
+            window.location.reload();
```
zzzeid: It would be good to file a bug under https://bugzilla.mozilla.org/show_bug.cgi?id=2000954 to update this to instead re-trigger the checks without a full refresh.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -226,6 +226,90 @@ $.fn.stack = function () {
...
+          } else {
+            console.error("error updating pull request:", response);
```
zzzeid: Console errors should be used for debugging, and should be removed before this is merged.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -226,6 +226,90 @@ $.fn.stack = function () {
...
+        document.getElementById("commit-title").classList.remove("is-danger");
+        document.getElementById("commit-body").classList.remove("is-danger");
+        document.getElementById("commit-title").value =
+          document.getElementById("commit-title").defaultValue;
+        document.getElementById("commit-body").value =
+          document.getElementById("commit-body").defaultValue;
+        document.getElementById("save-pr").dataset.mode = "edit";
+        document.getElementById("save-pr").textContent = "Edit Commit Message";
+        document.getElementById("cancel-edit-pr").classList.add("is-hidden");
+        document.getElementById("post-landing-job").disabled = false;
```
zzzeid: nit: you can save the elements in variables and re-use them, instead of doing `document.getElementById` multiple times for each element.

src/lando/ui/jinja2/stack/pull_request.html
zzzeid: It may be a good time to reduce clutter here. If `.post-landing-job` is used elsewhere we could change it to `#post-landing-job`.

Having said that, not sure if having them have the same name is an issue. I suspect you added this to be able to easily use `document.getElementById`?.

src/lando/ui/jinja2/stack/pull_request.html
```diff
@@ -61,13 +62,28 @@ <h1>
             </tr>
             <tr>
                 <th>Commit Title</th>
-                <td>{{ pull_request.title }}</td>
+                <td>
+                    <textarea readonly id="commit-title" class="textarea">{{ pull_request.title }}</textarea>
+                    <p class="help is-danger" id="commit-title-error"></p>
```
zzzeid: This is a bit out of scope for this PR, I [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2041769) to remove the use of a table here altogether.

src/lando/utils/github.py
zzzeid: I think this method is more generic than just updating the description. It can be used to update any attribute on the PR.

package-lock.json
zzzeid, line 1: This is probably erroneously showing up here because of the merge. I would recommend squashing this entire PR at some point, before attempting to merge it, to clean up the commit history.

zzzeid (changes requested):

src/lando/ui/jinja2/stack/pull_request.html
shtrom: Having the same name is probably not a technical issue, but from a human differentiation perspective, it's probably best if they are different.

src/lando/static_src/legacy/js/components/Stack.js
shtrom, line 1: I'm realising that the PR logic is getting added to the `Stack.js`, and now wondering whether we should have anything PR-related in separate files?

@zzzeid wdyt?

zzzeid: Yes, this is already filed/being tracked -> https://bugzilla.mozilla.org/show_bug.cgi?id=1995754.

src/lando/api/views.py
shtrom: What if the user pushing the PR doesn't have SCM3? They could edit their PR in GitHub, but not here. Are we assuming that only the landing requester will edit the commit message here?

src/lando/api/views.py
```diff
@@ -339,10 +339,16 @@ def get(
...
+    @method_decorator(require_authenticated_user)
     def put(
         self, request: WSGIRequest, repo_name: str, pull_number: int
     ) -> JsonResponse:
         """Update pull request content"""
+        if not self.target_repo.user_can_push(request.user):
+            return JsonResponse(
+                {"errors": ["You are not allowed to push to this repository."]},
+                status=403,
+            )
```
shtrom, line 355: I wonder if we should make a little assertion util that would return a consistent error everywhere it's called.

src/lando/ui/jinja2/stack/pull_request.html
```diff
@@ -63,26 +63,26 @@ <h1>
             <tr>
                 <th>Commit Title</th>
                 <td>
-                    <textarea readonly id="commit-title" class="textarea">{{ pull_request.title }}</textarea>
-                    <p class="help is-danger" id="commit-title-error"></p>
+                    <p id="commit-title">{{ pull_request.title }}</p>
+                    <p id="commit-title-error" class="help is-danger"></p>
```
shtrom: I wouldn't set the `is-danger` class on those, even if empty. It can be set as needed by the JS.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -227,24 +227,49 @@ $.fn.stack = function () {
...
-          save_pr_button.textContent = "Save Commit Message";
-          document
-            .getElementById("cancel-edit-pr")
-            .classList.remove("is-hidden");
-          document.getElementById("commit-title").focus();
-          document.getElementById("post-landing-job").disabled = true;
+      $("#save-edit-pr").on("click", function (e) {
+        var save_edit_pr_button = $(this);
+
+        if (save_edit_pr_button.attr("data-mode") === "saved") {
```
shtrom: Could we swap the logic of this method, so we do the case for `saved` by default, and the `fetch` only when we are sure we have `data-mode===edit`?

It may be worth splitting those two into separate methods, e.g., 

```javascript
if (save_edit_pr_button.attr("data-mode") === "edit") {
  return save_pr_description();
} 
return edit_pr_description();
```

src/lando/static_src/legacy/js/components/Stack.js
shtrom: What if we just... Don't reload, and instead set the title and body to what the textareas contain?

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -256,59 +281,58 @@ $.fn.stack = function () {
...
-          document.getElementById("commit-body").defaultValue;
-        document.getElementById("save-pr").dataset.mode = "edit";
-        document.getElementById("save-pr").textContent = "Edit Commit Message";
-        document.getElementById("cancel-edit-pr").classList.add("is-hidden");
-        document.getElementById("post-landing-job").disabled = false;
+      $("#cancel-edit-pr").on("click", function (e) {
+        const pTitle = document.createElement("p");
+        const pBody = document.createElement("p");
+        const textareaTitle = $("#commit-title")[0];
+        const textareaBody = $("#commit-body")[0];
```
shtrom: Why do we use array indexing here? This seems odd, and we don't do it before in tho code, e.g., https://github.com/mozilla-conduit/lando/pull/1161/changes/8726f39b11c08695895b0a506e34cad6c652edd4..e6cbe4c36dfa3a2b52f622538df1f06eaa3bbc86#diff-4d791453e88ede62fd62f9e15ecccd2a546f9d0e6fe016091adb198d09a6ca67R272.

If it's not needed (which I'm not sure, tbh), then we should probably be consistent and use the same access type throughout the code.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -256,59 +281,58 @@ $.fn.stack = function () {
...
-              .classList.add("is-hidden");
-
+            save_edit_pr_button.addClass("is-loading");
+            $("#commit-title-error").text("");
+            $("#commit-body-error").text("");
+            $("#commit-title").removeClass("is-danger");
+            $("#commit-body").removeClass("is-danger");
+            $("#commit-title").prop("disabled", true);
+            $("#commit-body").prop("disabled", true);
+            $("#cancel-edit-pr").addClass("is-hidden");
```
shtrom: Perhaps we should have utility functions to set/reset the errors, e.g., something like

```javascript
function set_error_for(selector, error_text, error_class) {
  $(selector + "-error").text(error_text);
  $(selector).removeClass("is-hidden");
  $(selector).removeClass("is-danger");
  $(selector).addClass(error_class);
  $(selector).prop("disabled", error_text === "");
}

File: src/lando/static_src/legacy/js/components/Stack.js (hunk 27 above)
Comment (Magnolia-Liu) [reply src/lando/static_src/legacy/js/components/Stack.js]: I can't seem to leave a reply on the relevant comment, but the reason the page reloads after a successful save is to re-run the landing checks.

I agree this isn't an ideal solution. It would be better to re-run the checks without reloading the page, but my goal for this bug was to land the simplest working solution first and then improve the user experience in a follow-up bug.

Comment (shtrom) [reply src/lando/static_src/legacy/js/components/Stack.js]: Ah, yep, that makes sense, as some checks are dependent on the commit description.

It may make sense to make the checks UI more functional, so we can call it on demand.

That's for a later bug, though. Reload here is fine for now 👍🏼

Comment (shtrom) [reply src/lando/static_src/legacy/js/components/Stack.js]: (I'll hide this convo as resolved, so we don't hijack this comment's thread)

File: src/lando/api/tests/test_views.py [hunk 29]
```diff
@@ -162,3 +163,125 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
...
+
+    result = client.put(
+        "/api/pulls/git-repo/100",
+        data={"title": "Valid New Title", "body": "Valid New Body"},
+        content_type="application/json",
+    )
+
+    assert result.status_code == 403
+    assert b"403 Forbidden" in result.content
+    assert b"CSRF" not in result.content
```
zzzeid: What is the purpose of this line? Error messages in production do not contain debug info, and this seems to be testing a debug output. Lastly, CSRF is not so much auth related as it is to validate that the request is coming from an expected source (even though it could be an anonymous user).

Also since this is tested below, this line can be removed from this test.

src/lando/api/tests/test_views.py
```diff
@@ -162,3 +163,125 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
...
+    github_api_client.return_value = mock_github_api_client
+
+    result = client.put(
+        "/api/pulls/git-repo/100",
+        data={"title": "Valid New Title", "body": "Valid New Body"},
+        content_type="application/json",
+    )
+
+    assert result.status_code == 403
+    assert b"403 Forbidden" in result.content
```
zzzeid, line 269: Technically, this should return a `401`. `403` is generally when a user is authenticated, but does not have the correct authorization. I'm wondering if this is a pre-existing issue with the auth decorator.

src/lando/api/views.py
zzzeid: I already have this functionality partially implemented in https://github.com/mozilla-conduit/lando/pull/1160.

zzzeid (commented):
Looking good, just a couple of questions/comments. Is it possible to see a new video with the latest UI implementation?

src/lando/api/tests/test_views.py
Magnolia-Liu: This seems to be an issue with the auth decorator, see: https://github.com/django/django/blob/main/django/core/handlers/exception.py#L75C51-L75C54

src/lando/ui/jinja2/stack/pull_request.html
shtrom: > If .post-landing-job is used elsewhere we could change it to #post-landing-job.

@zzzeid I think you meant “if it ISN'T used elsewhere”? (otherwise the ids will collide).

src/lando/api/views.py
zzzeid: >Are we assuming that only the landing requester will edit the commit message here?

I think this is a conservative assumption. More ideally we could check if the Lando user is also the author, but I think that allowing a person who has the ability to land a patch to also modify the commit message is fine.

src/lando/api/views.py
```diff
@@ -334,3 +334,31 @@ def get(
...
+    """Handle pull request content updates in the API."""
+
+    @method_decorator(require_authenticated_user)
+    def put(
+        self, request: WSGIRequest, repo_name: str, pull_number: int
+    ) -> JsonResponse:
+        """Update pull request content"""
+        if not self.target_repo.user_can_push(request.user):
+            return JsonResponse(
+                {"errors": ["You are not allowed to push to this repository."]},
```
zzzeid: This error message is a little misleading. Though we are checking the push permission, this error should reference the fields being edited.
```suggestion
                {"errors": ["You do not have permission to modify this pull request."]},
```

Or something similar.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -201,6 +201,7 @@ $.fn.stack = function () {
         fetch(`/api/pulls/${repo_name}/${pull_number}/landing_jobs`, {
           method: "POST",
           body: JSON.stringify({ head_sha: head_sha }),
+          //list of warnings:
```
zzzeid: Is this forgotten by mistake here?

src/lando/api/tests/test_views.py
```diff
@@ -162,3 +163,124 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
...
+):
+    """An authenticated PUT without a CSRF token should be rejected."""
+    repo_mc(SCMType.GIT, name="git-repo")
+
+    mock_github_api_client = mock.MagicMock()
+    mock_github_api_client.repo_is_private = False
+    github_api_client.return_value = mock_github_api_client
+
+    csrf_client = Client(enforce_csrf_checks=True)
+    csrf_client.login(username=user.username, password=user_plaintext_password)
```
zzzeid: nit: this could go into a fixture (e.g., `csrf_client`).

src/lando/api/tests/test_views.py
```diff
@@ -162,3 +163,124 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
...
+@pytest.mark.django_db(transaction=True)
+def test__views__pull_request_content_api_view__missing_csrf_token(
+    github_api_client, user, user_plaintext_password, repo_mc
+):
+    """An authenticated PUT without a CSRF token should be rejected."""
+    repo_mc(SCMType.GIT, name="git-repo")
+
+    mock_github_api_client = mock.MagicMock()
+    mock_github_api_client.repo_is_private = False
+    github_api_client.return_value = mock_github_api_client
```
zzzeid: This section is repeated a few times in the new tests. It could go into a fixture that gets reused across the different tests (e.g., `repo_github_api_client`).

zzzeid (changes requested):
Looking good! Just a couple of small changes and I think this will be ready to go.

src/lando/api/tests/test_views.py
shtrom: Let's add this observation as a comment (:

src/lando/api/tests/test_views.py
shtrom: This could be the basis for the GitHubDouble https://bugzilla.mozilla.org/show_bug.cgi?id=2005686

src/lando/api/views.py
shtrom: We check this in multiple parts of the code. We should refactor this into a small  utility that performs the check and returns a consistent error and message (403 is appropriate, this time (: ).

shtrom: Ok... I'm leaving this comment here just because... It's funny that I commented the same thing twice 😆  Zeid has replied more usefully https://github.com/mozilla-conduit/lando/pull/1161/changes/c8a10e0e33f728aae106c100be1887bf68841ab0..fbdff96ca8c76e048d672b4a45825d15cba9c332#r3357861821

src/lando/static_src/legacy/css/pages/StackPage.scss
shtrom: I think this is fixed in the code, by swapping the elements.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -1,277 +1,348 @@
...
+          fetch(`/api/pulls/${repo_name}/${pull_number}`, {
+            method: "PUT",
+            body: JSON.stringify({ body: body, title: title }),
+            headers: {
+              Accept: "application/json",
+              "Content-Type": "application/json",
+              "X-CSRFToken": csrf_token,
+            },
+          }).then(async (response) => {
+            if (response.status === 400) {
```
shtrom: This should be >=400 && <500, to catch all non-retryable errors.

Or perhaps we could swap the logic, and have the `===200` early-exit happy path first, and anything else as the unhappy path.

```
if (response.status === 200) {
 ...
} else {
 ...
}
```

That said, if the server errors out with something like a 500, it may be worth letting the user retry without getting rid of all their edits.

shtrom (changes requested):
A few nits, and one remaining thing regarding the handling of the status from the fetch to the API (return early on 200, and handle all 4xx).

This looks pretty good overall, though. Nice job!

zzzeid (approved):

src/lando/api/tests/test_views.py
```diff
@@ -6,6 +6,22 @@
 from lando.main.models import Repo, SCMType
 
 
+@pytest.fixture
+def repo_github_api_client(repo_mc):
+    repo_mc(SCMType.GIT, name="git-repo")
```
shtrom, line 11: nit: it may be safer to name the repo after the fixture it comes from, to avoid risks of collisions

shtrom (approved):
:shipit:

zzzeid (approved):

PR #1163, opened 2026-05-14
ci: run build and tests on pushes to `main` (Bug 2039679)

cgsheeh:


zzzeid (approved):
This is probably more relevant now that we use Lando to do merges.

PR #1175, opened 2026-05-20
tests: add `test_migrations.py` to check migration integrity (Bug 2041054)

cgsheeh:
Add `test_migrations.py` which contains a single test. This test
creates a `MigrationLoader` with the `MIGRATION_MODULES` setting
overridden to empty, causing the loader to search for migrations
files directly. We then run `detect_conflicts()` and assert no
conflicts were returned.

### DIFF
--- a/src/lando/tests/test_migrations.py
+++ b/src/lando/tests/test_migrations.py
@@ -1,0 +1,18 @@
+from django.db.migrations.loader import MigrationLoader
+from django.test import override_settings
+
+
+def test_no_migration_conflicts():
+    """Each app's migration graph must have exactly one leaf node."""
+    # `connection=None` skips the DB. `MIGRATION_MODULES={}` defeats
+    # `pytest-django`'s `--no-migrations` patch, which would otherwise
+    # hide every migration from the loader once `django_db_setup` runs.
+    with override_settings(MIGRATION_MODULES={}):
+        loader = MigrationLoader(connection=None, ignore_no_migrations=True)
+        conflicts = loader.detect_conflicts()
+
+    assert not conflicts, (
+        "Conflicting migrations detected (multiple leaf nodes per app): "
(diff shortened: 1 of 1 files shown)


src/lando/tests/test_migrations.py
```diff
@@ -0,0 +1,18 @@
+from django.db.migrations.loader import MigrationLoader
+from django.test import override_settings
+
+
+def test_no_migration_conflicts():
+    """Each app's migration graph must have exactly one leaf node."""
```
zzzeid:
```suggestion
    """Test that each app's migration graph has exactly one leaf node."""
```

Though we can make this more human readable: check that new migrations do not cause any conflicts with existing migrations." This should never happen for existing migrations.

src/lando/tests/test_migrations.py
```diff
@@ -0,0 +1,18 @@
...
+    # `connection=None` skips the DB. `MIGRATION_MODULES={}` defeats
+    # `pytest-django`'s `--no-migrations` patch, which would otherwise
+    # hide every migration from the loader once `django_db_setup` runs.
+    with override_settings(MIGRATION_MODULES={}):
+        loader = MigrationLoader(connection=None, ignore_no_migrations=True)
+        conflicts = loader.detect_conflicts()
+
+    assert not conflicts, (
+        "Conflicting migrations detected (multiple leaf nodes per app): "
+        f"{conflicts}. Renumber/rebase the offending migration or run "
```
zzzeid: I don't think "renumber/rebase" makes sense here in the context of migrations. Also, there are multiple scenarios which can cause this outcome (e.g., multiple branches linked together, outdated branch vs. base branch, etc.). I would suggest saying "check any new migration files that may need to be regenerated".

zzzeid (approved):
Test looks good though I would change the wording as suggested to make it more accurate / understandable.

PR #1176, opened 2026-05-20
repo: create srcdir-specific `MOZBUILD_STATE_PATH` and pass to `mach` commands (Bug 2029056)

cgsheeh:
Define `MOZBUILD_STATE_PATH` as a srcdir-specific path adjacent to
the working copy, generated by the `Repo` instance. Send this
path into the environment when calling `mach` commands from
within Lando. This will allow us to run `mach bootstrap` (and
possibly `mach artifact` in the future) to install srcdir-specific 
versions of formatting toolchains, to run autoformatting on
more branches than just autoland.


src/lando/main/scm/abstract_scm.py
```diff
@@ -19,8 +19,13 @@ class AbstractSCM(ABC):
...
     path: str
 
-    def __init__(self, path: str, **kwargs):
+    # The path to use as `MOZBUILD_STATE_PATH` when running `mach` commands against
+    # this repo. `None` means no per-checkout state path has been configured.
+    mozbuild_state_path: Path | None
+
+    def __init__(self, path: str, mozbuild_state_path: Path | None = None, **kwargs):
         self.path = path
+        self.mozbuild_state_path = mozbuild_state_path
```
shtrom: This feels more like a Repo thing that something that the SCM needs to know.

src/lando/main/models/repo.py
```diff
@@ -286,9 +286,18 @@ def scm(self) -> AbstractSCM:
         return self._scm
 
     def get_system_path(self) -> str:
-        """Calculate system path based on REPO_ROOT and repository name."""
+        """Calculate system path based on `REPO_ROOT` and repository name."""
         return str(Path(settings.REPO_ROOT) / self.name)
 
+    def get_mozbuild_state_path(self) -> Path:
```
shtrom: nit: this could be a property

We do have `get_system_path` above, but here we return a `Path` (which is an improvement) rather than an `str`, so we are breaking the convention a bit anyway.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -422,7 +423,7 @@ def apply_autoformatting(
         return replacements
 
     def format_stack(
-        self, landoini_config: configparser.ConfigParser, repo_path: str
+        self, landoini_config: configparser.ConfigParser, scm: AbstractSCM
```
shtrom: nit: My preference would be to not pass complex objects down to functions that don't use them.

I see why it's done here, as we otherwise would need to pull the Repo object down.

Maybe we should be more logic-agnostic by keeping the `repo_path` as is, and passing down a dict of environment_variables that would contain `{"MOZBUILD_STATE_PATH": repo.get_mozbuild_state_path()}`

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -446,23 +447,34 @@ def format_stack(
...
         """
         return self.run_mach_command(
-            path, ["format", "--fix", "--outgoing", "--verbose", "--skip-android"]
+            scm, ["format", "--fix", "--outgoing", "--verbose", "--skip-android"]
         )
 
-    def run_mach_command(self, path: str, args: list[str]) -> str:
-        """Run a command using the local `mach`, raising if it is missing."""
-        if not self.mach_path(path):
+    def run_mach_command(self, scm: AbstractSCM, args: list[str]) -> str:
```
shtrom: nit (cont'd): it would definitely make sense to have an optional dict of environment variables here.

shtrom (commented):
Passing the MOZBUILD_STATE_PATH via the SCM seems a bit smelly to me.

Could we instead pass an explicit dict of env variables from the LandingWorker which could build them from info on the Repo?

I won't block hard on this if you don't, but if it's a relatively easy change, I think it would be an improvement.

WDYT?

src/lando/main/models/repo.py
```diff
@@ -286,9 +286,18 @@ def scm(self) -> AbstractSCM:
...
+        """Calculate the per-checkout `MOZBUILD_STATE_PATH` for this repo.
+
+        Lives alongside the checkout under `REPO_ROOT` so that `mach` toolchains
+        and bootstrapped state are isolated per repo rather than sharing the
+        worker's homedir.
+        """
+        return Path(settings.REPO_ROOT) / "srcdir-mozbuilds" / self.name
```
zzzeid: It would be better to keep `REPO_ROOT` for repo clones only (which the name implies). I would suggest creating a different path variable to house this new directory (e.g., `settings.MOZBUILDS_ROOT` or something similar).

Also `"srcdir-mozbuilds"` should be defined as a variable elsewhere, but this would be resolved by having it in `settings.MOZBUILDS_ROOT`.

src/lando/api/tests/test_landings.py
```diff
@@ -1184,6 +1185,29 @@ def test_format_stack_success_changed(
...
+    mozbuild_dir = tmp_path / "srcdir-mozbuilds" / "test-repo"
+
+    # `mach` echoes `$MOZBUILD_STATE_PATH` so we can verify it was exported.
+    mach_file = tmp_path / "mach"
+    mach_file.write_text('#!/bin/sh\necho "$MOZBUILD_STATE_PATH"\n')
+    mach_file.chmod(0o755)
+
+    scm = GitSCM(str(tmp_path), mozbuild_state_path=mozbuild_dir)
+
+    output = hg_landing_worker.run_mach_command(scm, [])
```
zzzeid: Why is this using `hg_landing_worker` instead of `git_landing_worker`? If it's needed for both then this test should be parametrized, but it's out of place here given that we're using `GitSCM` above.

src/lando/main/models/repo.py
zzzeid: I agree, this should be a simple property.

Also, we are breaking the convention (which is not ideal). This should return a string (but we can still use `Path` to determine the value).

src/lando/main/tests/test_models.py
```diff
@@ -348,6 +349,31 @@ def test_repo_default_branch_to_scm(branch: str, expected_branch: str):
...
 
+def test__models__Repo__get_mozbuild_state_path():
+    """`get_mozbuild_state_path` should return a per-repo subdir of `REPO_ROOT`."""
+    repo = Repo(name="firefox-autoland")
+
+    expected = Path(settings.REPO_ROOT) / "srcdir-mozbuilds" / "firefox-autoland"
+    assert repo.get_mozbuild_state_path() == expected, (
+        "`get_mozbuild_state_path` should return "
+        "`<REPO_ROOT>/srcdir-mozbuilds/<name>` as a `Path`."
+    )
```
zzzeid: If we continue with the convention of using a string here instead of a Path object, the assert itself will tell you that the strings are different and where they are different.

nit: This is a bit redundant/cluttered in this case. I don't see the value in adding a verbose string explaining the outcome, it's already stored in the `expected` value.

src/lando/main/tests/test_models.py
```diff
@@ -348,6 +349,31 @@ def test_repo_default_branch_to_scm(branch: str, expected_branch: str):
...
+
+
+def test__models__Repo__scm_propagates_mozbuild_state_path():
```
zzzeid: This test should be written as a unit test since it is re-testing what `test__models__Repo__get_mozbuild_state_path` is testing along with the small addition . Instead, `Repo` could be mocked to return the expected result, and ensure that `scm` will pick it up.

However, I do agree with Olivier that this doesn't belong in the abstracted scm.

src/lando/api/legacy/workers/landing_worker.py
zzzeid: > nit: My preference would be to not pass complex objects down to functions that don't use them.

I agree with this. There are some cases where it makes sense to do this, but it doesn't pass that threshold in this case.

zzzeid (changes requested):
Feels like this change should be decoupled from the the abstract SCM, so an overall comment would be to try and do that throughout this PR.

src/lando/main/models/repo.py
cgsheeh: Made it a property and reverted to the `str` convention.

src/lando/api/legacy/workers/landing_worker.py
cgsheeh: Switched back to `repo_path` and added an optional dict of env-vars.

src/lando/main/models/repo.py
cgsheeh: Added `MOZBUILDS_ROOT` settings variable.

src/lando/api/tests/test_landings.py
cgsheeh: Defaulted to `hg` due to bug 2037553, but switching back worked as expected. I'll fix that one shortly.

shtrom (approved):
lgtm!

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -447,34 +452,40 @@ def format_stack(
...
 
-        env = os.environ.copy()
-        if scm.mozbuild_state_path:
-            scm.mozbuild_state_path.mkdir(parents=True, exist_ok=True)
-            env["MOZBUILD_STATE_PATH"] = str(scm.mozbuild_state_path)
+        subprocess_env = os.environ.copy()
+        if env:
+            subprocess_env.update(env)
+            if mozbuild_state_path := env.get("MOZBUILD_STATE_PATH"):
+                Path(mozbuild_state_path).mkdir(parents=True, exist_ok=True)
```
zzzeid, line 490: nit: could roll this into one line to reduce nesting.

```python
subprocess_env.update(env or {})

if env and mozbuild_state_path := env.get("MOZBUILD_STATE_PATH"):
    ...
```

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -545,7 +556,11 @@ def bootstrap_repos(self):
 
         for repo in repos:
             try:
-                self.run_mach_command(repo.scm, command)
+                self.run_mach_command(
+                    repo.path,
+                    command,
+                    env={"MOZBUILD_STATE_PATH": repo.mozbuild_state_path},
```
zzzeid: nit: maybe this should be called `extra_env` to clarify that the dict only contains additional items and not a full env.

zzzeid (approved):
lgtm

src/lando/api/legacy/workers/landing_worker.py
shtrom: second that!

cgsheeh: Agreed, done!

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -482,10 +483,10 @@ def run_mach_command(
...
 
         subprocess_env = os.environ.copy()
-        if env:
-            subprocess_env.update(env)
-            if mozbuild_state_path := env.get("MOZBUILD_STATE_PATH"):
-                Path(mozbuild_state_path).mkdir(parents=True, exist_ok=True)
+        subprocess_env.update(extra_env or {})
+
+        if extra_env and (mozbuild_state_path := extra_env.get("MOZBUILD_STATE_PATH")):
+            Path(mozbuild_state_path).mkdir(parents=True, exist_ok=True)
```
shtrom, line 489: nit: To be more resilient, we should probably use the full `subprocess_env` here, rather than just `extra_env`.

```suggestion
        if mozbuild_state_path := subprocess_env.get("MOZBUILD_STATE_PATH"):
            Path(mozbuild_state_path).mkdir(parents=True, exist_ok=True)
```

shtrom (approved):
r+ w/ nit-fix suggestion

PR #1177, opened 2026-05-20
style: format Javascript and CSS with Prettier (Bug 1995833)

cgsheeh:
Add Prettier to the Dockerfile and run over our `static_src`
directory. Add runs of `prettier` to `make format` and the
`test_formatting.py` test file.


shtrom (approved):
Lovely!

Dockerfile
```diff
@@ -51,6 +51,9 @@ RUN pip install -r /code/requirements.txt
 COPY package.json package-lock.json /deps/
 RUN npm install --prefix /deps
 
+# Add node_modules to PATH so `prettier` can be run directly.
+ENV PATH="/deps/node_modules/.bin:${PATH}"
```
zzzeid, line 55: Out of scope for this PR, but we shouldn't necessarily have these node packages installed in all production builds, as they are not needed during runtime.

I'll file a bug for this.

package.json
```diff
@@ -6,6 +6,9 @@
     "bulma": "^1.0.4",
     "sass": "^1.97.0"
   },
+  "devDependencies": {
+    "prettier": "^3.4.2"
+  },
```
zzzeid, line 16: Is this actually doing anything? We're not using `npm install --production` as far as I can tell.

zzzeid (changes requested):
lgtm other than the potentially redundant `devDependencies`.

package.json
cgsheeh: At the moment, it doesn't do anything. However this is semantically correct, and it will make splitting out our dev/production resources easier in the follow-up as we can run with `--omit=dev` or equivalent.

Edit: one thing to note is that `bulma` and `sass` aren't runtime dependencies for Lando, but they are build-time dependencies - still used to produce the production image, as opposed to Prettier which is a dev-only tool.

Dockerfile
cgsheeh: Agreed, we should be able to run a very slim image for Lando. We could add a `production` layer to our Dockerfile and move all the production dependencies into it, leaving toolchains (`rust`, `node`, `prettier`, etc) behind. Then we can have `docker-compose.yml` use the appropriate Docker image target for each task.

Could you link the bug here for reference?

package.json
zzzeid: > At the moment, it doesn't do anything. However this is semantically correct, and it will make splitting out our dev/production resources easier in the follow-up as we can run with `--omit=dev` or equivalent.

Can you add a comment here noting that this currently does not do anything? In case someone ends up stumbling on this code thinking it is not part of the production build.


> Edit: one thing to note is that `bulma` and `sass` aren't runtime dependencies for Lando, but they are build-time dependencies - still used to produce the production image, as opposed to Prettier which is a dev-only tool.

Yes, indeed the reason I said "runtime" is this. Although strictly speaking, they are currently a runtime dependency since they're actually done during a deployment, and not during build (see https://bugzilla.mozilla.org/show_bug.cgi?id=2022849).

Dockerfile
zzzeid: Sure, here it is: https://bugzilla.mozilla.org/show_bug.cgi?id=2041305.

src/lando/settings.py
```diff
@@ -257,6 +257,7 @@
 ]
 
 LINT_PATHS = (BASE_DIR,)
+PRETTIER_PATHS = (BASE_DIR / "static_src",)
```
zzzeid:
```suggestion
# Currently used by Prettier for linting CSS/JS
STATIC_LINT_PATHS = (BASE_DIR / "static_src",)
```

.prettierignore
zzzeid, line 1: Is this necessary given the paths provided in the settings module? I'd rather we were explicit there, vs. having to add this file.

package.json
cgsheeh: Ah, I didn't know that about the CSS compile step happening during the deploy, thanks for the heads up.

Added a comment indicating this.

.prettierignore
cgsheeh: Makes sense, removed this file.

cgsheeh: After removing `.prettierignore`, the `/vendor/` directory is reformatted. I tried adding a shim for this but prettier doesn't have an option to avoid reformatting specific directories via the CLI. Pruned `.prettierignore` to just ignore `/vendor/` and re-added, let me know what you think.

zzzeid: Maybe this is a good sign that we need to move `vendor` to its own directory. How about we move it to `static_vendor` and add that to `settings.STATICFILES_DIRS`? Should be a straightforward change. Thoughts?

zzzeid: We could also move `vendor` up to the root of static_src, and set STATIC_LINT_PATHS to `assets`, `css`, and `legacy`. Either should work.

zzzeid (approved):
lgtm, small nit would be to avoid the .prettierignore altogether by moving the `vendor` directory.

.prettierignore
cgsheeh: Even better would be to remove the `vendored/` directory altogether and install `jquery`/`fontawesome` from NPM. Filed [Bug 2041459](https://bugzilla.mozilla.org/show_bug.cgi?id=2041459) for this.

PR #1178, opened 2026-05-21
setup_dev: various improvements (bug 2041100)

zzzeid:
- add api token to admin user in setup_dev
- add automation job permissions to user
- move common hashing/prefix logic to separate method
- also update conduit-admin group permissions with all permissions
- refactor local fail-safe

### DIFF
--- a/src/lando/utils/management/commands/setup_dev.py
+++ b/src/lando/utils/management/commands/setup_dev.py
@@ -1,9 +1,11 @@
 from django.conf import settings
-from django.contrib.auth.models import Group, User
+from django.contrib.auth.hashers import make_password
+from django.contrib.auth.models import Group, Permission, User
 from django.core.management import call_command
 from django.core.management.base import BaseCommand, CommandError
 
 from lando.environments import Environment
+from lando.headless_api.models.tokens import ApiToken
 from lando.main.models import Repo, Worker
 from lando.main.models.worker import WorkerType
 from lando.main.scm import SCMType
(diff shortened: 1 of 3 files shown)


src/lando/utils/management/commands/setup_dev.py
```diff
@@ -71,11 +73,24 @@ def setup_users(self):
...
+                "admin", password="password", email="test@example.org"
+            )
             self.stdout.write(f"Admin user ({user}) created.")
         user.is_staff = True
         user.save()
+        add_automationjob = Permission.objects.get(codename="add_automationjob")
+        user.user_permissions.add(add_automationjob)
         Group.objects.get(name=CONDUIT_ADMIN_GROUP_NAME).user_set.add(user)
+        token = "a" * 128
+        ApiToken.objects.create(
```
cgsheeh, line 92: We should use `ApiToken.create_token` here instead, so we don't duplicate the token creation/prefixing/hashing logic.

If we need to use pre-defined token, we should add an optional kwarg where we can pass the token in.

cgsheeh (changes requested):

src/lando/utils/management/commands/setup_dev.py
zzzeid: Generally I would agree with this, but in this case I'm hesitant to add any functionality into production code that is used exclusively for testing. This would add clutter that is never used in production (best case), or in the worst case, make it possible to do something that should not be done (e.g., create a token that is not so secret, or in ways that could end up getting it leaked).

Even though the implementation is nicer and simpler, in this case I don't think we should go with this DRY approach (which is why I chose this approach in the first place).

zzzeid: I added a more DRY implementation, without affecting the original `create_token_method`, so at least we are not duplicating the prefixing/hashing logic. `ApiToken.create_token` will still always generate a secure token.

shtrom (approved):
lgtm!

cgsheeh (approved):

PR #1179, opened 2026-05-21
vendor: remove in-tree `jquery`/`fontawesome` and install via npm (Bug 2041459)

cgsheeh:
Remove the vendored copies of `jquery` and `fontawesome` from
the repo and install via `npm` instead.


shtrom (approved):
lgtm!

src/lando/settings.py
```diff
@@ -197,6 +197,8 @@
 # Directories to include in static file collection.
 STATICFILES_DIRS = [
     BASE_DIR / "static_src",
+    ("font-awesome", "/deps/node_modules/font-awesome"),
+    ("jquery", "/deps/node_modules/jquery/dist"),
```
zzzeid: We should avoid hard-linking into the docker image from a settings variable like this.

We can create copies into a specified and known directory and then link it here instead (e.g., `BASE_DIR / "static_vendor"`).

zzzeid (changes requested):

src/lando/settings.py
cgsheeh: Makes sense to me. Moved to a separate `/static_vendored` dir instead of under `BASE_DIR`, to avoid unnecessary Docker image rebuilds of the vendored files after every code change.

zzzeid (approved):
This is good for now, but this implementation will also be affected by [bug 2019284](https://bugzilla.mozilla.org/show_bug.cgi?id=2019284)) and other dockerfile cleanup, since ultimately most if not all of the static files functionality does not need to exist in the main docker image.

PR #1180, opened 2026-05-23
tests: use mocked Git repos where possible (Bug 2037553)

cgsheeh:
Add Git repos to `mocked_repo_config`, mirroring the setup
of repos in production. Update tests which were hg-only to
include Git coverage as well.

### DIFF
--- a/src/lando/api/tests/test_transplants.py
+++ b/src/lando/api/tests/test_transplants.py
@@ -709,8 +709,13 @@ def test_integrated_transplant_simple_partial_stack_saves_data_in_db(
     assert job.landed_phabricator_revisions == {1: 1, 2: 2}
 
 
+@pytest.mark.parametrize(
+    "scm_type",
+    [SCMType.HG, SCMType.GIT],
+)
 @pytest.mark.django_db
 def test_integrated_transplant_records_approvers_peers_and_owners(
+    scm_type,
     user,
     authenticated_client,
     treestatusdouble,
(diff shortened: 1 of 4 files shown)


src/lando/api/tests/test_transplants.py
```diff
@@ -746,11 +744,11 @@ def test_integrated_transplant_records_approvers_peers_and_owners(
     reviewer = phabdouble.user(username="reviewer")
     user2 = phabdouble.user(username="reviewer2")
 
-    d1 = phabdouble.diff(rawdiff=normal_patch(1))
+    d1 = phabdouble.diff(rawdiff=normal_patch(0))
     r1 = phabdouble.revision(diff=d1, repo=phabrepo)
     phabdouble.reviewer(r1, reviewer)
 
-    d2 = phabdouble.diff(rawdiff=normal_patch(2))
+    d2 = phabdouble.diff(rawdiff=normal_patch(1))
```
shtrom, line 757: Curious about this change.

shtrom (approved):
r+ with one question

zzzeid (changes requested):
I think the changes in conftest should be made exclusively of the rest of the changes here (i.e., add git repos to mocked config). This PR somewhat alters the coverage, rather than increases it. For some of the existing tests (mainly `src/lando/api/tests/test_transplants.py`, `src/lando/api/tests/test_landings.py`, in the scope of this PR) it's probably better to parametrize them instead of switching them over to git (recall that we are still using hg for certain repos).

cgsheeh:
> I think the changes in conftest should be made exclusively of the rest of the changes here (i.e., add git repos to mocked config). This PR somewhat alters the coverage, rather than increases it. For some of the existing tests (mainly `src/lando/api/tests/test_transplants.py`, `src/lando/api/tests/test_landings.py`, in the scope of this PR) it's probably better to parametrize them instead of switching them over to git (recall that we are still using hg for certain repos).

Switched to adding coverage for both Mercurial and Git, instead of moving to just Git.

src/lando/api/tests/test_transplants.py
cgsheeh: The test hg repo already has patch 0 applied, where the git repo doesn't.

shtrom: Maybe we should apply it to the git repo, too?

This may spill over to a number of other Git-only tests that would need fixing, though \\:

zzzeid (approved):

PR #1195, opened 2026-05-26
main.models: implement CryptographyMixin (bug 2042216)

zzzeid:
- add CryptographyMixin to provide helpers to facilitate encrypted fields
- adding this mixin provides setters/getters for fields beginning with `encrypted_`
- also addresses bug 2042218

### DIFF
--- a/src/lando/main/models/base.py
+++ b/src/lando/main/models/base.py
@@ -1,7 +1,10 @@
 import logging
 from contextlib import ContextDecorator
+from typing import Callable
 
 from django.db import connection, models, transaction
+
+from lando.utils.crypto import ENCRYPTED_FIELD_PREFIX, FUNC_PREFIXES, cryptography
 
 logger = logging.getLogger(__name__)
 
@@ -25,6 +28,106 @@ class LockTableContextManager(ContextDecorator):
 
     def __exit__(self, exc_type, exc_value, traceback):  # noqa: ANN001
         pass
+
+
+class CryptographyMixin:
+    """
+    Provide a more convenient way of encrypting and decrypting fields.
+
+    String-based fields that are prefixed with `encrypted_` will have this feature enabled.
+    For example, a field such as MyModel.encrypted_secret_name will provide the following:
+    - MyModel.secret_name: The decrypted value of MyModel.encrypted_secret_name
+    - MyModel.set_secret_name: Encrypt and set the value of MyModel.encrypted_secret_name
+    - MyModel.clear_secret_name: Clear and encrypt the value of MyModel.encrypted_secret_name
+    - MyModel.rotate_secret_name: Use when rotating the secret keys used to encrypt values.
+    """
+
+    def __getattr__(self, name: str) -> object:
+        """Redirect to appropriate function call."""
+        obj_dir = object.__dir__(self)
+        encrypted_fields = tuple(
+            field.removeprefix(ENCRYPTED_FIELD_PREFIX)
+            for field in obj_dir
+            if field.startswith(ENCRYPTED_FIELD_PREFIX)
+        )
+        crypto_methods = [
+            f"{prefix}{field_name}"
+            for prefix in FUNC_PREFIXES
+            for field_name in encrypted_fields
+        ]
+
+        if name in crypto_methods:
+            method, field_name = name.split("_", 1)
+            return object.__getattribute__(self, f"get_{method}_encrypted_field")(
+                field_name
+            )
+
+        if f"{ENCRYPTED_FIELD_PREFIX}{name}" in obj_dir:
+            # The request is for the decrypted value of the field.
+            return self._get_decrypted_value_or_empty(name)
+
+        return object.__getattribute__(self, name)
+
+    def _get_decrypted_value_or_empty(self, field_name: str) -> str:
+        """Return decrypted value, or an empty string if there is no value."""
+        encrypted_value = bytes(getattr(self, f"{ENCRYPTED_FIELD_PREFIX}{field_name}"))
+        if encrypted_value:
+            return self._decrypt_value(encrypted_value)
+        else:
+            return ""
+
+    def _encrypt_value(self, value: str) -> bytes:
+        """Encrypt a given string value."""
+        return cryptography.encrypt(value.encode("utf-8"))
+
+    def _decrypt_value(self, value: bytes) -> str:
+        """Decrypt a given bytes value."""
+        return cryptography.decrypt(value).decode("utf-8")
+
+    def _rotate_value(self, value: bytes) -> bytes:
+        """Return a rotated encrypted bytes value."""
+        return cryptography.rotate(value)
+
+    def get_clear_encrypted_field(self, field_name: str) -> Callable:
+        """Return the corresponding method to clear the encrypted field."""
+        set_encrypted_field = self.get_set_encrypted_field(field_name)
+
+        def clear_encrypted_field(save: bool = True):
+            """Set value to encrypted blank string."""
+            set_encrypted_field("", save=save)
+
+        return clear_encrypted_field
+
+    def get_set_encrypted_field(self, field_name: str) -> Callable:
+        """Return the corresponding method to set the encrypted field."""
+
+        def set_encrypted_field(value: str, save: bool = True):
+            """Set encrypted value to given field and save if needed."""
+            setattr(
+                self,
+                f"{ENCRYPTED_FIELD_PREFIX}{field_name}",
+                self._encrypt_value(value),
+            )
+            if save:
+                self.save()
+
+        return set_encrypted_field
+
+    def get_rotate_encrypted_field(self, field_name: str) -> Callable:
+        """Return the corresponding method to rotate the encrypted field."""
+
+        def rotate_encrypted_field():
+            """Rotate encryption on given field."""
+            setattr(
+                self,
+                f"{ENCRYPTED_FIELD_PREFIX}{field_name}",
+                self._rotate_value(
+                    bytes(getattr(self, f"{ENCRYPTED_FIELD_PREFIX}{field_name}"))
+                ),
+            )
+            self.save()
+
+        return rotate_encrypted_field
 
 
 class BaseModel(models.Model):
--- a/src/lando/main/models/profile.py
+++ b/src/lando/main/models/profile.py
@@ -1,10 +1,9 @@
-from cryptography.fernet import Fernet, MultiFernet
 from django.conf import settings
 from django.contrib.auth.models import Permission, User
 from django.contrib.contenttypes.models import ContentType
-from django.db import models
+from django.db import models, transaction
 
-from lando.main.models.base import BaseModel
+from lando.main.models.base import BaseModel, CryptographyMixin
 
 SCM_PERMISSIONS = (
     ("scm_allow_direct_push", "SCM_ALLOW_DIRECT_PUSH"),
@@ -62,16 +61,13 @@ def filter_claims(claims: dict) -> dict:
     return claims
 
 
-class Profile(BaseModel):
+class Profile(CryptographyMixin, BaseModel):
     """A model to store additional information about users."""
 
     class Meta:
         permissions = SCM_PERMISSIONS + (
             ("can_view_private_repos", "CAN_VIEW_PRIVATE_REPOS"),
         )
-
-    # Provide encryption/decryption functionality.
-    cryptography = MultiFernet([Fernet(key) for key in settings.ENCRYPTION_KEYS])
 
     user = models.OneToOneField(User, null=True, on_delete=models.SET_NULL)
 
@@ -86,18 +82,6 @@ class Profile(BaseModel):
     phabricator_phid = models.CharField(
         max_length=255, null=True, blank=True, unique=True
     )
-
-    def _encrypt_value(self, value: str) -> bytes:
-        """Encrypt a given string value."""
-        return self.cryptography.encrypt(value.encode("utf-8"))
-
-    def _decrypt_value(self, value: bytes) -> str:
-        """Decrypt a given bytes value."""
-        return self.cryptography.decrypt(value).decode("utf-8")
-
-    def _rotate_value(self, value: bytes) -> bytes:
-        """Return a rotated encrypted bytes value."""
-        return self.cryptography.rotate(value)
 
     def _has_scm_permission_groups(self, codename: str, groups: list[str]) -> bool:
         """Return whether the group membership provides the correct permission.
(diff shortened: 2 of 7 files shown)


src/lando/main/models/base.py
```diff
@@ -27,6 +29,105 @@ def __exit__(self, exc_type, exc_value, traceback):  # noqa: ANN001
...
+    String-based fields that are prefixed with `encrypted_` will have this feature enabled.
+    For example, a field such as MyModel.encrypted_secret_name will provide the following:
+    - MyModel.secret_name: The decrypted value of MyModel.encrypted_secret_name
+    - MyModel.set_secret_name: Encrypt and set the value of MyModel.encrypted_secret_name
+    - MyModel.clear_secret_name: Clear and encrypt the value of MyModel.encrypted_secret_name
+    - MyModel.rotate_secret_name: Use when rotating the secret keys used to encrypt values.
+    """
+
+    # Provide encryption/decryption functionality.
+    cryptography = MultiFernet([Fernet(key) for key in settings.ENCRYPTION_KEYS])
```
cgsheeh: This feels like it should go in the `settings` or `auth` modules, instead of being attached to every model that needs encryption.

src/lando/main/models/base.py
```diff
@@ -27,6 +29,105 @@ def __exit__(self, exc_type, exc_value, traceback):  # noqa: ANN001
...
+SAVE_FIELD_PREFIX = "save"
+ROTATE_FIELD_PREFIX = "rotate"
+FUNC_PREFIXES = (
+    CLEAR_FIELD_PREFIX,
+    SAVE_FIELD_PREFIX,
+    ROTATE_FIELD_PREFIX,
+)
+
+
+class CryptographyMixin:
```
cgsheeh, line 33: This works, but the mixin feels like the wrong layer to me: it's a lot of magic on the class for something that's really about an individual field.

If we had something like a custom `EncryptedBinaryField` which extended or was built from a `models.BinaryField`, we could do:

```python
class Profile(BaseModel):
    phabricator_api_key = EncryptedBinaryField()
    ...
    # Clear:
    self.phabricator_api_key = None
    # Set unencrypted, encrypts under the hood:
    self.phabricator_api_key = b"api-asdfasdf..."
    # Rotate:
    self.phabricator_api_key.rotate()
    # Save is explicit:
    self.save()
```

This gives us all of the same functionality with a lot less dynamic "magic", LSP support for the various fields, and makes it easier to reason about the code.

WDYT?

src/lando/main/tests/__init__.py
```diff
@@ -102,3 +113,52 @@ def test_phabricator_api_key_encryption():
...
+
+def test_cryptography_mixin__clear_field(mock_crypto_model):
+    mock_crypto_model.encrypted_made_up_field = None
+    mock_crypto_model.clear_made_up_field()
+    assert mock_crypto_model.cryptography.encrypt.call_count == 1
+    assert mock_crypto_model.cryptography.encrypt.call_args[0] == (b"",)
+    assert mock_crypto_model.save.call_count == 1
+
+
+def test_cryptography_mixin__rotate_field(mock_crypto_model):
```
shtrom: We could probably do with a less mocked test that goes into the real `cryptography` code, and make sure the `made_up_field` value survives through encryption and rotation.

shtrom (commented):
Looks good in principle, though a field-type, as @cgsheeh suggests, feels like it would be a cleaner fit.

src/lando/main/models/base.py
zzzeid: >it's a lot of magic on the class for something that's really about an individual field.

The mixin does provide this functionality for an arbitrary number of fields on the same model, even though it's only for a single field at this time per model.

I do think the idea of a custom field class for encrypted fields is good, though the approach in this PR takes our existing implementation for the Phabricator API key and makes it available for other fields. It doesn't add any new logic or implementation details with respect to the encryption (i.e., it's backwards compatible in the case of `encrypted_phabricator_api_key`), and the mixin can be removed at any time if we wanted to change the implementation. It's basically in lieu of copying and pasting the existing functionality into a new field.

I think it's acceptable as a less-hacky and more DRY way of having new encrypted fields that use the existing encryption methods, and keeping them consistent across the board, until we decide to implement a custom field class.

Thoughts?

src/lando/main/models/base.py
zzzeid: Yep, this is a relic of moving the existing functionality from the Profile model, will fix.

src/lando/main/tests/__init__.py
zzzeid: As this is a unit test, I think we can assume that the cryptography module is well tested and behaves as expected.

In terms of testing the real `cryptography` code in the context of these fields (more of an integration test), the coverage already exists in `test_phabricator_api_key_encryption` a few lines above, as that is for an existing field in our models. It is however missing coverage for rotation, so I can add it there or in a separate, similar test. Thoughts?

zzzeid: Modified existing test to test for key rotation.

src/lando/main/models/base.py
zzzeid: Moved to own module.

src/lando/main/tests/__init__.py
zzzeid: Let me know if this new coverage is good enough (added to `test_phabricator_api_key_encryption`).

src/lando/main/models/base.py
zzzeid: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2045046.

cgsheeh: Gotcha, that sounds reasonable to me!

src/lando/main/models/base.py
```diff
@@ -27,6 +29,92 @@ def __exit__(self, exc_type, exc_value, traceback):  # noqa: ANN001
...
+    For example, a field such as MyModel.encrypted_secret_name will provide the following:
+    - MyModel.secret_name: The decrypted value of MyModel.encrypted_secret_name
+    - MyModel.set_secret_name: Encrypt and set the value of MyModel.encrypted_secret_name
+    - MyModel.clear_secret_name: Clear and encrypt the value of MyModel.encrypted_secret_name
+    - MyModel.rotate_secret_name: Use when rotating the secret keys used to encrypt values.
+    """
+
+    def __getattr__(self, name: str) -> object:
+        """Redirect to appropriate function call."""
+        if name.startswith(FUNC_PREFIXES):
```
cgsheeh: The prefixes should have `_` appended to the end of them, or we should add the `_` suffix here. As it stands now a method like `settings()` would trigger the `set` prefix, which isn't the intention.

src/lando/main/models/base.py
```diff
@@ -27,6 +29,92 @@ def __exit__(self, exc_type, exc_value, traceback):  # noqa: ANN001
...
+
+    def _decrypt_value(self, value: bytes) -> str:
+        """Decrypt a given bytes value."""
+        return cryptography.decrypt(value).decode("utf-8")
+
+    def _rotate_value(self, value: bytes) -> bytes:
+        """Return a rotated encrypted bytes value."""
+        return cryptography.rotate(value)
+
+    def get_clear_encrypted_field(self, field_name: str) -> callable:
```
cgsheeh: We need to use `typing.Callable` here, as the builtin `callable` isn't a valid type hint. Unlike `list` or `tuple` which are standalone types, `callable` is just a method that returns "is this a callable object".

src/lando/main/models/base.py
```diff
@@ -27,6 +29,92 @@ def __exit__(self, exc_type, exc_value, traceback):  # noqa: ANN001
...
+                self,
+                f"{ENCRYPTED_FIELD_PREFIX}_{field_name}",
+                self._encrypt_value(value),
+            )
+            if save:
+                self.save()
+
+        return set_encrypted_field
+
+    def get_rotate_encrypted_field(self, field_name: str) -> callable:
```
cgsheeh: We should add docstrings for these methods.

src/lando/main/models/profile.py
```diff
@@ -62,17 +61,14 @@ def filter_claims(claims: dict) -> dict:
     return claims
 
 
-class Profile(BaseModel):
```
cgsheeh, line 65: Should we put the mixin first? Seems that's the usual convention, so MRO is easier to reason about.

src/lando/main/tests/__init__.py
cgsheeh, line 1: Can you put these tests in `src/lando/utils/tests/test_crypto.py` (or similar)?

Side note: we should move all the other tests in this `__init__.py` out in to their own well-named files. Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2045823 for this.

src/lando/main/tests/__init__.py
```diff
@@ -94,11 +108,67 @@ def test_phabricator_api_key_encryption():
...
     key = "test-key"
-    profile.save_phabricator_api_key(key)
+    profile.set_phabricator_api_key(key)
     assert key.encode("utf-8") not in profile.encrypted_phabricator_api_key
     assert base64.b64decode(profile.encrypted_phabricator_api_key)
     assert profile.phabricator_api_key == key
 
+    # Rotate encryption key.
+    new_key = Fernet.generate_key()
+    with override_settings(ENCRYPTION_KEYS=(new_key,)):
```
cgsheeh: The `MultiFernet` is built at import time, so I don't think this `override_settings` does anything.

src/lando/main/models/profile.py
```diff
@@ -123,31 +107,18 @@ def get_all_scm_permissions(cls) -> dict[str, Permission]:
...
+        self.phabricator_phid = None
+        self.save()
 
-    def save_phabricator_api_key(self, key: str, phid: str | None = None):
+    def save_phabricator_elements(self, key: str, phid: str | None = None):
         """Given a raw API key and PHID, encrypt the key and store both."""
-        self.encrypted_phabricator_api_key = self._encrypt_value(key)
+        self.set_phabricator_api_key(key)
         self.phabricator_phid = phid or None
         self.save()
```
cgsheeh: The `set_` methods introduced by these PRs do a save by default. We should wrap both the `clear` and `set` operations in transactions so the operations happen atomically.

cgsheeh (changes requested):

src/lando/main/models/base.py
shtrom: Good catch!

shtrom: Though this would only happen if the `settings` method doesn't exist.

src/lando/main/models/base.py
```diff
@@ -27,6 +29,92 @@ def __exit__(self, exc_type, exc_value, traceback):  # noqa: ANN001
         pass
 
 
+class CryptographyMixin:
```
shtrom, line 33: nit: this is generic enough to not be limited to models, it could go in `lando.utils.crypto`

src/lando/main/tests/__init__.py
shtrom: We could add an assertion that the ciphertext has changed to confirm this.

src/lando/main/models/base.py
shtrom: Nevermind, it's models-specific. Ignore me (:

src/lando/main/tests/__init__.py
zzzeid: >The MultiFernet is built at import time, so I don't think this override_settings does anything.

This actually reminds me why I had it in the model object originally. It might make sense to keep it there instead of putting it at the top level of a module (i.e., [this discussion](https://github.com/mozilla-conduit/lando/pull/1195/changes#discussion_r3352438048)). I'll look into that. It would be good to not require a restart for updating the

src/lando/main/models/profile.py
zzzeid: Can switch it to make it align more with conventions.

src/lando/main/models/base.py
zzzeid: Good catch. I think it's better to be more explicit about this, than to simply append a `_` (otherwise we will trigger this for anything that starts with `set_` if it does not exist (e.g., `set_nonexistent_field()`). I'll add a check that makes sure `name` is a valid one for an encrypted field, if it starts with the prefix.

zzzeid: Made this more robust by generating a list of acceptable field names. Also simplified the return flow.

src/lando/main/tests/__init__.py
zzzeid: The new tests here actually should have gone into `test_models`, moving them there.

src/lando/main/models/base.py
zzzeid: ~I'm going to move this back to the model based on [this](https://github.com/mozilla-conduit/lando/pull/1195#discussion_r3374876927) conversation.~

After trying out a few things, I realized that though this would work, it would require modifying the test to do API calls rather than test the model directly, along with other changes, rendering this change too out of scope just to implement testing key rotation. I will file a separate bug to figure out a good way to do this without altering the code too much just to accommodate a test.

zzzeid:
I'll rebase and update migrations after this is approved.

src/lando/main/tests/test_models.py
```diff
@@ -525,3 +536,80 @@ def test_hook_choices_all_checks():
...
+
+@mock.patch("lando.main.models.base.cryptography")
+def test_cryptography_mixin__rotate_field(cryptography, mock_crypto_model):
+    mock_crypto_model.encrypted_made_up_field = None
+    cryptography.encrypt.return_value = b"something encrypted"
+    mock_crypto_model.set_made_up_field("this is a test")
+    assert cryptography.encrypt.call_args[0] == (b"this is a test",)
+    assert mock_crypto_model.save.call_count == 1
+
+    mock_crypto_model.rotate_made_up_field()
```
cgsheeh, line 605: This test doesn't actually verify rotation. We should compare the encrypted values pre/post rotation to confirm they don't match, and confirm the unencrypted values pre/post rotation do match.

cgsheeh (approved):
LGTM, with a comment that the test coverage for rotation doesn't actually check correctness.

src/lando/main/tests/test_models.py
zzzeid: Addressed that in a comment [here](https://github.com/mozilla-conduit/lando/pull/1195#discussion_r3389225508).

cgsheeh: @zzzeid not sure I follow the comment, testing the field rotation shouldn't require any API calls. Can you explain in more detail?

zzzeid: @cgsheeh overriding the settings would not work in the current state of the test, since we are testing the model directly, and `Profile` is imported already with the existing settings. We would have to rewrite the test in a different way.

I was referencing the API as an example of how we would not need to import the Profile model, so we can use that to set/re-set the key but it would still not solve the rotation problem since that is not triggered via the API. So, this needs to be revisited - we have to rewrite the test so that the key rotation would propagate in an individual test.

shtrom (approved):

PR #1196, opened 2026-05-26
testing: add Vitest framework to JS dev dependencies (Bug 2042511)

cgsheeh:
Add Vitest as a development dependency to `package.json` with
`vitest.config.js` and `vitest.setup.js` config files. Implement
a test file for `formatTime` to show the framework working
in Lando. Add a `make test-js` command to run the JS tests
locally, as well as a `test-js` CI workflow. Convert `make test`
to `make test-py`, and add a new `make test` which runs both
the JS and Python tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,3 +18,12 @@ jobs:
         run: docker build -t lando .
       - name: Test
         run: make test
+
+  test-js:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build
+        run: docker build -t lando .
+      - name: Test
+        run: make test-js
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Build
         run: docker build -t lando .
       - name: Test
-        run: make test
+        run: make test-py
 
   test-js:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,8 @@ help:
 	@echo "target is one of:"
 	@echo "    help                 show this message and exit"
 	@echo "    format               run ruff and djLint on source code"
-	@echo "    test                 run the test suite"
+	@echo "    test                 run the Python test suite"
+	@echo "    test-js              run the JavaScript test suite (Jest)"
 	@echo "    migrations           generates migration files to reflect model changes in the database"
 	@echo "    upgrade-requirements upgrade packages in requirements.txt"
 	@echo "    upgrade-npm-packages update package-lock.json"
@@ -34,6 +35,10 @@ help:
 .PHONY: test
 test:
 	$(BASE_COMMAND) lando tests $(ARGS_TESTS)
+
+.PHONY: test-js
+test-js:
+	$(BASE_COMMAND) npm test
 
 .PHONY: test-use-suite
 test-use-suite:
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ help:
 	@echo "    help                 show this message and exit"
 	@echo "    format               run ruff and djLint on source code"
 	@echo "    test                 run the Python test suite"
-	@echo "    test-js              run the JavaScript test suite (Jest)"
+	@echo "    test-js              run the JavaScript test suite (Vitest)"
 	@echo "    migrations           generates migration files to reflect model changes in the database"
 	@echo "    upgrade-requirements upgrade packages in requirements.txt"
 	@echo "    upgrade-npm-packages update package-lock.json"
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,8 @@ help:
 	@echo "target is one of:"
 	@echo "    help                 show this message and exit"
 	@echo "    format               run ruff and djLint on source code"
-	@echo "    test                 run the Python test suite"
+	@echo "    test                 run the Python and JavaScript test suites"
+	@echo "    test-py              run the Python test suite"
 	@echo "    test-js              run the JavaScript test suite (Vitest)"
 	@echo "    migrations           generates migration files to reflect model changes in the database"
 	@echo "    upgrade-requirements upgrade packages in requirements.txt"
@@ -33,7 +34,10 @@ help:
 	@echo "    test-use-local       run the testsuite using the local environment"
 
 .PHONY: test
-test:
+test: test-py test-js
+
+.PHONY: test-py
+test-py:
 	$(BASE_COMMAND) lando tests $(ARGS_TESTS)
 
 .PHONY: test-js
--- a/compose.yaml
+++ b/compose.yaml
@@ -21,7 +21,7 @@ services:
     image: lando
     command: bash -c "
       lando migrate &&
-      lando collectstatic --clear --no-input &&
+      lando collectstatic --clear --no-input --ignore=__tests__ &&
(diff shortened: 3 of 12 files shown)


zzzeid (changes requested):
I think we should hold off on adding new dependencies related to our existing JS until we have a discussion and determination on what JS framework to use for the frontend (this was one of the larger items in the internship doc, but also a general need for Lando). E.g., using something like vue.js, ember.js, angular, react, backbone.js etc... Many of the JS frameworks come with their own testing library.

This is something we should do in the next few months since it is necessary to clean up and port all of our existing JS code into a cleaner, framework-based approach.

cgsheeh:
> I think we should hold off on adding new dependencies related to our existing JS until we have a discussion and determination on what JS framework to use for the frontend (this was one of the larger items in the internship doc, but also a general need for Lando). E.g., using something like vue.js, ember.js, angular, react, backbone.js etc... Many of the JS frameworks come with their own testing library.
> 
> This is something we should do in the next few months since it is necessary to clean up and port all of our existing JS code into a cleaner, framework-based approach.

A few thoughts on this:

- Jest is framework-agnostic, and several of the frameworks mentioned can be tested with Jest (or Vitest, which has an almost identical API). Test coverage we add now is likely to survive a framework migration, or at least migrate mechanically.
- In my experience, migrating to a new framework is much easier with tests in place, as we will have a safety net to port against instead of migrating blind. The tests can act as a spec for what the new framework needs to achieve.
- Should we go a different direction with the test framework, throwing this away is cheap. We simply remove the new dependencies and the config files.

Happy to defer if the framework decision is imminent, however IMO the upside to having some light test coverage outweighs the downside of waiting.

Makefile
```diff
@@ -35,6 +36,10 @@ help:
 test:
 	$(BASE_COMMAND) lando tests $(ARGS_TESTS)
 
+.PHONY: test-js
+test-js:
```
shtrom, line 44: Should we have a matching `test-py`, and a default `test: test-py test-js`?

zzzeid:
>Jest is framework-agnostic, and several of the frameworks mentioned can be tested with Jest (or Vitest, which has an almost identical API).

Maybe we should incorporate this into a design decision document: which testing frameworks were considered, and why Jest is chosen at this time, what impact adding this external dependency could have on core features development, etc... However this may not be needed given my last point below.

>Test coverage we add now is likely to survive a framework migration, or at least migrate mechanically.

I think most if not all our JS needs a full rewrite (or rather a paradigm shift) and will likely not survive the migration to a framework. Something like `formatTime.js` which was added with this PR is quite different/simpler than the rest of the JS code, e.g., in the `components` subdirectory (all hooked up to `main.js`) or anything handling manipulation of the UI. On the other hand, we don't have that much JS code anyway.

>In my experience, migrating to a new framework is much easier with tests in place, as we will have a safety net to port against instead of migrating blind. The tests can act as a spec for what the new framework needs to achieve.

Generally would agree with this, but I am not sure if the current state of our JavaScript code is amenable to this approach. There would likely need to be major rewriting of existing code before it can be tested, making it somewhat redundant if the code should be rewritten in a new framework. A plan here would be helpful to see and would make it easier to determine if this is worthwhile. 

>Should we go a different direction with the test framework, throwing this away is cheap. We simply remove the new dependencies and the config files.

Yes throwing the dependencies and configs away would be cheap, however, if we add any more tests, or have to rewrite code to be testable in this way, it would not be as cheap since that work could be redundant if have to rewrite the code.

> Happy to defer if the framework decision is imminent, however IMO the upside to having some light test coverage outweighs the downside of waiting.

The decision process is imminent (we really need that for the pull request pages, along with a number of other UI improvements already planned -- check the internship doc under "Robust UI"). I'd like to roll it as a blocker of [bug 1995754](https://bugzilla.mozilla.org/show_bug.cgi?id=1995754). The original idea I had was to use a framework for one page (as a POC) and slowly migrate things to the framework from there, rather than switching everything all at once. With that approach, I think we could add tests as we move things over.

I do think that for the JS framework we should have an ADR/decision document though, so I can start one sooner than later if that's something that's blocking you from writing new tests for new code.

cgsheeh:
Updated this PR to add Vitest instead of Jest. Vitest is a modern alternative to Jest, and it plays nicely with Vue.js which is the heavy favorite for future front-end work. The test itself migrates over with almost no changes.

compose.yaml
```diff
@@ -21,7 +21,7 @@ services:
     image: lando
     command: bash -c "
       lando migrate &&
-      lando collectstatic --clear --no-input &&
+      lando collectstatic --clear --no-input --ignore=__tests__ &&
```
zzzeid: Any way to avoid having to pass `--ignore` here? Would putting tests in a staticfiles directory outside of what `collectstatic` picks up be an option? (and updating `vitest.config.js` accordingly).

.github/workflows/build.yml
```diff
@@ -17,4 +17,13 @@ jobs:
...
+        run: make test-py
+
+  test-js:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build
+        run: docker build -t lando .
+      - name: Test
+        run: make test-js
```
zzzeid, line 29: Is there an advantage, currently, of doing this vs. running `make test` above in a single job? Or even as two separate steps in the same job? I imagine this is to make them run in parallel but the build + setup time may be longer than the actual test run time.

src/lando/static_tests/legacy/js/utils/formatTime.test.js
zzzeid, line 1: nit: would be good to add non-happy paths here (e.g., invalid input, no input, etc.)

zzzeid (approved):
lgtm, with a couple of comments/questions.

src/lando/static_tests/legacy/js/utils/formatTime.test.js
cgsheeh: Added non-happy path tests.

This made me realize that we have an uncaught exception coming from this method for certain invalid date formats. I'll fix that as a follow-up to keep this PR targeted in scope.

cgsheeh: https://bugzilla.mozilla.org/show_bug.cgi?id=2043328

.github/workflows/build.yml
cgsheeh: Technically it's slightly faster to run the tests separately. We will also see which tech stack had failures instead of having a single red/green result for all tests.

I did look at trying to move the docker image build to a separate step and then have the test jobs re-use the image, but it seemed like a pre-mature optimization for this PR.

zzzeid: Yeah I agree stack-wise it is good to separate these. I have some WIP related to [bug 2019284](https://bugzilla.mozilla.org/show_bug.cgi?id=2019284) that will also touch on this stuff. Still too rough to put up though.

compose.yaml
cgsheeh: Good call, moved the tests to a `static_tests` directory in the most recent commit and added an alias in `vitest.config.js`. Let me know what you think.

zzzeid (approved):
lgtm. I noticed that `package-lock.json` seems to be picking up a lot of updates outside the scope of packages in this PR. May be good to pin some packages more strictly to avoid this, or to add a make command to only add new packages and not upgrade existing ones (e.g., like `make add-requirements`).

PR #1199, opened 2026-05-27
Makefile: add `build` target and self-documenting logic (bug 2042614)

shtrom:
* Makefile: add build target
 * Makefile: sort targets lexicographically
 * Makefile: self-documenting Makefile
    The `help` target auto-generates an entry for each target with an inline comment introduced by a `##`.
 * Makefile: use column(1) to dynamically tabulate the help

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build
-        run: docker build -t lando .
+        run: make build
       - name: Test
         run: make test-py
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build
-        run: make build
+        run: docker build -t lando .
       - name: Test
         run: make test-py
 
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ else
 	BASE_COMMAND := ${DOCKER_COMPOSE} run --rm lando
 endif
 
+# This target needs to be the first in the file, so it's called by default.
 .PHONY: help
 help:
 	@$(DOCKER) --version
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -13,69 +13,65 @@ else
...
 .PHONY: test-js
-test-js:
+test-js:              ## run the JavaScript test suite (Vitest)"
 	$(BASE_COMMAND) npm test
 
-.PHONY: test-use-suite
-test-use-suite:
-	echo 1 > ${SUITE_STAMP}
+.PHONY: test-py
+test-py:              ##run the Python test suite"
```
cgsheeh: Using `sed` to self-document the commands like this is quite brittle. You can see this by the fact that `test-py` is currently missing from help output, due to the missing space after `##` here. :)

Also, there is a `"` suffixed to the end of the help, probably from the mechanical move over from proper strings.

Could we instead have a commented line that explicitly defines the help, something akin to this:

```diff
+# test-py-help: run the Python test suite
 .PHONY: test-py
-test-py:              ##run the Python test suite"
+test-py:
        $(BASE_COMMAND) lando tests $(ARGS_TESTS)
```

cgsheeh (dismissed):

Makefile
shtrom: Ah, I can update the RE (the whitespace matcher was too strict) and fix the ". That poor `test-py` got all the flaws!

I'm not sure using the `# <target-help>:` format makes much of a difference in brittleness. REs are gonna RE. I think the difference is that the current state breaks on whitespaces, while the separate comment has more visible characters to hang on to.

One advantage, maybe, of in-line comments here is that they also allow to keep the formatting relatively straightforwards, otherwise the help would be wonky. It could be a case of too much magic vs. just changing the format so we don't have to worry about alignment.

What about the new minor fixes? I don't mind going forwards with the multiline otherwise, but my preference remains with inline comments. WDYT?

cgsheeh: Yes, I think my main concern is the alignment requirement. Seems very easy to mess that up while making a change and end up with bad formatting in the help output. Also, if you do remember to update the alignment, you make the diff against this file very noisy. Can we update the formatting so the alignment doesn't matter?

Out-of-scope proposal -- I've been using `just` for a personal project and it supports help via comments on recipes out of the box. Maybe something we could consider adopting. :)

The other fixes to this PR look good to me. :+1:

.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build
-        run: docker build -t lando .
+        run: make build
```
zzzeid: This will now do:
```
docker compose build

```
Instead of
```
docker build -t lando .
```

In its current state, I think it would be better to keep the CI/CD command separate from the rest of the Make commands for a couple of reasons. With this change, changes to `compose.yaml` or changes to the Makefile can have an impact on the production build, possibly unintentionally (vs. current behaviour, where none of that matters).

I think we could make some modifications to avoid this scenario (e.g., decouple the `make build` command from the compose file).

Thoughts?

zzzeid (dismissed):

.github/workflows/build.yml
shtrom: Yeah, that's fair, the `compose build` may build more than what we need here.

Makefile
shtrom: TIL: `column`(1) https://www.man7.org/linux/man-pages/man1/column.1.html that takes care of the alignments need automatically.

I kept the comment inline, because multiline `sed` is not for the faint-hearted (;

cgsheeh (approved):
LGTM.

PR #1200, opened 2026-05-27
models.repo: add encrypted_hmac_secret field (bug 2042221)

zzzeid:
- add encrypted_hmac_secret field
- add mechanism to add/modify secret via admin
- update tests


src/lando/main/models/base.py
shtrom, line 1: nit: those get/set changes should be in #1195

shtrom (approved):
lgtm w/ nit

src/lando/main/models/base.py
zzzeid: Will move them over.

src/lando/main/admin.py
```diff
@@ -327,6 +328,21 @@ def desc(self, instance: Revision) -> str:
...
+            model = Repo
+            fields = "__all__"
+            exclude = ("encrypted_hmac_secret",)
+
+        hmac_secret = forms.CharField(
+            required=False, help_text="Enter a new value that will be encrypted."
+        )
+
+        def save(self, *args, **kwargs) -> Repo:
+            hmac_secret = self.cleaned_data.get("hmac_secret", None)
```
cgsheeh: Seems like this will always set the hmac secret, even if it wasn't submitted with the form. Should we guard against an empty value here, in case we end up overwriting the field?

src/lando/main/admin.py
```diff
@@ -327,6 +328,21 @@ def desc(self, instance: Revision) -> str:
 
 
 class RepoAdmin(admin.ModelAdmin):
+    class form(forms.ModelForm):
```
cgsheeh, line 331: Is there a reason we can't put this at the top-level as a `RepoAdminHmacForm` or similar, and then do `form = RepoAdminHmacForm`? That would improve readability and explain the purpose of the nested form.

cgsheeh (changes requested):
A few questions/nits, but blocking on moving the tests out of `__init__.py` from the parent PR. 👍

src/lando/main/admin.py
zzzeid: I'll double check + mitigate.

src/lando/main/admin.py
zzzeid: I think it's pretty clean + self-contained here, especially since it's not intended as a reusable form. It also would be `RepoAdminForm` if it were outside of this, since it overrides the whole repo form not just for the new field. Either way, I can change it if you think it would make things easier to read.

src/lando/main/models/repo.py
```diff
@@ -79,7 +79,7 @@ def get_default_hooks() -> list[str]:
     ]
 
 
-class Repo(BaseModel):
+class Repo(BaseModel, CryptographyMixin):
```
shtrom:
```suggestion
class Repo(CryptographyMixin, BaseModel):
```

src/lando/main/tests/__init__.py
```diff
@@ -133,6 +133,14 @@ def test_cryptography_mixin__set_field(mock_crypto_model):
     assert mock_crypto_model.save.call_count == 1
 
 
+def test_cryptography_mixin__set_field_do_not_save(mock_crypto_model):
```
shtrom: This should go in #1195.

zzzeid:
> A few questions/nits, but blocking on moving the tests out of `__init__.py` from the parent PR. 👍

Moved out of `__init__.py` and into `test_models.py` in #1195.

src/lando/main/models/repo.py
```diff
@@ -269,6 +269,7 @@ def path(self) -> str:
     )
 
     pr_enabled = models.BooleanField(default=False)
+    encrypted_hmac_secret = models.BinaryField(default=b"", blank=True)
```
cgsheeh: I'm about to add an HMAC secret for a Phab webhook. Could we rename this to something Github-specific to avoid a name collision/confusion?

src/lando/main/admin.py
```diff
@@ -374,6 +395,10 @@ def worker_count(self, instance: Repo) -> int:
         """Return the count of repositories associated to the Worker."""
         return instance.worker_set.count()
 
+    def hmac_secret_is_set(self, instance: Repo) -> bool:
+        "Return True if hmac_secret is set."
```
cgsheeh: This should be a docstring.
```suggestion
        """Return True if hmac_secret is set."""
```

src/lando/main/admin.py
cgsheeh: I think it's slightly confusing that `RepoAdmin.form` is being overridden implicitly here, I didn't even clock that on first read. It's not a huge deal though, we can leave it as-is. :)

src/lando/main/models/repo.py
cgsheeh: Nit: I think we said we were going to put mixins first for MRO in a previous PR.

cgsheeh (changes requested):

cgsheeh:
Also: can we add tests for the HMAC form clearing?

zzzeid:
> Also: can we add tests for the HMAC form clearing?

Adding this test is fairly easy, but it's going to fail currently because our repo fixtures don't actually validate in the form. I'll file a separate bug to fix that as it is out of scope.

src/lando/main/admin.py
shtrom: We could use the `@override` decorator to make it obvious.

zzzeid: Added `@override`.

cgsheeh (approved):

PR #1203, opened 2026-05-28
formatTime: handle invalid timestamps gracefully (Bug 2043328)

cgsheeh:


src/lando/static_tests/legacy/js/utils/formatTime.test.js
```diff
@@ -62,22 +62,45 @@ describe("$.fn.formatTime", () => {
     expect($("div").text()).toBe("untouched");
   });
 
-  test("throws when `data-timestamp` cannot be parsed as a date", () => {
+  test("leaves the element alone when `data-timestamp` cannot be parsed", () => {
     document.body.innerHTML =
-      '<time data-timestamp="not-a-date"></time>';
+      '<time data-timestamp="not-a-date">fallback</time>';
```
shtrom, line 67: We should add the fallbacks in the templates, e.g., https://github.com/mozilla-conduit/lando/blob/c111bf466be5bc8d2e61c766c3ba45d3cd5b0765/src/lando/ui/jinja2/partials/job.html#L9

shtrom (commented):
LGTM, and yay for tests!

One thing that remains to be done we be to make the template output the raw timestamp in the `<time>` elements, so the fallback mechanism is useful.

shtrom (approved):
lgtm!

PR #1205, opened 2026-06-02
repo: add configurable autoformat setup command sequence (Bug 2029056)

cgsheeh:
This PR adds a per-repo `autoformat_setup_commands` field that lets each repo
define the exact `./mach` commands run at worker startup when autoformatting is
enabled, with failures isolated per command. It also switches the default away
from `mach bootstrap` to a specific set of `mach artifact toolchain` fetches for
the toolchains we actually need. This is the way to set up clang-format,
rustfmt, and eslint as CI artifacts. Note that eslint additionally requires a
`lint --setup -l eslint` call to work, so that's included as its own step.


src/lando/main/models/repo.py
```diff
@@ -162,6 +178,15 @@ def path(self) -> str:
 
     approval_required = models.BooleanField(default=False)
     autoformat_enabled = models.BooleanField(default=False)
+    autoformat_setup_commands = models.JSONField(
```
shtrom, line 181: could we add an `autoformat_command[s]` too?

shtrom (approved):
lgtm as is; though I think we should take this opportunity to also add an `autoformat_commands` to take care of everything. This can be done is a subsequent PR.

I also wonder whether we could move away from hardcoding `./mach`. This would allow us to easily extend to other systems (e.g., `make format` for lando). Clearly not a high importance item, but it could come in handy.

src/lando/main/models/repo.py
zzzeid, line 181: nit: technically we could use ArrayField here (2d, str).

src/lando/main/models/repo.py
```diff
@@ -162,6 +178,15 @@ def path(self) -> str:
...
+        default=get_default_autoformat_setup_commands,
```
zzzeid: I think the default should be set only if `autoformat_enabled` is `True`. Thus, this should default to empty and set on save.

zzzeid (changes requested):

src/lando/main/models/repo.py
cgsheeh: I considered this, however we do this for `commit_flags` and we had to make that field readonly in the admin panel due to bug 1963919. Using a `JSONField` also gives us more flexibility in case we want to change the schema in the future.

src/lando/api/tests/test_landings.py
```diff
@@ -1206,6 +1207,60 @@ def test_run_mach_command_sets_mozbuild_state_path(tmp_path, git_landing_worker)
     )
 
 
+@pytest.mark.django_db
+def test_bootstrap_repos_runs_configured_command_sequence(repo_mc, git_landing_worker):
+    """`bootstrap_repos` runs each configured command in order, with the repo's state
+    path, and keeps going even when an earlier command fails."""
```
zzzeid: Multiline docstring should be a comment instead, or modified to have the summary in one line.

src/lando/api/tests/test_landings.py
```diff
@@ -1206,6 +1207,60 @@ def test_run_mach_command_sets_mozbuild_state_path(tmp_path, git_landing_worker)
...
+        name="test-bootstrap-git",
+        autoformat_enabled=True,
+        autoformat_setup_commands=commands,
+    )
+    git_landing_worker.worker_instance.applicable_repos.set([repo])
+
+    with mock.patch.object(
+        git_landing_worker,
+        "run_mach_command",
+        side_effect=[subprocess.CalledProcessError(1, "mach"), "ok"],
```
zzzeid, line 1227: What does the `ok` here mean?

zzzeid (approved):
lgtm, minus the small comment re: docstring, and one question.

src/lando/api/tests/test_landings.py
cgsheeh: We don't check that value, we just need to return something to indicate the command ran successfully, so we just return `"ok"` as a placeholder.

PR #1206, opened 2026-06-02
navbar: render admin button when user is a Conduit admin (Bug 2042232)

cgsheeh:
Extract a `user_is_conduit_admin` helper from the
`MaintenanceModeMiddleware` and switch the navbar
template to use it as a conditional check when
rendering the Admin button. Add a test for the extracted
helper while we're here.

### DIFF
--- a/src/lando/middleware.py
+++ b/src/lando/middleware.py
@@ -7,13 +7,13 @@ from typing import Optional
 
 from django.conf import settings
 from django.contrib.auth import authenticate
-from django.contrib.auth.models import User
 from django.core.handlers.wsgi import WSGIRequest
 from django.http import HttpRequest, HttpResponse
 from django.template.loader import render_to_string
 from django.template.response import TemplateResponse
 from django.urls import resolve
 
+from lando.main.auth import user_is_conduit_admin
 from lando.main.models import ConfigurationKey, ConfigurationVariable
 from lando.utils.phabricator import (
     PhabricatorAPIException,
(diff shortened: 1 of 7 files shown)


cgsheeh:
Side note: I noticed the Admin button is still displayed on dev, we need to update privileges there to match production.

src/lando/middleware.py
```diff
@@ -24,6 +25,18 @@
...
+def user_is_conduit_admin(user: User) -> bool:
+    """Return whether `user` is a Conduit administrator.
+
+    Administrators are staff users who belong to the `conduit-admin` group.
+    """
+    return (
+        user.is_authenticated
+        and user.is_staff
+        and user.groups.filter(name=CONDUIT_ADMIN_GROUP_NAME).exists()
+    )
```
zzzeid: I think this should go in `lando.main.auth`.

zzzeid (approved):

src/lando/ui/jinja2/partials/navbar.html
```diff
@@ -44,7 +44,7 @@
                                 </a>
                             </p>
                         {% endif %}
-                        {% if user_is_authenticated and request.user.is_superuser %}
+                        {% if user_is_authenticated and user_is_conduit_admin(request.user) %}
```
zzzeid: I also realized that `user_is_authenticated` here is now redundant, since that's the first check in `user_is_conduit_admin`.

src/lando/middleware.py
cgsheeh: Agreed, moved to `lando.main.auth`.

src/lando/ui/jinja2/partials/navbar.html
cgsheeh: Removed redundant check.

shtrom (approved):

PR #1219, opened 2026-06-03
main, utils, api: implement require_github_signature decorator (bug 2042208)

zzzeid:
- implement decorator to require github signature in webhook request
- add tests

### DIFF
--- a/src/lando/api/tests/test_views.py
+++ b/src/lando/api/tests/test_views.py
@@ -1,3 +1,5 @@
+import hashlib
+import hmac
 from unittest import mock
 
 import pytest
@@ -299,6 +301,23 @@ def test__views__pull_request_content_api_view__missing_csrf_token(
     ),
 )
 class TestViewsPullRequestUpdateWebHook:
+    hmac_secret = "test secret"
+
+    @pytest.fixture
+    def hmac_headers(self):
+        def calculate_signature():
+            _hmac = hmac.new(
+                self.hmac_secret.encode("utf-8"),
+                msg=b"--BoUnDaRyStRiNg--\r\n",
+                digestmod=hashlib.sha256,
+            )
+            return f"sha256={_hmac.hexdigest()}"
+
+        def _headers(signature=""):
+            return {"X-Hub-Signature-256": signature or calculate_signature()}
+
+        return _headers
+
     @pytest.fixture
     def webhook_gh_client(self):
         def wrapper(github_api_client, body):
@@ -319,14 +338,41 @@ class TestViewsPullRequestUpdateWebHook:
             mock_github_api_client.build_pull_request.return_value = mock_pr
             github_api_client.return_value = mock_github_api_client
 
-            Repo.objects.create(
+            repo = Repo.objects.create(
                 name="git-repo",
                 scm_type=SCMType.GIT,
                 pr_enabled=True,
             )
+            repo.set_gh_hmac_secret(self.hmac_secret)
             return mock_github_api_client
 
         return wrapper
+
+    @mock.patch("lando.api.views.generate_warnings_and_blockers")
+    @mock.patch("lando.api.views.GitHubAPIClient")
+    @pytest.mark.django_db(transaction=True)
+    def test__views__pull_request_update_webhook_no_hmac_header(
+        self,
+        github_api_client,
+        generate_warnings_and_blockers,
+        body,
+        expected_body,
+        client,
+        webhook_gh_client,
+        hmac_headers,
+    ):
+        """Test that the webhook fails when called without correct headers."""
+        mock_github_api_client = webhook_gh_client(github_api_client, body)
+
+        generate_warnings_and_blockers.return_value = {
+            "warnings": ["a warning"],
+            "blockers": ["a blocker"],
+        }
+
+        test = client.post("/api/pulls/git-repo/1/webhook")
+
+        assert mock_github_api_client.update_pull_request_body.call_count == 0
+        assert test.status_code == 403
 
     @mock.patch("lando.api.views.generate_warnings_and_blockers")
     @mock.patch("lando.api.views.GitHubAPIClient")
@@ -339,6 +385,7 @@ class TestViewsPullRequestUpdateWebHook:
         expected_body,
         client,
         webhook_gh_client,
+        hmac_headers,
     ):
         """Test that the webhook is calling the GitHub API with the correct parameters."""
         mock_github_api_client = webhook_gh_client(github_api_client, body)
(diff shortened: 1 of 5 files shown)


src/lando/main/support.py
```diff
@@ -31,3 +34,14 @@ def __init__(self, status, detail, extra=None):  # noqa: ANN001
         }
         if self.extra:
             self.json_detail.update(self.extra)
+
+
+def verify_github_signature(hmac_secret: str, payload: str, signature: str) -> bool:
```
shtrom: nit: This should go in the utils.github module.

shtrom (approved):
LGTM w/ nit: `verify_github_signature` could be in `utils.github`

src/lando/main/support.py
zzzeid: Done.

src/lando/main/auth.py
```diff
@@ -245,6 +247,24 @@ def wrapper(*args, **kwargs):
...
 
 
+def require_github_signature(post: Callable) -> Callable:
+    """Decorator to block requests requiring a valid signature header."""
+
+    def _post(view: View, request: WSGIRequest, repo_name: str, pull_number: int):
+        signature_header = "X-Hub-Signature-256"
+        if signature_header not in request.headers:
+            raise PermissionDenied(f"{signature_header} header is required")
+        repo = Repo.objects.get(name=repo_name)
```
cgsheeh: Seems like we're missing a check that `repo.hmac_secret` is actually set here. Since the default value is an empty byte string, a repo without this value set will always validate, bypassing the auth. We should add a test case for this as well.
```suggestion
        repo = Repo.objects.get(name=repo_name)
        if not repo.hmac_secret:
            raise PermissionDenied("No webhook secret set for this repo")
```

src/lando/utils/github.py
```diff
@@ -763,3 +765,14 @@ def parse_author_information(self) -> tuple[str, str]:
     def get_timestamp(self) -> str:
         """Return an `hg export` formatted timestamp."""
         return self.get_header("date")
+
+
+def verify_github_signature(hmac_secret: str, payload: str, signature: str) -> bool:
```
cgsheeh: This method is a pure function, we should add some unit test cases for it.

src/lando/api/views.py
```diff
@@ -374,6 +378,7 @@ def generate_context(self, request: HttpRequest) -> dict[str, str | list[str]]:
         context["commit_body"] = self.pull_request.commit_body
         return context
 
+    @require_github_signature
```
cgsheeh, line 414: This prevents the code in `post` from running without authentication. However the `dispatch` code still runs and makes API calls to Github. Worth calling out here as that may impact API usage limits if a bot started spamming the endpoint.

src/lando/main/auth.py
cgsheeh: I think we can use `view.target_repo` here instead.

Also, should `view` be `self`?

src/lando/api/tests/test_views.py
```diff
@@ -167,10 +169,23 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
...
 @mock.patch("lando.api.views.generate_warnings_and_blockers")
 @pytest.mark.django_db(transaction=True)
-def test__views__pull_request_update_webhook(
+def test__views__pull_request_update_webhook__integration(
     generate_warnings_and_blockers, github_api_client, client
 ):
     """Test that the webhook is calling the GitHub API with the correct parameters."""
+    secret = "test secret"
+
+    def calculate_signature():
```
cgsheeh: We should add a test for an incorrect secret here.

cgsheeh (changes requested):

src/lando/api/views.py
zzzeid: True, this is the same as with any other endpoint that calls GitHub (e.g., visiting a PR page). We should find a fix for this though. If this becomes an issue we should probably limit all PR endpoints to authorized users.

src/lando/main/auth.py
zzzeid: I don't think it makes sense to use `self` outside the instance method context.

zzzeid: >I think we can use view.target_repo here instead.

Possibly. As it's currently written, this decorator makes no assumption about which view it is being used on.

zzzeid: >Seems like we're missing a check that repo.hmac_secret is actually set here

Yep, this should be part of this check, will add.

src/lando/main/auth.py
```diff
@@ -245,6 +246,32 @@ def wrapper(*args, **kwargs):
...
+        if signature_header not in request.headers:
+            raise PermissionDenied(f"{signature_header} header is required")
+
+        if not view.target_repo.hmac_secret:
+            # This is technically a configuration error, should surface as a 500.
+            raise ValueError("HMAC secret is not set on repo")
+
+        is_valid = verify_github_signature(
+            view.target_repo.hmac_secret,
+            request.body.decode("utf-8"),
```
cgsheeh: We pass `request.body.decode("utf-8")` here, but then we do `payload.encode("utf-8")` in `verify_github_signature`. We should pass `request.body` directly instead to avoid potential failures if the body isn't valid utf-8.

src/lando/main/auth.py
```diff
@@ -245,6 +246,32 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def require_github_signature(post: Callable) -> Callable:
+    """Decorator to block requests requiring a valid signature header."""
+
+    def _post(view: View, request: WSGIRequest, *args, **kwargs):
```
cgsheeh, line 253: Missing `@functools.wraps()` decorator here.

src/lando/main/auth.py
```diff
@@ -245,6 +246,32 @@ def wrapper(*args, **kwargs):
...
+            raise PermissionDenied(f"{signature_header} header is required")
+
+        if not view.target_repo.hmac_secret:
+            # This is technically a configuration error, should surface as a 500.
+            raise ValueError("HMAC secret is not set on repo")
+
+        is_valid = verify_github_signature(
+            view.target_repo.hmac_secret,
+            request.body.decode("utf-8"),
+            request.headers.get(signature_header),
```
cgsheeh: Nit: `request.headers[signature_header]` should be okay since we assert it is in the headers above. This fixes a type-checker warning, since the argument type is `str` but `.get()` makes it `str | None`.

cgsheeh (approved):
r+ w/ nits. I think we should avoid the unnecessary pass through utf-8 before merging.

PR #1233, opened 2026-06-09
dependabot: update config (bug 2045933)

shtrom:
* dependabot: update pip config
 * dependabot: add npm config
 * deps: implement 7days cooldown
 * generate_requirements: ignore dubious git ownership in setuptools_scm

### DIFF
--- a/src/lando/utils/management/commands/generate_requirements.py
+++ b/src/lando/utils/management/commands/generate_requirements.py
@@ -36,7 +36,7 @@ class Command(BaseCommand):
 
         command.append("pyproject.toml")
 
-        env = os.environ
+        env = os.environ.copy()
         # Tell setuptools_scm to ignore git complaints about ownership of /code in
         # container [0].
         # [0] https://github.com/pypa/setuptools-scm/pull/1235
--- a/src/lando/utils/management/commands/generate_requirements.py
+++ b/src/lando/utils/management/commands/generate_requirements.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 import subprocess
 
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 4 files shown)


cgsheeh (approved):

zzzeid (approved):
Would be nice if it had a `suffix` option.

Edit: rewording!

shtrom:
Ok, I added cooldown support in our scripts. We only have npm 10.x from bookworm, which doesn't support cooldown, so the Makefile determines the cut-off date dynamically, but there is an `.npmrc` in place, ready for the future.

(There was also a bug with setuptools_scm preventing us from running upgrade-requirements, so that's fixed, too).

Given how we update the deps, the `prefix` is not terribly useful, save for the fact that it points to the relevant bug, which now contains instructions about how to update!

src/lando/utils/management/commands/generate_requirements.py
```diff
@@ -34,6 +36,11 @@ def handle(self, *args, **options):
 
         command.append("pyproject.toml")
 
+        env = os.environ
```
cgsheeh: Should this be `os.environ.copy()`? Might come back to bite us if we ever call this from somewhere else, and it feels like the intention was to create a copy since we could just set the value without the `env` variable otherwise.

cgsheeh (approved):

src/lando/utils/management/commands/generate_requirements.py
shtrom: I swear, I had the `.copy()` in my head. I'll file a bug against my... keyboard, yeah, keyboard.

```suggestion
        env = os.environ.copy()
```

PR #1234, opened 2026-06-09
autoformat: add `AutoformatChange` model to track formatting changes (Bug 2046117)

cgsheeh:
Add a new `AutoformatChange` model to track changes made by
automated code formatting. The model references the landing
job where changes were applied, the changed files and the
diff created by the formatters. A follow-up commit will add
this new model to the ETL and display autoformatting results
in the landing job UI.


src/lando/main/models/autoformat.py
```diff
@@ -0,0 +1,22 @@
...
+
+from lando.main.models.base import BaseModel
+
+
+class AutoformatChange(BaseModel):
+    """Record of autoformatting changes applied to a commit during landing."""
+
+    landing_job = models.ForeignKey(
+        "LandingJob",
+        on_delete=models.CASCADE,
```
zzzeid: Why CASCADE here? In general I think SET_NULL is safer. We don't delete landing jobs usually anyway, so this value in theory should not have a real-world impact but I think better keep records if we can.

This may also have an impact on audit logs.

src/lando/main/models/autoformat.py
```diff
@@ -0,0 +1,22 @@
+from django.db import models
+
+from lando.main.models.base import BaseModel
+
+
+class AutoformatChange(BaseModel):
```
zzzeid: nit: may be OK to put this in `models.landing_job`. By itself here feels like it is orphaned, and I don't think it's enough of a top-level model to warrant its own module.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -50,6 +52,30 @@
 """.strip()
 
 
+@dataclass
+class AutoformattingResult:
```
zzzeid: nit: since you already have this dataclass, might as well add a descriptive property that returns `replacements[0]` since it's repeated a couple of times in this module.

zzzeid (changes requested):

src/lando/main/models/autoformat.py
cgsheeh: Makes sense to me.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -384,15 +400,25 @@ def autoformat(
 
             return message
 
-        # If autoformatting added any changesets, note those in the job.
-        if replacements:
-            job.formatted_replacements = replacements
+        # If autoformatting added any changesets, note those in the job and record
+        # the changed files and diff for later analysis.
+        if result is not None:
+            job.formatted_replacements = [result.commit_sha]
```
zzzeid, line 406: This is a change from current behaviour. Is this intentional? Did `replacements` previously only contain a single SHA? If so, we should change that in `commit_autoformatting_changes` to return only a single SHA (instead of a list of a single element).

zzzeid (commented):
Just a clarification comment, not sure if this is a legacy thing or intentional for a different reason.

src/lando/api/legacy/workers/landing_worker.py
zzzeid: In other words, seems like we are transforming a list of a single element to a single element then back to a list with a single element, and it's unclear why that is.

cgsheeh: This is a legacy implementation detail. Previously when autoformatting used `hg fix` there may have been multiple changed commits, so `formatted_replacements` was a list of all changed commit SHAs. Nowadays there is always a single changed commit SHA, so we can change `format_stack_*` methods to return `str` instead of `list[str]` as a nice cleanup. :)

zzzeid (approved):
lgtm

PR #1239, opened 2026-06-11
repo: add a configurable autoformatting command (Bug 1932005)

cgsheeh:
Add a field on the `Repo` model to manage the autoformatting command.


src/lando/main/tests/test_models.py
```diff
@@ -403,6 +406,40 @@ def test__models__Repo__autoformat_setup_commands_not_overwritten():
...
+        url="http://example.com",
+        scm_type=SCMType.GIT,
+        autoformat_enabled=True,
+        autoformat_run_command=custom,
+    )
+    repo.save()
+
+    assert repo.autoformat_run_command == custom, (
+        "An explicitly-set run command should not be overwritten on save."
+    )
```
zzzeid: We have many fields that follow this pattern (basically everything in `Repo.save`). I think we can have one test parametrized with all of the different properties instead of testing each individual one. 

Within the scope of this PR I think at least factoring these tests plus the two above would be acceptable.

src/lando/main/models/repo.py
```diff
@@ -67,6 +67,16 @@ def get_default_autoformat_setup_commands() -> list[list[str]]:
     ]
 
 
+def get_default_autoformat_run_command() -> list[str]:
+    """Return the default autoformat run command for a repo.
+
+    The returned list is a single `./mach` invocation (without the `./mach` prefix),
+    run against the patch stack to apply formatting. Repos that don't support the
+    `--skip-android` flag can override this with their own command.
```
zzzeid: nit: I think the `--skip-android` mention here is too specific, maybe it belong in a comment, and the docstring should be more generic.

zzzeid (changes requested):
Looks good, main comment is re: repeated test pattern.

shtrom (approved):

src/lando/main/tests/test_models.py
```diff
@@ -358,48 +361,58 @@ def test__models__Repo__mozbuild_state_path():
...
+        ["format", "--fix", "--outgoing"],
+    ),
+)
 
 
 @pytest.mark.django_db
-def test__models__Repo__autoformat_setup_commands_set_on_save():
-    """Enabling autoformatting populates the default `mach artifact` sequence on save."""
+@pytest.mark.parametrize(
+    "field, default_factory, custom_value", AUTOFORMAT_DEFAULT_FIELDS
```
zzzeid: nit: assuming linting/proper indentation, I think it is better to put `AUTOFORMAT_DEFAULT_FIELDS` directly here, it's part of the test and would make this easier to interpret. I.e., we don't need to store it in a variable.

src/lando/main/tests/test_models.py
```diff
@@ -358,48 +361,58 @@ def test__models__Repo__mozbuild_state_path():
...
+    (
+        pytest.param(False, False, id="disabled-stays-empty"),
+        pytest.param(True, False, id="enabled-gets-default"),
+        pytest.param(True, True, id="explicit-preserved"),
+    ),
+)
+def test__models__Repo__autoformat_field_default(
+    field, default_factory, custom_value, autoformat_enabled, explicit
+):
+    """`Repo.save` defaults each autoformat field only when enabled and unset."""
```
zzzeid:
```suggestion
    """Test that `Repo.save` defaults each autoformat field only when enabled and unset."""
```

zzzeid (approved):

PR #1251, opened 2026-06-12
prettier: update configs to more closely match Python formatting (Bug 2047147)

cgsheeh:
We're currently using the default tab width of 2. As a Python team, this
seems like the wrong choice. Switch to a tab width of 4 spaces to align
with the black/`ruff format` convention of 4-space indentation.

We also use the Prettier default line length of 80, when `ruff format`
uses `88`. Add `printWidth: 88` to align with Python as well.


cgsheeh:
One thing worth discussing: CSS uses 2-space tab width as the convention. I could update this to use 2 spaces for just CSS files and 4 spaces for JS, if we wanted to keep that convention.

zzzeid (approved):
r+ with caveat that this is going to cause major conflicts for anyone working on JS files, we should probably not merge it until after those changes are in place (or for whoever has to deal with those conflicts, can do a checkout + reformat.)

cgsheeh:
> r+ with caveat that this is going to cause major conflicts for anyone working on JS files, we should probably not merge it until after those changes are in place (or for whoever has to deal with those conflicts, can do a checkout + reformat.)

Yes, I was thinking the same thing, we should allow https://github.com/mozilla-conduit/lando/pull/1161 to land before this. :)

I'm about to post a PR which adds new JS code, which was the motivation for this change. I'll format that with the new 4-space indentation for the review, with a note that the test failure is expected.

PR #1252, opened 2026-06-15
uplift: add Vue train-selector widget for uplift target selection (bug 2044143, bug 2047613)

cgsheeh:
Bootstrap Vue 3 + TypeScript into the project via a Vite build (vitest and
@vue/test-utils for tests, vue-tsc as a type-check gate). The bundle is built
into static_src/dist during the Docker image build and served by Django as a
static asset; the frontend sources live in frontend/.

Add a train-selector widget that progressively enhances the Django-rendered
"Uplift repositories" checkboxes in the uplift request modal. The native
checkboxes remain the source of truth for the form submission, so the backend
is unchanged -- the widget only reads and toggles them, and falls back to the
plain checkboxes if it cannot load.

When the "Request Uplift" button is first clicked, the widget fetches
release-train guidance from the whattrainisitnow.com API (see bug 2045812) and
validates the response with a Zod schema. That response describes the current
state of the release cycle (which version is on each train, whether betas
remain, whether the release candidate has shipped); the held response drives
every recommendation, landing description, and warning the widget produces.

The widget presents two tabs:

- "Select Firefox Version": choosing a target version resolves -- based on the
  current cycle state -- to the correct uplift train(s) (beta, release, or
  both), ticks the matching repositories, and shows a single line describing
  where the patch will land.
- "Select Uplift Train": manual repository selection with per-train landing
  hints that reflect the cycle state, including a warning when beta is selected
  after betas have closed.

The tabs are keyboard- and screen-reader-accessible.


cgsheeh:
Screenshots:

Screenshot 2026-06-12 at 11-39-45 D3997- Lando - Mozilla
Screenshot 2026-06-12 at 11-41-23 D3997- Lando - Mozilla
Screenshot 2026-06-12 at 11-41-43 D3997- Lando - Mozilla
Screenshot 2026-06-12 at 11-42-43 D3997- Lando - Mozilla

API could not retrieve data from whattrainisitnow.com:

Screenshot 2026-06-12 at 11-46-41 D3997- Lando - Mozilla

zzzeid (changes requested):
This should be split into multiple PRs and bugs, specifically the bootstrapping portion (needs its own bug) and the uplift-specific portions.

frontend/src/components/TrainSelector.test.ts
```diff
@@ -0,0 +1,242 @@
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { mount, flushPromises } from "@vue/test-utils";
+import TrainSelector from "./TrainSelector.vue";
+import type { ReleaseSchedule } from "../trainGuidance";
```
cgsheeh, line 4: In other projects I have defined a `path` in the TypeScript config like this:

```json
        "paths": {
            "@/*": ["./src/*"]
        },
```

Which would then allow doing `@/components/TrainSelector.vue` and `@/trainGuidance` instead of using relative paths. Can add that now or as a follow-up if we think it's useful.

cgsheeh:
Closing in favour of two PRs with the Vue/TypeScript bootstrapping changes and uplift-guidance widget implementation separated out for easier review.

PR #1253, opened 2026-06-15
landing: apply patches at their base and rebase for a 3-way merge (Bug 2047605)

cgsheeh:
Lando applied patches with a 2-way `git apply`, which fails when the surrounding
context shifted even from an unrelated edit. Track the diff's base revision from
Phabricator's `differential.diff.search` and, when that commit exists in the
target repo, apply the stack onto it and rebase onto the target branch for a true
3-way merge. Fall back to applying at the tip when the base is unavailable, and
skip rebasing entirely when `target_commit_hash` pins the target.

Post-rebase commit hashes are recorded on each revision, and rebase conflicts are
captured and surfaced through the existing merge-conflict error breakdown.


src/lando/main/scm/abstract_scm.py
```diff
@@ -119,6 +119,15 @@ def apply_patch_git(self, patch_bytes: bytes):
     def cherry_pick_commit(self, commit_id: str):
         """Cherry-pick the specified commit onto the current branch."""
 
+    def supports_3way_apply(self) -> bool:
```
shtrom, line 123: nit: Could we make that a `@property`? Though I seem to recall some issue with overriding properties... 🤔

shtrom: less nit: we should make `rebase_onto` part of the AbstractSCM interface, raising a `NotImplementedError` in the base classe.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -245,25 +245,45 @@ def apply_patch(revision: Revision):
...
             self.handle_new_commit_failures(apply_patch, repo, job, scm, revision)
 
-            new_commit = scm.describe_commit()
-            logger.debug(f"Created new commit {new_commit}")
+        # If we reconstructed at the base, rebase the stack onto the landing base
+        # to merge it against the target branch.
+        if rebase_base:
+            self.handle_new_commit_failures(
+                rebase_stack, repo, job, scm, job.revisions.last()
+            )
```
shtrom, line 285: It may be useful to report whether the conflicts have occurred during the initial apply (unlikely), or subsequent rebase, and make that visible to the requester.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -348,6 +368,55 @@ def apply_patch(revision: Revision):
 
         return bug_ids, commit_id
 
+    def determine_rebase_base(self, job: LandingJob, scm: AbstractSCM) -> str | None:
+        """Return the base commit to reconstruct the stack on, or `None`.
+
+        Returns `None` — applying directly onto the landing base — when the SCM
```
shtrom: Hmmmm. emdash q:

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -348,6 +368,55 @@ def apply_patch(revision: Revision):
...
+        cannot rebase, when an exact target commit is already known
+        (`target_commit_hash`), or when the recorded base is missing from the
+        repo.
+        """
+        if job.target_commit_hash:
+            # The exact target commit is known, so apply onto it without rebasing.
+            return None
```
shtrom, line 390: I wonder if we wouldn't be better off handling the `rebase_base` in the same way as the `target_commit_hash`, when calling `self.update_repo(repo, job, scm, job.target_commit_hash)`. We should already have the logic there, which would save us a reset (though maybe we'd need to pull first, to make sure the commit exists; this would make this point a bit moot?).

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -348,6 +368,55 @@ def apply_patch(revision: Revision):
...
+        landed_commits = scm.describe_local_changes(landing_base)
+        revisions = list(job.revisions.all())
+
+        if len(landed_commits) != len(revisions):
+            logger.warning(
+                f"Expected {len(revisions)} commits after landing, found "
+                f"{len(landed_commits)}; commit IDs may be misaligned."
+            )
+
+        for revision, commit in zip(revisions, landed_commits, strict=False):
```
shtrom: nit: we could set `strict=True` and except the `ValueError` if the length don't match.

We'd have to save() separately at the end, though. Maybe not that practical...

src/lando/api/tests/test_landings.py
```diff
@@ -289,6 +290,42 @@
...
+    ],
+)
+@pytest.mark.django_db
+def test_determine_rebase_base(
+    git_landing_worker,
+    target_commit_hash,
+    supports_3way,
+    base_revision,
+    base_exists,
+    expected,
```
shtrom: nit: typehints

src/lando/api/tests/test_landings.py
```diff
@@ -1471,3 +1508,207 @@ def test_worker_active_repos_updated_when_tree_closed(
...
+    )
+
+    worker = get_landing_worker(SCMType.GIT)
+    assert worker.run_job(job), "`run_job` returns `True` in both permanent states."
+    assert job.status == expected_status, (
+        "Base availability should determine whether the context shift lands."
+    )
+
+    if expected_status != JobStatus.LANDED:
+        return
```
shtrom, line 1635: Could we check that the `job.error` contains something useful?

src/lando/main/models/revision.py
```diff
@@ -54,6 +54,12 @@ class Revision(BaseModel):
     pull_head_sha = models.CharField(max_length=40, blank=True, null=True)
     pull_base_sha = models.CharField(max_length=40, blank=True, null=True)
 
+    # The git commit hash the patch was authored against, as reported by
+    # Phabricator's `differential.diff.search` `base` ref. When this commit
+    # exists in the target repo, the landing worker can apply the patch onto it
+    # and rebase, yielding a true 3-way merge. Empty when unknown.
+    base_revision = models.CharField(max_length=40, blank=True, default="")
```
shtrom, line 61: Do we want to add this to the ETL?

src/lando/main/scm/git.py
```diff
@@ -191,6 +191,62 @@ def cherry_pick_commit(self, commit_id: str):
...
+
+    def rebase_onto(self, new_base: str, upstream: str):
+        """Rebase the commits in `upstream..HEAD` onto `new_base`.
+
+        Replays each commit as a 3-way merge against `new_base`, recovering the
+        context-shift failures that a 2-way apply would reject. On a genuine
+        same-line conflict, abort the rebase and raise `PatchConflict`.
+        """
+        try:
+            self._git_run("rebase", "--onto", new_base, upstream, cwd=self.path)
```
shtrom: nit: we should add `--no-autosquash` here to be sure we don't try to reorder `!fixups` and `!squash` (which should be the default, but let's be explicit).

src/lando/main/scm/git.py
```diff
@@ -381,6 +443,39 @@ def process_merge_conflict(
...
+        rather than as `.rej` files on disk.
+        """
+        breakdown: dict[str, Any] = {
+            "failed_paths": [],
+            "rejects_paths": {},
+            "revision_id": revision_id,
+        }
+
+        for path, content in conflicts.items():
+            commit = self.last_commit_for_path(path)
```
shtrom: I wonder if we should move this to the logic that extracts all the conflicts. This would make this method only work on its input, with no interaction with the underlying repo.

src/lando/main/tests/test_git.py
```diff
@@ -1381,6 +1381,207 @@ def test_GitSCM_commit_exists(
...
+ line 4
+-line 5
++line 5 - changed on work branch
+ line 6
+ line 7
+ line 8
+"""
+
+PATCH_AUTHOR = "Py Test <pytest@lando.example.net>"
+PATCH_DATE = "1970-01-01T00:00:00"
```
shtrom: Could we make those into fixtures, so we can reuse them in the landing tests, rather than defining similar constants there?

src/lando/main/tests/test_git.py
```diff
@@ -1381,6 +1381,207 @@ def test_GitSCM_commit_exists(
...
+
+
+def commit_diff(scm: GitSCM, diff: str, message: str):
```
shtrom: nit: could we name this `apply_patch`? 😅

src/lando/main/tests/test_git.py
```diff
@@ -1381,6 +1381,207 @@ def test_GitSCM_commit_exists(
...
+    git_repo: Path,
+    git_setup_user: Callable,
+    request: pytest.FixtureRequest,
+    tmp_path: Path,
+):
+    """`reset_to_commit` moves the work branch back to the given commit."""
+    scm = clone_git_repo(git_repo, tmp_path / request.node.name, git_setup_user)
+
+    base_commit = scm.head_ref()
+    commit_diff(scm, SHARED_FILE_DIFF, "Add shared file")
```
shtrom: Could we reuse the `create_git_commit` fixture here? https://github.com/mozilla-conduit/lando/blob/1edd8ea018169496832894aac33aed5c6b4b8674/src/lando/conftest.py#L1013

src/lando/main/tests/test_git.py
```diff
@@ -1381,6 +1381,207 @@ def test_GitSCM_commit_exists(
...
+    assert scm.head_ref() != base_commit, "A new commit should advance `HEAD`."
+
+    scm.reset_to_commit(base_commit)
+    assert scm.head_ref() == base_commit, (
```
shtrom, line 1419: Yes! head_ref my old friend!

src/lando/main/tests/test_git.py
shtrom: Or, alternatively, make `commit_diff`/`apply_patch` a fixture there, as I can see how it could be useful to reuse, but we'll lose track of it if it's in here.

shtrom (changes requested):
Looking good. Mainly some suggestions and discussion to clarify.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -348,6 +368,55 @@ def apply_patch(revision: Revision):
...
+        if job.target_commit_hash:
+            # The exact target commit is known, so apply onto it without rebasing.
+            return None
+
+        if not scm.supports_3way_apply():
+            return None
+
+        first_revision = job.revisions.first()
+        if not first_revision:
+            return None
```
Magnolia-Liu, line 398: Question: In what situation would a landing job reach determine_rebase_base with zero revisions / when would we need to land an empty stack? If that can't happen, would a comment or assertion documenting that case be clearer returning None?

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -348,6 +368,55 @@ def apply_patch(revision: Revision):
...
+        """Store each revision's final commit hash after applying and rebasing.
+
+        Commits are read against `landing_base` in ascending topological order,
+        which matches the index order of `job.revisions`.
+        """
+        landed_commits = scm.describe_local_changes(landing_base)
+        revisions = list(job.revisions.all())
+
+        if len(landed_commits) != len(revisions):
+            logger.warning(
```
Magnolia-Liu: Question: Why is a warning used here instead of failing the job if the commit IDs are misaligned? What issues could arise from the incorrect mapping? Or is this expected to happen in some cases and the issues are relatively benign?

Magnolia-Liu (commented):
Looks good to me, just some clarifying questions.

src/lando/api/tests/test_landings.py
cgsheeh: Love me some type hints, added. :)

At some point we should go back and add type hints to all our test code, and then remove the exception from `ruff`.

src/lando/main/tests/test_git.py
cgsheeh: I went with the alternative of `apply_patch`, since `create_git_commit` is making a random commit and depends on commit signing.

src/lando/api/tests/test_landings.py
cgsheeh: Good call, added.

src/lando/main/models/revision.py
cgsheeh: Yes, we should add it to the ETL. However I'll do that as a follow-up, since it requires a BigQuery schema update first.

src/lando/main/scm/abstract_scm.py
cgsheeh: Yes, I think this makes sense. Made `supports_3way_apply` a property, and added `rebase_onto` to the interface.

src/lando/main/scm/git.py
cgsheeh: Worth it as a safeguard, added.

src/lando/main/tests/test_git.py
cgsheeh: These are no longer needed, removed.

src/lando/api/legacy/workers/landing_worker.py
cgsheeh: Good question! This was added as a defensive check so we would fall back to 2-way apply if we didn't have the data required to complete the 3-way apply (a revision with a `base_revision` that exists in the repo).

However looking again, if we were to hit this case we would fall back to the 2-way merge, where we would simply apply nothing and the job would "land" despite making no changes in the push. So yes, I think we should handle this case, but we will do it outside of this specific method. :)

src/lando/api/legacy/workers/landing_worker.py
cgsheeh: My reasoning was that if we get to this point, we managed to apply the patches and rebase them without an error, so the stack should be able to land. However the misaligned commit ID has the potential to interfere with the cherry-picking workflow in uplift jobs, since we use the recorded revision in `git cherry-pick <SHA>`.

I'll change this to a hard failure instead. Nice catch. :)

src/lando/api/legacy/workers/landing_worker.py
cgsheeh: They will see this due to the `PatchConflict` exception message making it's way through to `LandingJob.error` in `handle_new_commit_failures`. As a follow-up I'd like to fall back to the 2-way apply when applying to the base fails, but I want to see if that is a common error state before I spend time on it.

src/lando/api/legacy/workers/landing_worker.py
cgsheeh: Good idea, but I think we can do that as a follow-up. The current logic for `GitSCM` assumes the `target_cset` is a branch, so it will require a minor refactor.

src/lando/api/legacy/workers/landing_worker.py
cgsheeh: Hardened the check for revision length/non-emptiness, which allowed me to use `strict=True`.

src/lando/main/scm/git.py
cgsheeh: Good call, done.

shtrom (approved):
:shipit:

PR #1254, opened 2026-06-15
frontend: bootstrap Vue 3 + TypeScript via a Vite build (bug 2047613)

cgsheeh:
Stand up a Vue 3 + TypeScript frontend built by Vite (vitest and
@vue/test-utils for tests, vue-tsc as a type-check gate). The bundle is built
into `static_src/dist` during the Docker image build and served by Django as a
static asset; frontend sources live in `frontend/`.

The base layout gains a `scripts` block so pages can load the bundle. The entry
point is a placeholder that mounts nothing until widgets are added in follow-up
changes.


tsconfig.node.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.node.tsbuildinfo",
+    "target": "es2023",
+    "lib": ["ES2023"],
+    "module": "esnext",
+    "types": ["node"],
+    "skipLibCheck": true,
+
```
cgsheeh, line 9: In other projects I have defined a `path` in the TypeScript config like this:

```json
        "paths": {
            "@/*": ["./src/*"]
        },
```

Which would then allow doing an import from `@` instead of using relative paths, like so:

```typescript
import {} from @/components/TrainSelector.vue
```

Worth doing now?

vitest.config.js
cgsheeh, line 1: Content moved to `vite.config.ts`.

cgsheeh:
Note: added the new code in a new `frontend/` directory, happy to discuss that and move to `src/frontend` or elsewhere.

src/lando/settings.py
```diff
@@ -260,8 +260,9 @@
 ]
 
 LINT_PATHS = (BASE_DIR,)
-# Currently used by Prettier for linting CSS/JS.
-STATIC_LINT_PATHS = (BASE_DIR / "static_src",)
+# Used by Prettier for linting CSS/JS and the Vue/TypeScript frontend. The
+# `frontend` source lives at the repository root (`BASE_DIR` is `src/lando`).
+STATIC_LINT_PATHS = (BASE_DIR / "static_src", BASE_DIR.parent.parent / "frontend")
```
cgsheeh, line 268: While writing this I considered that it would make more sense to run the front-end formatters in the Vitest test suite instead of in the Python test suite. That way `make test-js` actually asserts the JS code is properly formatted.

shtrom: The `parent.parent` seems a bit smelly, though I suspect it may not be worth trying to fix: I can only think about moving `fronted` to `static_src`, but it would hide the fontend code within the Python code, which I think would be worse.

Unless we have some way to reach the root of the tree (probably `BASE_DIR.parent.parent` 😅)

tsconfig.node.json
shtrom: Can't see why not.

shtrom (approved):
lgtm.

src/lando/settings.py
zzzeid: > While writing this I considered that it would make more sense to run the front-end formatters in the Vitest test suite instead of in the Python test suite. That way `make test-js` actually asserts the JS code is properly formatted.

I think it may be better to go with this approach vs. what is here now.

Agree that `BASE_DIR.parent.parent` is not great. In fact, we shouldn't be referencing directories outside of `BASE_DIR` in the django app (though we do that already in the `tests` command). I would suggest moving this specific addition out of `settings` and into the command module, to keep it separated from app-level stuff.

Since the frontend implementation is in flux and we will likely have better separation in the near future, I don't think this is a blocker though. But to keep things cleaner maybe move this out of settings.

vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+import vue from "@vitejs/plugin-vue";
+
+// The frontend is built into `static_src/dist` and served by Django as a plain
+// static asset. There is no dev server; rebuild with `make build-js` after
+// changing the frontend sources.
+export default defineConfig({
+  plugins: [vue()],
+  build: {
+    outDir: fileURLToPath(
+      new URL("./src/lando/static_src/dist", import.meta.url),
```
zzzeid: We mainly use `static_src` for manually-created (or copied), legacy static files. I think this should go somewhere else, particularly because it will be the odd one out. Adding things in static_src to `.gitignore` also breaks the convention.

Could go with `static_dist`, or `frontend_dist`. This will be cleaned up in the future (likely as part of [bug 2019284](https://bugzilla.mozilla.org/show_bug.cgi?id=2019284) et al).

```suggestion
      new URL("./src/lando/static_dist/", import.meta.url),
```

vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+    rollupOptions: {
+      input: fileURLToPath(new URL("./frontend/src/main.ts", import.meta.url)),
+    },
+  },
+  resolve: {
+    alias: {
+      "@static_src": fileURLToPath(
+        new URL("./src/lando/static_src", import.meta.url),
+      ),
+    },
```
zzzeid, line 30: Maybe add a comment here as to what this does now? Since it is moved from the old config I assume it's unrelated to the new bootstrapping, however it still references the parent directory of all the built assets.

Dockerfile
```diff
@@ -67,6 +67,11 @@ RUN mkdir -p /static_vendor \
 # Copy code into the container.
 COPY ./ /code
 
+# Build the Vue frontend bundle into `static_src/dist`, served by Django as a
```
zzzeid: There's references here and elsewhere to the destination directory. Instead they should be referencing the config file setting (e.g., `build.outDir`).

frontend/src/main.ts
zzzeid, line 1: Could we put this in `frontend/main.ts` instead? Anything else going in `frontend/`? Not a blocker since I think this will be modified in the near future to be compatible with the more long term approach for vue.js.

src/lando/ui/jinja2/partials/layout.html
```diff
@@ -112,5 +112,7 @@
         {% include "partials/notifications.html" %}
         {% block main %}{% endblock %}
         {% include "partials/footer.html" %}
+        {# Page-specific scripts, loaded after the global bundle in `<head>`. #}
```
zzzeid: nit: A bit redundant, since it's pretty clear that this block is in `<body>` (this is a small file).

```suggestion
        {# Page-specific scripts #}
```

zzzeid (changes requested):

src/lando/settings.py
cgsheeh: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2048013 for putting the JS formatting tests in the JS test suite. Will handle fixing that as a follow-up and leave this as-is for now. :+1:

frontend/src/main.ts
cgsheeh: Hmmm I could see a future where we move all the TypeScript configs, build tooling etc into `frontend/`, in which case having `src/` would make that migration simpler. Also, `src/` matches our convention for the existing Python code.

I'll leave it for now since it's not a blocker, to keep this review simpler. We can re-visit in the future as we migrate more of the UI to Vue. :+1:

zzzeid: >Also, src/ matches our convention for the existing Python code

That's mainly a convention for the repo/project, rather than an individual package inside the repo, though this is not an easy problem to solve. Happy to punt this to the future since we need to revisit this structure anyway.

zzzeid (approved):

PR #1255, opened 2026-06-15
uplift: add Vue train-selector widget for uplift target selection (bug 2044143)

cgsheeh:
Add a train-selector widget that progressively enhances the Django-rendered
"Uplift repositories" checkboxes in the uplift request modal. The native
checkboxes remain the source of truth for the form submission, so the backend
is unchanged -- the widget only reads and toggles them, and falls back to the
plain checkboxes if it cannot load.

When the "Request Uplift" button is first clicked, the widget fetches
release-train guidance from the whattrainisitnow.com API (see bug 2045812) and
validates the response with a Zod schema. That response describes the current
state of the release cycle (which version is on each train, whether betas
remain, whether the release candidate has shipped); the held response drives
every recommendation, landing description, and warning the widget produces.

The page CSP's `connect-src` is widened to the API origin, which a `csp_origin`
helper derives from the configured URL.

The widget presents two tabs:

- "Select Firefox Version": choosing a target version resolves -- based on the
  current cycle state -- to the correct uplift train(s) (beta, release, or
  both), ticks the matching repositories, and shows a single line describing
  where the patch will land.
- "Select Uplift Train": manual repository selection with per-train landing
  hints that reflect the cycle state, including a warning when beta is selected
  after betas have closed.

The tabs are keyboard- and screen-reader-accessible.

Builds on the Vue/TypeScript frontend bootstrap (bug 2047613).


cgsheeh:
Screenshots:

Screenshot 2026-06-12 at 11-39-45 D3997- Lando - Mozilla
Screenshot 2026-06-12 at 11-41-23 D3997- Lando - Mozilla
Screenshot 2026-06-12 at 11-41-43 D3997- Lando - Mozilla
Screenshot 2026-06-12 at 11-42-43 D3997- Lando - Mozilla

API could not retrieve data from whattrainisitnow.com:

Screenshot 2026-06-12 at 11-46-41 D3997- Lando - Mozilla

cgsheeh:
Note: the PR has test failures due to being formatted with the changes from https://github.com/mozilla-conduit/lando/pull/1251. I figured using that formatting would make this easier to review and reduce churn once we merge that PR.

frontend/src/trainGuidance.ts
```diff
@@ -0,0 +1,228 @@
...
+ * the train `<->` repository mapping. Only `beta` and `release` participate in
+ * uplift recommendations; `nightly` is included for completeness since patches
+ * landing there go through autoland, not uplift. `as const satisfies` keeps the
+ * repository names as literal types so `RepoName` can be derived from them.
+ */
+export const TRAIN_REPOS = {
+    nightly: "firefox-main",
+    beta: "firefox-beta",
+    release: "firefox-release",
+} as const satisfies Record<Train, string>;
```
shtrom: Could we lift this into a stand-alone, more config-like file, so we can find it more easily when we need to add to it?

frontend/src/trainGuidance.ts
```diff
@@ -0,0 +1,228 @@
+/**
+ * Translates whattrainisitnow.com release-train guidance into recommendations
+ * for Lando's uplift repository checkboxes. The data is keyed by mainline train
+ * (`nightly`, `beta`, `release`); see bug 2045812 for the response shape.
+ */
+
+import { z } from "zod";
+
+export type Train = "nightly" | "beta" | "release";
```
shtrom, line 9: Not sure about how thunderbird handles this, but would we just need to add something like `| "thunderbird-nightly" | ... ` if/when we need to?

I wonder if this could be lifted into a config-like file too.

Barring that maybe we should bring the declaration of `TRAIN_REPOS` to here, so everything that may need update is co-located at the top of the file.

frontend/src/trainGuidance.ts
```diff
@@ -0,0 +1,228 @@
...
+    beta: "firefox-beta",
+    release: "firefox-release",
+} as const satisfies Record<Train, string>;
+
+/** The Lando repository names, derived from `TRAIN_REPOS`. */
+export type RepoName = (typeof TRAIN_REPOS)[Train];
+
+/**
+ * Reverse lookup of `TRAIN_REPOS`. A linear scan over three entries avoids
+ * maintaining a second mapping.
```
shtrom: But. But. PERFORMANCE!

frontend/src/trainGuidance.ts
shtrom: ... of course, I don't know if Thunderbird has whatTrainIsIt, so this may not happen soon.

frontend/src/trainGuidance.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    if (train === "release") {
+        switch (stage) {
+            case "beta-shipping":
+                return `the next Firefox ${schedule.release.version} dot release`;
+            case "rc-shipping":
+                return `Firefox ${schedule.beta.version}.0 (major release)`;
+            case "dot-releases-only":
+                return `the next Firefox ${schedule.beta.version} dot release`;
```
shtrom: Could we make the version numbers pop out? Maybe bold or strong or something? It looks a bit subdued on the screenshot, but it feels like important info.

shtrom (changes requested):
LGTM (though I'm a bit over my head about Vue 😅).

Blocking for discussion point, but barring those, I have no issue.

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+</script>
+
+<template>
+    <div class="block">
+        <p v-if="status === 'loading'" class="help is-info">
+            Loading release schedule…
+        </p>
+        <p v-else-if="status === 'error'" class="help is-warning">
+            Could not load release-train guidance. Select repositories manually below.
+        </p>
```
zzzeid: This would be better as a single `<p>` element with the class and content set based on the response result. We could probably simplify this even further by interpreting the response directly, but this would be a step towards that:

```html
<p v-if="loadingMessage" class="help" :class="{'is-info': isLoading, 'is-warning': isErrored>
    {{ loadingMessage }}
</p>
```

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+        if (!response.ok) {
+            throw new Error(`Unexpected response status ${response.status}.`);
+        }
+
+        const result = releaseScheduleSchema.safeParse(await response.json());
+        if (!result.success) {
+            throw new Error(
+                `Train guidance response had an unexpected shape: ${result.error.message}`,
+            );
+        }
```
zzzeid: What's the point of throwing these exceptions here? Why not just check the response directly and set the error message? This seems redundant.

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+} from "../trainGuidance";
+import { useUpliftRepositories } from "../composables/useUpliftRepositories";
+
+const props = withDefaults(
+    defineProps<{ apiUrl: string; managedRepos?: RepoName[] }>(),
+    { managedRepos: () => [TRAIN_REPOS.beta, TRAIN_REPOS.release] },
+);
+
+/** Current state of the API response retrieval. */
+const status = ref<"loading" | "ready" | "error">("loading");
```
zzzeid, line 26: I think it would be better for these to be booleans. E.g., `isLoading`, `isErrored` etc... This would simplify the templates and avoid having to do string comparisons. `isReady` would also automatically be `!isLoading && !isError`.

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+const selectionSummary = computed(() => {
+    const repos = selectedRepos.value;
+    if (!repos) {
+        return "";
+    }
+
+    const labels = repos
+        .map((repo) => trainForRepo(repo))
+        .filter((train): train is Train => train !== null)
+        .map((train) => train.charAt(0).toUpperCase() + train.slice(1));
```
zzzeid: May be cleaner to use CSS for this to alter the display, rather than modifying the values themselves.

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+    if (labels.length === 0) {
+        return "";
+    }
+
+    if (labels.length === 1) {
+        return `Selected the ${labels[0]} uplift train.`;
+    }
+
+    const last = labels[labels.length - 1];
+    return `Selected the ${labels.slice(0, -1).join(", ")} and ${last} uplift trains.`;
```
zzzeid: Can you add comments here describing the expected state / what is being done?

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+const manualGuidance = computed(() =>
+    schedule.value
+        ? summarizeRepos(repositories.checkedRepos.value, schedule.value)
+        : { landing: "", warnings: [] },
+);
+
+/**
+ * The native checkbox field is shown in manual mode, and whenever the guidance
+ * is unavailable so the form remains usable.
+ */
```
zzzeid, line 130: I think this is a little hacky. These inputs should probably be changed to be hidden inputs on the rendered page, or better yet, put into a data attribute since they are no longer needed. We can generate equivalent checkboxes here if needed.

nit: there may be a better term to use that "native" here.

frontend/src/composables/useUpliftRepositories.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    /** Show or hide the entire native checkbox field. */
+    setFieldVisible(visible: boolean): void;
+}
+
+/**
+ * Bridge the Vue widget to the server-rendered "Uplift repositories" checkboxes.
+ * Those native inputs remain the source of truth for the Django form
+ * submission; this composable reads their state and toggles them so the form
+ * keeps working with JavaScript disabled or if the guidance fetch fails.
```
zzzeid: We don't need to support the case when JavaScript is disabled.

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+const props = withDefaults(
+    defineProps<{ apiUrl: string; managedRepos?: RepoName[] }>(),
+    { managedRepos: () => [TRAIN_REPOS.beta, TRAIN_REPOS.release] },
+);
+
+/** Current state of the API response retrieval. */
+const status = ref<"loading" | "ready" | "error">("loading");
+
+/** Current mode in the component. */
+const mode = ref<"version" | "manual">("version");
```
zzzeid, line 29: Same goes for this. Avoiding string comparisons would be better, if possible.

frontend/src/components/TrainSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Fetch the schedule the first time the modal is opened, then leave it cached.
+onMounted(() => {
+    openButton = document.querySelector(".uplift-request-open");
+    openButton?.addEventListener("click", loadSchedule, { once: true });
+});
+
+onUnmounted(() => {
+    openButton?.removeEventListener("click", loadSchedule);
```
zzzeid: Why is this event listener removed here? I assume this is what you mean by "leave it cached" above?

From a UX perspective this seems not great, since the content is only fresh the first time it's loaded, and there's not much to be gained by preventing this from getting updated if the modal is reopened.

frontend/src/trainGuidance.ts
```diff
@@ -0,0 +1,228 @@
...
+ * maintaining a second mapping.
+ *
+ * @param repo - A Lando repository name.
+ * @returns The train the repository belongs to, or `null` if it is not a
+ *   mainline repository.
+ */
+export function trainForRepo(repo: string): Train | null {
+    const match = Object.entries(TRAIN_REPOS).find(([, name]) => name === repo);
+    return match ? (match[0] as Train) : null;
+}
```
zzzeid, line 90: This seems a little hacky. Can we not just invert the mapping?

frontend/src/trainGuidance.ts
```diff
@@ -0,0 +1,228 @@
...
+): RepoGuidance {
+    const phrases = repos
+        .map((repo) => landingPhrase(repo, schedule))
+        .filter((phrase): phrase is string => phrase !== null);
+
+    const warnings = repos
+        .map((repo) => landingWarning(repo, schedule))
+        .filter((warning): warning is string => warning !== null);
+
+    const landing = phrases.length ? `This will land in ${joinWithAnd(phrases)}.` : "";
```
zzzeid: Would it not make more sense to have these as a bullet list, vs. comma-separated?

src/lando/middleware.py
```diff
@@ -38,8 +39,17 @@ def __call__(self, request: WSGIRequest) -> HttpResponse:
...
         response["X-Content-Type-Options"] = "nosniff"
 
+        # The uplift train-selector widget fetches release guidance from an
+        # external API, so `connect-src` must allow that origin in addition to
+        # `'self'` (otherwise it falls back to `default-src 'self'`).
+        connect_src = "connect-src 'self'"
+        train_api_origin = csp_origin(settings.WHATTRAINISITNOW_UPLIFT_TRAIN_API_URL)
+        if train_api_origin:
+            connect_src = f"{connect_src} {train_api_origin}"
+
```
zzzeid: `connect_src` is potentially beings set twice here. It should be set only once.

src/lando/middleware.py
```diff
@@ -70,6 +80,18 @@ def __call__(self, request: WSGIRequest) -> HttpResponse:
...
+    """Return the `scheme://host` origin of `url` for use in a CSP directive.
+
+    A relative or same-origin URL has no distinct origin to allow and yields an
+    empty string, so the caller can omit it and fall back to `'self'`.
+    """
+    parts = urlsplit(url)
+    if parts.scheme and parts.netloc:
+        return f"{parts.scheme}://{parts.netloc}"
+    return ""
+
```
zzzeid: I think the empty string handling / fallback should happen in the caller. Otherwise this method is not doing anything specific to CSP.

src/lando/middleware.py
```diff
@@ -38,8 +39,17 @@ def __call__(self, request: WSGIRequest) -> HttpResponse:
         response["X-Frame-Options"] = "DENY"
         response["X-Content-Type-Options"] = "nosniff"
 
+        # The uplift train-selector widget fetches release guidance from an
+        # external API, so `connect-src` must allow that origin in addition to
+        # `'self'` (otherwise it falls back to `default-src 'self'`).
+        connect_src = "connect-src 'self'"
+        train_api_origin = csp_origin(settings.WHATTRAINISITNOW_UPLIFT_TRAIN_API_URL)
```
zzzeid: Not ideal to be doing this for production, before we actually have a production URL. So I would say this should be explicitly disabled in production until we have a production URL.

zzzeid (changes requested):
First pass, did not review the tests thoroughly but can do that on the second pass.

frontend/src/trainGuidance.ts
cgsheeh: Yeah, we can cross that bridge when Thunderbird has a similar API available. :+1:

frontend/src/components/TrainSelector.vue
cgsheeh: I disagree that this would be better as several independent booleans. If we move to independent booleans, we open up the possibility of entering an illegal state due to a programming error (for example, nothing would prevent `isLoading` and `isReady` from both being true). Keeping this as a typed union allows one of three possible states, enforced by TypeScript at compile time. The type hint makes this variable closer to a Rust enum than a generic string.

We could achieve better ergonomics in the template by adding `computed` booleans which derive from the status so the templates can use `isLoading` etc, like this:

```typescript
const isLoading = computed(() => status.value === "loading");
const isErrored = computed(() => status.value === "error");
const isReady = computed(() => status.value === "ready");
```

However I'll note that string comparisons for typed unions are typical in TypeScript, as the compiler will catch typos and other incorrect values when comparing against string literals.

frontend/src/components/TrainSelector.vue
cgsheeh: I'd prefer to keep the mode as a typed union to match the above.

frontend/src/components/TrainSelector.vue
cgsheeh: Any of these code paths which result in a thrown exception are caught below, the widget status is set to `error` and the mode set to `manual`. Setting the state/mode in each failure state would be duplicating the "API request hit an error, fall back to the regular 'Select repositories'" logic. So essentially, it's putting the fallback mechanism in one place instead of three.

frontend/src/components/TrainSelector.vue
cgsheeh: This would be a moderate design change. The goal of this PR was to leave the backend Django-forms rendering as-is and only write a component that handles the `whattrainisitnow.com` API response, presents the new helper widget to the user and selects the existing options on the user's behalf. Changing the fields to hidden forces this widget to manage more state, and have to un-hide the inputs in case we enter the `error` state. I'd rather leave that management to happen when we create a new uplift-specific page, at which point we will remove the Django-forms dependency altogether.

Good point on the "native" terminology, I've updated this to "server-rendered".

frontend/src/trainGuidance.ts
cgsheeh: The suggestion widget will only ever suggest beta or release, so it would be a bulleted list with 1 or 2 elements. IMO it reads better as a single sentence describing where the uplift will land.

src/lando/middleware.py
cgsheeh: Updated to point to the production URL. Should have made this clearer, but the intention is to hold off on landing this until the production URL is live. :)

src/lando/middleware.py
cgsheeh: `connect_src` is set only once, this line just appends the `train_api_origin` to the end of the original declaration. It is a little hacky though, let me know if there's a better way to go about this.

frontend/src/components/TrainSelector.vue
cgsheeh: The event listener being removed is standard lifecycle cleanup. The `once: true` bit above is what causes the request to be made only once and then cached. My thought is that if the API returns an error, the user would refresh the entire page instead of opening and closing the modal. The API response will change infrequently, so there's no need to re-fetch the content every time the modal is loaded.

frontend/src/components/TrainSelector.vue
cgsheeh: I gave this a shot but with the current setup it became a larger change than I had hoped. I'm going to look at doing this, combined with @shtrom's suggestion of making the version numbers "pop", as a follow-up.

frontend/src/trainGuidance.ts
cgsheeh: I didn't want to move it to a separate config file as it's just 3 repos, and unlikely to change in the medium/long-term. Having the mapping in this file keeps it close to the logic that uses it, which feels like a better choice.

I did move all the code that uses this variable closer to the definition, to improve discoverability. :)

frontend/src/trainGuidance.ts
cgsheeh: Done.

frontend/src/trainGuidance.ts
cgsheeh: I'm going to look at doing this as a follow-up.

frontend/src/components/TrainSelector.vue
cgsheeh: Done.

frontend/src/composables/useUpliftRepositories.ts
cgsheeh: Done.

src/lando/middleware.py
cgsheeh: Makes sense, done.

src/lando/middleware.py
shtrom: You could build a list and `" ".join()` it, but I don't know if there's a real benefit.

shtrom (approved):
lgtm

frontend/src/components/TrainSelector.vue
zzzeid: >We could achieve better ergonomics in the template by adding computed booleans which derive from the status so the templates can use isLoading etc, like this:

Whatever we can do to simplify the template (see my comment [here](https://github.com/mozilla-conduit/lando/pull/1255/changes#r3422172122)).

> as the compiler will catch typos and other incorrect values

Good to know, though main motivation for this suggestion is to reduce the need for new string literals whose meaning could be gleaned from the variable name. But I think in this case there is a larger design issue with the way the API calls are processed. Different states that need to be tracked are being munged into a string, when they should be separated into different stages:

- pre API call: the state/result is not determined
- during API call: the state/result is not determined
- post API call: the state/result is determined, state could be a success or an error state, which can be handled elsewhere.

I think within the scope of this PR what you have is fine, but I don't want to establish this convention particularly since the API call state is going to be a more global one once we implement this on a broader level.

frontend/src/components/TrainSelector.vue
zzzeid: > I'd prefer to keep the mode as a typed union to match the above.

Since it only has two possible values and you can glean the meaning of the value from the variable name (e.g., `isManualMode`), I don't see any benefit here, but won't block this since overall it's an improvement on what we have today.

Also want to flag that we don't need to establish a convention where every check is using typed unions/strings.

frontend/src/components/TrainSelector.vue
zzzeid: >I'd rather leave that management to happen when we create a new uplift-specific page

Sounds good.

frontend/src/components/TrainSelector.vue
zzzeid: I think this could be also resolved by proxying the calls via Lando, doing some of the preprocessing on the server side.

frontend/src/components/TrainSelector.vue
zzzeid: Ok I won't nit pick too much here, but this could also likely be simplified with proxying the requests via Lando (maybe in the future).

frontend/src/components/TrainSelector.vue
zzzeid: Not sure if `once: true` is necessary (might be too conservative) but I guess we can see if people end up finding it problematic.

src/lando/middleware.py
zzzeid: > connect_src is set only once, this line just appends...

Appending a value to a string variable is setting it to a different value (strings are [immutable](https://docs.python.org/3.13/reference/datamodel.html#immutable-sequences) in Python). I was suggesting this could be done in a way that avoids setting it to a value that can potentially be overridden in the next line. This is a nit but still nicer to not have a variable be overridden in this way.

frontend/src/components/TrainSelector.vue
zzzeid: This comment is not addressed. I think it would be nice for us to start using a template-centric approach here, so that we shape the payloads (where we can) to simplify the templates, rather than the other way around. This will make it easier to modularize the templates/components.

src/lando/middleware.py
zzzeid: Oops missed this

> let me know if there's a better way to go about this

Few different ways but here's one that does not set two different values:

```python
train_api_origin = csp_origin(settings.WHATTRAINISITNOW_UPLIFT_TRAIN_API_URL)
connect_src = f"connect-src 'self' {train_api_origin or ''}".strip()

**Review by zzzeid — COMMENTED** [review]
Meant to add this to a comment as well, but if we add a proxy API endpoint in Lando we can avoid having to modify CSP settings.

File: frontend/src/components/TrainSelector.vue (hunk 5 above)
Comment (cgsheeh) [reply frontend/src/components/TrainSelector.vue]: Added a `statusMessage` which is a `computed()` that returns a string for display in a single `<p>`, or nothing if the widget should be rendered.

**Comment by cgsheeh** [comment]
I added tracking on the `UpliftSubmission` model so we can see how often the widget is used once we roll this out. Just wanted to call that out as an extra addition on this iteration.

File: src/lando/api/tests/test_uplift.py [hunk 19]
```diff
@@ -272,6 +275,33 @@ def test_uplift_creation_fails_when_seeding_fails(
...
+    repo = repo_mc(scm_type=SCMType.GIT, name="firefox-beta", approval_required=True)
+    revision = create_patch_revision(123, patch=normal_patch(0))
+
+    url = reverse("uplift-page")
+    form_data = {
+        "source_revisions": [revision.revision_id],
+        "repositories": [repo.name],
+        "target_selection_method": "widget_version",
+    }
+    form_data |= CREATE_FORM_DATA
```
shtrom, line 294: TIL sugar for `.update`

shtrom (approved):
lgtm!

zzzeid (commented):
This branch needs to be updated as it's no longer showing only the changes in this branch (merging/rebasing should work). Currently it's showing changes already in main.

cgsheeh:
> This branch needs to be updated as it's no longer showing only the changes in this branch (merging/rebasing should work). Currently it's showing changes already in main.

@zzzeid this is fixed.

zzzeid:
> Meant to add this to a comment as well, but if we add a proxy API endpoint in Lando we can avoid having to modify CSP settings.

@cgsheeh @shtrom any thoughts about this? This will give us more control over the API response shape, error handling, and would simplify the implementation / streamline all client calls to be via Lando instead of third party sites.

Can be done in a follow up, but I think this would be good to consider.

cgsheeh:
> > Meant to add this to a comment as well, but if we add a proxy API endpoint in Lando we can avoid having to modify CSP settings.
> 
> @cgsheeh @shtrom any thoughts about this? This will give us more control over the API response shape, error handling, and would simplify the implementation / streamline all client calls to be via Lando instead of third party sites.
> 
> Can be done in a follow up, but I think this would be good to consider.

I think it depends on which external sites we'll be contacting. For something like `whattrainisitnow.com`, we can either ping SRE or RelMan for support, so I'm not worried about having to simplify or change the API response shape. Github/Phab have API key and rate limiting considerations, so keeping those going through the Lando backend makes sense to me.

Seems like a reasonable idea but we should approach it on a case-by-case basis as the need arises. :+1:

zzzeid (approved):

PR #1256, opened 2026-06-15
celery: pin worker concurrency to 2 (bug 2047652)

cgsheeh:
Celery defaulted its pool size to the node's CPU count, inflating baseline CPU
and tripping the celeryworker HPA. Pin it to 2 (env-overridable).


shtrom (approved):
Nice find!

PR #1257, opened 2026-06-18
admin: add `landing_strategy` and remove handover fields from admin (Bug 2048687)

cgsheeh:
Add the `landing_strategy` field to the `LandingJobAdmin`. While we're
here, remove the handover repo fields from the admin UI to clear up some
space.

### DIFF
--- a/src/lando/main/admin.py
+++ b/src/lando/main/admin.py
@@ -170,15 +170,12 @@ class LandingJobAdmin(JobAdmin):
         "landing_strategy",
         "is_pull_request_job",
         "target_repo",
-        "handover_repo",
-        "is_handed_over",
         "created_at",
         "updated_at",
     )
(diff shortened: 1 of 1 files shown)


src/lando/main/admin.py
```diff
@@ -179,12 +179,13 @@ class LandingJobAdmin(JobAdmin):
         "formatted_replacements",
         "is_pull_request_job",
         "is_handed_over",
+        "landing_strategy",
```
shtrom, line 179: nit: could you make handover_repo read-only, too?

shtrom (approved):
lgtm

src/lando/main/admin.py
cgsheeh: I removed it fully instead.

zzzeid (approved):

PR #1259, opened 2026-06-18
github, views: ignore bot requests (bug 2048724)

zzzeid:
- ignore webhook events triggered by lando bot
- this prevents a loop when lando updates the pull request

### DIFF
--- a/src/lando/api/tests/test_views.py
+++ b/src/lando/api/tests/test_views.py
@@ -1,5 +1,6 @@
 import hashlib
 import hmac
+import json
 from unittest import mock
 
 import pytest
@@ -91,14 +92,14 @@ def test__views__phabricator_auth_backend(
     phabdouble, client, user, user_phab_api_key, user_linked_to_phab, monkeypatch
 ):
     """Test that the Phabricator authentication backend behaves as expected."""
-    test = client.get("/__version__")
-    assert test.wsgi_request.user.is_anonymous
+    response = client.get("/__version__")
+    assert response.wsgi_request.user.is_anonymous
 
     # NOTE: due to limitations in phabdouble, the value of the token
     # is irrelevant here. This should be fixed in bug 2019413.
     headers = {"X-Phabricator-API-Key": user_phab_api_key}
-    test = client.get("/__version__", headers=headers)
-    assert test.wsgi_request.user.is_authenticated
+    response = client.get("/__version__", headers=headers)
+    assert response.wsgi_request.user.is_authenticated
 
 
 @pytest.mark.django_db(transaction=True)
@@ -111,8 +112,8 @@ def test__views__phabricator_auth_backend_unknown_phid(
     phabdouble.user(username="unknown_phab_user", email="unknown@example.com")
 
     headers = {"X-Phabricator-API-Key": user_phab_api_key}
-    test = client.get("/__version__", headers=headers)
-    assert not test.wsgi_request.user.is_authenticated, (
+    response = client.get("/__version__", headers=headers)
+    assert not response.wsgi_request.user.is_authenticated, (
         "A valid Phabricator token whose PHID and email do not match any local "
         "profile should not result in an authenticated request."
     )
@@ -132,8 +133,8 @@ def test__views__phabricator_auth_backend_email_fallback(
     )
 
     headers = {"X-Phabricator-API-Key": user_phab_api_key}
-    test = client.get("/__version__", headers=headers)
-    assert test.wsgi_request.user.is_authenticated, (
+    response = client.get("/__version__", headers=headers)
+    assert response.wsgi_request.user.is_authenticated, (
         "Email fallback should authenticate the user when the PHID is not yet stored."
     )
 
@@ -155,8 +156,8 @@ def test__views__phabricator_auth_backend_invalid_token(
     # or not. This should be fixed (see bug 2019413.)
 
     headers = {"X-Phabricator-API-Key": "INVALID_TOKEN"}
-    test = client.get("/__version__", headers=headers)
-    assert not test.wsgi_request.user.is_authenticated
+    response = client.get("/__version__", headers=headers)
+    assert not response.wsgi_request.user.is_authenticated
 
 
 @mock.patch("lando.api.views.GitHubAPIClient")
@@ -176,12 +177,12 @@ def test__views__pull_request_api_view__private_repo(github_api_client, client):
         scm_type=SCMType.GIT,
     )
 
-    test = client.get(f"/api/pulls/{repo.name}/1/landing_jobs")
-    assert test.status_code == 404
+    response = client.get(f"/api/pulls/{repo.name}/1/landing_jobs")
+    assert response.status_code == 404
 
     mock_github_api_client.repo_is_private = False
-    test = client.get(f"/api/pulls/{repo.name}/1/landing_jobs")
-    assert test.status_code == 200
+    response = client.get(f"/api/pulls/{repo.name}/1/landing_jobs")
+    assert response.status_code == 200
 
 
 @mock.patch("lando.api.views.GitHubAPIClient")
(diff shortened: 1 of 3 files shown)


cgsheeh (changes requested):
Can you elaborate on what problem we're experiencing and how this change resolves the problem? There's not enough detail in the bug or PR description for me to follow along here.

zzzeid:
> Can you elaborate on what problem we're experiencing and how this change resolves the problem? There's not enough detail in the bug or PR description for me to follow along here.

Elaborating on the bug summary, which states:
>Ignore webhook events triggered by lando bot

When the webhook callback is triggered (i.e., GitHub makes a request to Lando after a user edits a PR), Lando will update the PR. This results in another webhook call that we want to ignore.

src/lando/api/tests/test_views.py
cgsheeh, line 1: Nit: missing type hints on the new code in this file.

I know we have the exception for this case on `main` but I'm about to remove it in my [bug 2048990](https://bugzilla.mozilla.org/show_bug.cgi?id=2048990) stack.

src/lando/api/views.py
```diff
@@ -377,6 +378,7 @@ def generate_context(self, request: HttpRequest) -> dict[str, str | list[str]]:
         context["commit_body"] = self.pull_request.commit_body
         return context
 
+    @ignore_bot_sender
     @require_github_signature
```
cgsheeh: `require_github_signature` should come first in the ordering, so we verify the request actually came from Github before deciding to ignore it. If we added more logic later the incorrect ordering would become a footgun, since we wouldn't expect to be handling the request body before it's been authenticated.

src/lando/utils/github.py
```diff
@@ -777,3 +784,22 @@ def verify_github_signature(hmac_secret: str, payload: bytes, signature: str) ->
...
+def ignore_bot_sender(post: Callable) -> Callable:
+    """Decorator that drops requests that originate by bots."""
+
+    @functools.wraps(post)
+    def _post(view: View, request: WSGIRequest, *args, **kwargs) -> HttpResponse:
+        """Drop the request if a bot triggered the original webhook."""
+        BOT_SENDER_TYPE = "Bot"
+        try:
+            sender_type = json.loads(request.body)["sender"]["type"]
+        except (JSONDecodeError, KeyError, ValueError):
```
cgsheeh: We should catch `TypeError` here as well, in case the body is a valid JSON but a non-dict (ie `[]`, a raw number, etc). This would be uncaught here and cause a 500, which is possible given the decorator ordering issue above.

src/lando/utils/github.py
```diff
@@ -777,3 +784,22 @@ def verify_github_signature(hmac_secret: str, payload: bytes, signature: str) ->
...
     return hmac.compare_digest(expected_signature, signature)
+
+
+def ignore_bot_sender(post: Callable) -> Callable:
+    """Decorator that drops requests that originate by bots."""
+
+    @functools.wraps(post)
+    def _post(view: View, request: WSGIRequest, *args, **kwargs) -> HttpResponse:
+        """Drop the request if a bot triggered the original webhook."""
+        BOT_SENDER_TYPE = "Bot"
```
cgsheeh, line 802: Nit: this would be better as a module-level constant.

src/lando/utils/github.py
```diff
@@ -777,3 +784,22 @@ def verify_github_signature(hmac_secret: str, payload: bytes, signature: str) ->
     )
     expected_signature = "sha256=" + hash_object.hexdigest()
     return hmac.compare_digest(expected_signature, signature)
+
+
+def ignore_bot_sender(post: Callable) -> Callable:
+    """Decorator that drops requests that originate by bots."""
```
cgsheeh: Nit:

```suggestion
    """Decorator that drops requests that originate from bots."""
```

src/lando/api/tests/test_views.py
```diff
@@ -183,16 +184,21 @@ class TestViewsPullRequestUpdateWebHook:
...
     @pytest.fixture
     def hmac_headers(self):
-        def calculate_signature():
+        def calculate_signature(body=None):
+            if isinstance(body, dict):
+                body = json.dumps(body)
             _hmac = hmac.new(
                 self.hmac_secret.encode("utf-8"),
-                msg=b"--BoUnDaRyStRiNg--\r\n",
+                msg=body.encode("utf=8") or b"--BoUnDaRyStRiNg--\r\n",
```
cgsheeh: Typo here on `utf-8`:

```suggestion
                msg=body.encode("utf-8") or b"--BoUnDaRyStRiNg--\r\n",
```

src/lando/api/tests/test_views.py
```diff
@@ -377,3 +407,26 @@ def test__views__pull_request_update_webhook_no_warnings_or_blockers(
...
+        body,
+        expected_body,
+        client,
+        webhook_gh_client,
+        hmac_headers,
+        webhook_content,
+    ):
+        mock_github_api_client = webhook_gh_client(github_api_client, body)
+
+        test = client.post(
```
cgsheeh: Nit: `response` would be a better variable name here.

cgsheeh (changes requested):

src/lando/api/tests/test_views.py
zzzeid: Hmm I don't think we enforce type hints in tests, and not sure we should at this stage.

cgsheeh: Ah, I missed that this was in a test file, so it's fine to ignore.

That said, some of the type hints are trivial additions (like `is_bot: bool`). We don't require type hints for test code, but we should still add them when possible.

src/lando/utils/github.py
zzzeid: I think it's fine here, until we need to reuse it elsewhere (if we ever do).

src/lando/utils/github.py
zzzeid: It wouldn't hurt.

cgsheeh (approved):

src/lando/api/tests/test_views.py
zzzeid: >but we should still add them when possible

It's always possible.

You could argue that all type hints are trivial additions. I think until we collectively decide on whether we should require them in tests (I don't think that's necessary), it's up to whoever wants to add them, if they want to add them, to add them (on existing or new tests). So I would disagree with the "should" above.

src/lando/api/tests/test_views.py
```diff
@@ -272,8 +286,12 @@ def test__views__pull_request_update_webhook_warnings_and_blockers(
             "warnings": ["a warning"],
             "blockers": ["a blocker"],
         }
-
-        test = client.post("/api/pulls/git-repo/1/webhook", headers=hmac_headers())
+        response = client.post(
+            "/api/pulls/git-repo/1/webhook",
+            content := webhook_content(),
```
shtrom, line 413: This walrus doesn't seem necessary (though TIL that it works within function parameters).
```suggestion
            content=webhook_content(),
```

src/lando/api/tests/test_views.py
```diff
@@ -314,14 +332,20 @@ def test__views__pull_request_update_webhook_blockers_only(
...
         mock_github_api_client = webhook_gh_client(github_api_client, body)
         generate_warnings_and_blockers.return_value = {
             "warnings": [],
             "blockers": ["a blocker"],
         }
 
-        client.post("/api/pulls/git-repo/1/webhook", headers=hmac_headers())
+        client.post(
+            "/api/pulls/git-repo/1/webhook",
+            content := webhook_content(),
```
shtrom, line 467:
```suggestion
            content=webhook_content(),
```

src/lando/api/tests/test_views.py
```diff
@@ -354,14 +378,20 @@ def test__views__pull_request_update_webhook_no_warnings_or_blockers(
...
         mock_github_api_client = webhook_gh_client(github_api_client, body)
         generate_warnings_and_blockers.return_value = {
             "warnings": [],
             "blockers": [],
         }
 
-        client.post("/api/pulls/git-repo/1/webhook", headers=hmac_headers())
+        client.post(
+            "/api/pulls/git-repo/1/webhook",
+            content := webhook_content(),
```
shtrom, line 513:
```suggestion
            content=webhook_content(),
```

src/lando/api/tests/test_views.py
```diff
@@ -377,3 +407,26 @@ def test__views__pull_request_update_webhook_no_warnings_or_blockers(
...
+        client,
+        webhook_gh_client,
+        hmac_headers,
+        webhook_content,
+    ):
+        mock_github_api_client = webhook_gh_client(github_api_client, body)
+
+        response = client.post(
+            "/api/pulls/git-repo/1/webhook",
+            content := webhook_content(is_bot=True),
```
shtrom, line 549:
```suggestion
            content=webhook_content(is_bot=True),
```

shtrom (approved):
LGTM, though I think there are unecessary walruses.

src/lando/api/tests/test_views.py
zzzeid: > This walrus doesn't seem necessary (though TIL that it works within function parameters).
> Suggested change
>             content := webhook_content(),
>             content=webhook_content(),

`content` is used a couple lines down (`headers=hmac_headers(body=content)`). Removing the walrus would mean `webhook_content()` would be called twice.

Thoughts?

shtrom: Ah, I missed the later use. Yeah, in this case that makes complete sense.

PR #1262, opened 2026-06-19
ruff: annotate return types of protected and nested functions (bug 2048990)

cgsheeh:
Adds the return type annotations that `ANN202` requires, ahead of
removing the rule's exception from `ruff.toml`.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


zzzeid (approved):
Reminder to remove some of the extra stuff in the description before landing with Lando (this and other PRs).

shtrom (approved):

PR #1267, opened 2026-06-19
Update Thunderbird staging repo to infra-testing

dandarnell:
We've renamed our staging repo for Thunderbird to infra-testing. The current URL, although it redirects to infra-testing, does not work (Lando returns HTTP 422).


src/lando/utils/management/commands/create_environment_repos.py
```diff
@@ -223,10 +223,10 @@
 for branch in ["main", "beta", "release", "esr140"]:
```
shtrom, line 223: We only support the `main` branch on infra-testing.
```suggestion
for branch in ["main"]:
```

shtrom (changes requested):
I don't think this is strictly needed. The infra-testing repo is a wholly separate one, that we should create separateling.

PR #1270, opened 2026-06-23
uplift_worker: automatically create try push for uplift requests (bug 2043376)

Magnolia-Liu:
When an uplift request is created at the end of an uplift job, Lando should automatically initiate a try push to run automated tests on the target release trains. This will give the developer some initial feedback on whether their uplift will land or has any obvious breakage.

Initially we run a small set of tasks outlined in a hardcoded try_task_config.js file. More targeted tests to the default push can be added as a follow-up. Developers can also use Treeherder to add specific tasks as needed, starting from the Lando-initiated push.


src/lando/api/tests/test_uplift.py
```diff
@@ -850,6 +856,23 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
...
+    )
+    assert try_job.requester_email == user.email, (
+        "Try-push job should record the original requester."
+    )
+    assert try_job.target_commit_hash, "Try-push job should have a base commit hash."
+
+    # One revision per uplift revision, plus the try_task_config revision.
+    assert try_job.revisions.count() == len(revisions) + 1, (
+        "Try-push job should bundle the uplift revisions and the config revision."
+    )
```
cgsheeh, line 886: Looks great! :)

src/lando/try_api/api.py
```diff
@@ -152,32 +176,23 @@ def patches(
...
-                error,
+        try:
+            mapping_repo = get_commit_map(
+                patches_request.base_commit_vcs, repo.name, repo.scm_type
             )
+        except ValueError as exc:
+            status = 400
             return status, ProblemDetail(
-                title="CommitMap not found", detail=error, status=status
+                title="CommitMap not found", detail=exc, status=status
```
cgsheeh: Previously we would pass the `error` variable to as the `detail` kwarg, which was a `str`. We're now passing `exc`, which is a `ValueError`. You should wrap this in a `str()`, which will extract the error string out of the exception.
```suggestion
                title="CommitMap not found", detail=str(exc), status=status
```

src/lando/try_api/api.py
```diff
@@ -152,32 +176,23 @@ def patches(
...
+            target_commit_hash = get_commit_hash(
+                mapping_repo, target_commit_hash, repo.scm_type
             )
+        except ValueError as exc:
+            status = 400
             return status, ProblemDetail(
-                title="Error converting SCM commit IDs",
-                detail=error,
-                status=status,
+                title="Error converting SCM commit IDs", detail=exc, status=status
```
cgsheeh: Same comment about `str(exc)` as above.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -155,6 +160,80 @@ def apply_uplift_revision(revision: Revision):
...
+            for revision in job.revisions.all():
+                revisions.append(
+                    Revision.new_from_patch(
+                        raw_diff=revision.diff, patch_data=patch_data
+                    )
+                )
+
+            # create and append config file revisions
+
+            with open("try_task_config.json", "r") as file:
```
cgsheeh: Opening the file with just the filename in this manner assumes the current working directory of the interpreter is this directory. This may be true in the tests, but there's no guarantee this is the case in production.

We should use `settings.BASE_DIR` to create a full path to the file, like: `try_config_path = settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"`.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -155,6 +160,80 @@ def apply_uplift_revision(revision: Revision):
...
+                raw_diff = file.read()
```
cgsheeh: An important thing to note here is that the file you're reading is a `.json` file, but the `Revision` model and variable name indicate we're expecting a Git-format diff file. You need to create a diff file using the JSON file.

src/lando/api/tests/test_uplift.py
```diff
@@ -745,6 +747,10 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
 
     repo = repo_mc(SCMType.GIT, name="firefox-beta", approval_required=True)
 
+    try_repo = repo_mc(SCMType.GIT, name="try", is_try=True)
+    (tmp_path / "try_task_config.json").write_text(normal_patch(0))
```
cgsheeh: This line is hiding the JSON/diff issues, as well as the issue with reading the JSON file using relative paths. :)

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -155,6 +160,80 @@ def apply_uplift_revision(revision: Revision):
         job.status = JobStatus.LANDED
         job.save()
 
+        try:
```
cgsheeh, line 169: This `try/except` block is quite large. It would improve readability if the logic that creates the try landing job was in a dedicated function, so we could do something like:

```suggestion
        try:
            self.create_uplift_try_push(...)
        except Exception as exc:
            ...
```

This makes it clear to the reader that we attempt to create the job and ignore exceptions when it fails. It would also allow us to keep the uplift-try-push code separate and de-dented.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
...
+            revisions.append(
+                Revision.new_from_patch(raw_diff=revision.diff, patch_data=patch_data)
+            )
+
+        # create and append config file revisions
+        try_config_path = (
+            settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"
+        )
+        with open(try_config_path, "r") as file:
+            config_contents = file.read()
```
cgsheeh: Since `try_config_path` is a `pathlib.Path` object, we can use the `Path` APIs to get the content of the file instead:


```suggestion
        try_config_path = (
            settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"
        )
        config_contents = try_config_path.read_text()
```

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
             )
             job.transition_status(JobAction.FAIL, message=message)
             raise PermanentFailureException(message) from exc
+
+    def create_uplift_try_push(
+        self, base_revision: str, repo_scm_type: str, job: UpliftJob
+    ) -> None:
+        try_repo = Repo.objects.get(name="try")
+        target_commit_hash = base_revision
```
cgsheeh: Nit: you can name the variable `target_commit_hash` in the method signature instead, to avoid this re-assignment.

src/lando/api/legacy/workers/try_task_config.json
```diff
@@ -0,0 +1,276 @@
+{
+    "parameters": {
+        "optimize_target_tasks": false,
+        "try_task_config": {
+            "disable-pgo": true,
+            "env": {
+                "TRY_SELECTOR": "fuzzy"
+            },
+            "tasks": [
```
shtrom, line 9: This is a very bespoke list. It may be worth having a way to determine it dynamically, or at least make it configurable via the admin UI.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
...
+        config_lines = config_contents.splitlines()
+        diff_header_lines = [
+            "diff --git a/try_task_config.json b/try_task_config.json",
+            "new file mode 100644",
+            "--- /dev/null",
+            "+++ b/try_task_config.json",
+            f"@@ -0,0 +1,{len(config_lines)} @@",
+        ]
+        added_lines = [f"+{line}" for line in config_lines]
+        raw_diff = "\n".join(diff_header_lines + added_lines) + "\n"
```
shtrom: This could be a separate method.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
...
+            "author_name": "Lando",
+            "author_email": job.requester_email,
+            "commit_message": "try_task_config",
+            "timestamp": str(int(time.time())),
+        }
+
+        try_revision = Revision.new_from_patch(
+            raw_diff=raw_diff, patch_data=try_patch_data
+        )
+        revisions.append(try_revision)
```
shtrom: This too, maybe wrapping the previous one.

src/lando/api/tests/test_uplift.py
```diff
@@ -745,6 +748,15 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
 
     repo = repo_mc(SCMType.GIT, name="firefox-beta", approval_required=True)
 
+    try_repo = repo_mc(SCMType.HG, name="try", is_try=True)
+
+    mapped_hg_hash = "a" * 40
+    CommitMap.objects.create(
+        git_hash="c82f7a174ed50727f951a6ea4aaee192eba1787a",
```
shtrom: Where does that come from? 😆

shtrom: FWIW, there's a `commit_maps` fixture https://github.com/mozilla-conduit/lando/blob/24a9d11d360b1fd93025df7b54975b98c8150c9f/src/lando/conftest.py#L1514-L1525

I'm not sure if it's useful here. It mainly depends on whether the value of that git_hash needs to be specific or not, for this test.

I guess that hash is the tip of what's on `repo`? In this case, I guess the fixture is not useful.

src/lando/api/legacy/workers/try_task_config.json
marco-c: We could generate this list using `./mach try fuzzy -q "^build-" --no-push`.
@ahal other ideas?

cgsheeh: Yes, this is planned as a follow-up: https://bugzilla.mozilla.org/show_bug.cgi?id=2050634

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -155,6 +161,14 @@ def apply_uplift_revision(revision: Revision):
         job.status = JobStatus.LANDED
         job.save()
 
+        try:
+            self.create_uplift_try_push(base_revision, repo.scm_type, job)
```
cgsheeh: We don't log anything indicating we created the try landing job here. Let's update `create_uplift_try_push` to return the new `LandingJob`, and add a log line in an `else` block here to indicate this was successful.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
...
+        for revision in job.revisions.all():
+            patch_data = {
+                "author_name": revision.author_name,
+                "author_email": revision.author_email,
+                "commit_message": revision.commit_message,
+                "timestamp": revision.timestamp,
+            }
+            revisions.append(
+                Revision.new_from_patch(raw_diff=revision.diff, patch_data=patch_data)
+            )
```
cgsheeh: This is using the patches from the `Revision` objects. However in uplift jobs, we try to do `git cherry-pick` of the revisions onto the uplift branches, where the content may be slightly different after application due to 3-way merges.

For correctness sake, we should create the new `Revision` objects from the newly created commits in Git. We should be able to use `GitSCM.get_patch_helpers_for_commits()` for this instead. :)

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
...
+
+        for revision in job.revisions.all():
+            patch_data = {
+                "author_name": revision.author_name,
+                "author_email": revision.author_email,
+                "commit_message": revision.commit_message,
+                "timestamp": revision.timestamp,
+            }
+            revisions.append(
+                Revision.new_from_patch(raw_diff=revision.diff, patch_data=patch_data)
```
cgsheeh: `Revision.new_from_patch` will create the new `Revision` objects directly. We should move the revision creation into the transaction.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
...
+            )
+
+        # create and append config file revisions
```
cgsheeh: Please use proper English (punctuation, capitalization, etc) for all your comments. The existing codebase is consistent about that, and it's good practice in general. :)

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +294,87 @@ def run_moz_phab_uplift(
             )
             job.transition_status(JobAction.FAIL, message=message)
             raise PermanentFailureException(message) from exc
+
+    def create_uplift_try_push(
+        self, base_revision: str, repo_scm_type: str, job: UpliftJob
+    ) -> None:
```
cgsheeh: We should add a docstring to this method.

cgsheeh (changes requested):

src/lando/api/tests/test_uplift.py
Magnolia-Liu: You're correct, the hash is specific to the tip of main. I'll change that to be fetched dynamically and also add a comment describing what the map is for :)

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -280,3 +305,98 @@ def run_moz_phab_uplift(
...
+        try_patch_data = {
+            "author_name": "Lando",
+            "author_email": requester_email,
+            "commit_message": "try_task_config",
+            "timestamp": str(int(time.time())),
+        }
+        raw_diff = self.create_try_diff_from_json()
+        return Revision.new_from_patch(raw_diff=raw_diff, patch_data=try_patch_data)
+
+    def create_revisions_from_patch_helpers(
```
cgsheeh: Nit: this creates just a single `Revision` from a single `PatchHelper`, we should update the method name to indicate that.

```suggestion
    def create_revision_from_patch_helper(
```

src/lando/api/tests/test_uplift.py
```diff
@@ -745,6 +748,17 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
 
     repo = repo_mc(SCMType.GIT, name="firefox-beta", approval_required=True)
 
+    try_repo = repo_mc(SCMType.HG, name="try", is_try=True)
+
+    # Hardcode the mapping between the Git base commit to it's Hg equivalent that would exist in the database in production so the lookup succeeds.
```
cgsheeh: Nit: this comment line is long, please break it into multiple lines.


```suggestion
    # Hardcode the mapping between the Git base commit to its Hg
    # equivalent that would exist in the database in production so the 
    # lookup succeeds.
```

src/lando/api/tests/test_uplift.py
```diff
@@ -730,6 +732,7 @@ def test_uplift_worker_applies_patches_and_creates_uplift_revision_success_git(
     monkeypatch,
     make_uplift_job_with_revisions,
     mock_uplift_email_tasks,
+    tmp_path,
```
cgsheeh: Looks like this is leftover from a previous state of the PR, we should remove it before landing.

cgsheeh (changes requested):
Looking good, we should remove the unused `tmp_path` fixture and then this should be good to land!

I'm going to deploy this to dev and do a quick test to see it in action as well. :)

cgsheeh (approved):

PR #1271, opened 2026-06-25
CommitMap: update Thunderbird mapping to use thunderbird-desktop (bug 2050335)

shtrom:
This is the basename of the Git repository, which is what the rest of the code uses for lookups.

Note: this will change the Git<->Hg mapping API URLs from, e.g., 

  https://lando.moz.tools/api/git2hg/thunderbird/6ec3d8940682b616d316eb30a6d6a6dcde1a3f18

to 

  https://lando.moz.tools/api/git2hg/thunderbird-desktop/6ec3d8940682b616d316eb30a6d6a6dcde1a3f18
                                                ^^^^^^^^

This will require a data migration to be run after deployment:

     lando migrate_data 2_bug2050335_update_commit_map_thunderbird_desktop

---

 * CommitMap: update Thunderbird mapping to use thunderbird-desktop (bug 2050335)
 * migrate_data: update `thunderbird` CommitMap records to `thunderbird-desktop`

---
Lando: [link](https://lando.moz.tools/pulls/lando/1271/)
Bugzilla: [bug 2050335](https://bugzilla.mozilla.org/show_bug.cgi?id=2050335)


shtrom:
Putting this out as a quick-fix for critique. This should work as desired, but is a bit smelly.

Alternatively, we could add a new attribute to the Repo model to allow to determine the mapping repo.

@zzzeid @cgsheeh thoughts?

cgsheeh:
> Alternatively, we could add a new attribute to the Repo model to allow to determine the mapping repo.

Do you mean mapping both `thunderbird` and `thunderbird-desktop` to `comm-unified`? Seems simpler than migrating every row in the DB for a slight rename on one field, so I'm on board with that.

cgsheeh (dismissed):
See my other comment. :)

shtrom:
Superseded by #1322

shtrom:
Reopening as we are still considering this approach.

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -97,3 +98,31 @@ def migrate_1_bug1971103_add_firefox_required_automation_permissions(
...
+            self.stdout.write("No CommitMaps entries for `thunderbird`.")
+            raise SystemExit()
+
+        self._get_confirmation(
+            ask_confirm,
+            f"Updating {len(tbird_commit_maps)} CommitMaps to use `thunderbird-desktop` instead of `thunderbird`.",
+            "",
+        )
+
+        tbird_commit_maps = CommitMap.objects.raw("""
```
zzzeid: Variable being set here but not for a known purpose.

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -97,3 +98,31 @@ def migrate_1_bug1971103_add_firefox_required_automation_permissions(
...
+        tbird_commit_maps = CommitMap.objects.filter(git_repo_name="thunderbird")
+        if not tbird_commit_maps:
+            self.stdout.write("No CommitMaps entries for `thunderbird`.")
+            raise SystemExit()
+
+        self._get_confirmation(
+            ask_confirm,
+            f"Updating {len(tbird_commit_maps)} CommitMaps to use `thunderbird-desktop` instead of `thunderbird`.",
+            "",
+        )
```
zzzeid, line 124: Strictly speaking this is not necessarily accurate: the query below is updating all matching records, but this count is for a point in time.

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -97,3 +98,31 @@ def migrate_1_bug1971103_add_firefox_required_automation_permissions(
...
+
+        tbird_commit_maps = CommitMap.objects.raw("""
+            UPDATE main_commitmap
+            SET git_repo_name='thunderbird-desktop'
+            WHERE git_repo_name='thunderbird'
+            """)
```
zzzeid: Curious why not use the [ORM](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#bulk-update) here?

zzzeid (dismissed):
Some nits but lgtm. May need a short maintenance window to run these after deployment.

src/lando/utils/management/commands/migrate_data.py
zzzeid: (by the way CommitMap objects are excluded from audit log tracking, in case that was a consideration)

shtrom: Wasn't aware of `bulk_update` 😅

src/lando/utils/management/commands/migrate_data.py
shtrom: True. It's mainly for getting an idea of the scale.

src/lando/utils/management/commands/migrate_data.py
shtrom: Updated!

zzzeid (approved):

PR #1272, opened 2026-06-25
api.views: retrofit webhook to determine PR/repo from POST (bug 2024944)

zzzeid:
- update test to include additional payload from webhook
- override dispatch method on webhook view

### DIFF
--- a/src/lando/api/tests/test_views.py
+++ b/src/lando/api/tests/test_views.py
@@ -7,7 +7,6 @@ import pytest
 from django.test import Client
 
 from lando.main.models import Repo, SCMType
-from lando.utils.github import PullRequest
 
 
 @pytest.fixture
@@ -326,26 +325,15 @@ class TestViewsPullRequestUpdateWebHook:
 
     @pytest.fixture
     def webhook_gh_client(self):
-        def wrapper(github_api_client, body):
-            data = {
-                "number": 1,
-                "title": "this is some title (bug 1111111, bug 2222222)",
-                "body": body,
-            }
-            mock_pr_data = mock.MagicMock()
-            mock_pr_data.__getitem__.side_effect = lambda k: (
-                data[k] if k in data else mock.MagicMock()
-            )
-
+        def wrapper(github_api_client):
             mock_github_api_client = mock.MagicMock()
             mock_github_api_client.repo_is_private = False
-
-            mock_pr = PullRequest(mock_github_api_client, mock_pr_data)
-            mock_github_api_client.build_pull_request.return_value = mock_pr
             github_api_client.return_value = mock_github_api_client
 
             repo = Repo.objects.create(
                 name="git-repo",
+                default_branch="test_branch",
+                url="https://example.org/test-org/test-repo.git",
                 scm_type=SCMType.GIT,
                 pr_enabled=True,
             )
@@ -355,9 +343,27 @@ class TestViewsPullRequestUpdateWebHook:
         return wrapper
 
     @pytest.fixture
-    def webhook_content(self):
-        def _webhook_content(is_bot=False):
-            return {"sender": {"type": "User" if not is_bot else "Bot"}}
+    def webhook_content(self, pull_request_data, update_dict):
+        def _webhook_content(is_bot=False, overrides=None):
+            data = {
+                "sender": {"type": "User" if not is_bot else "Bot"},
+                "pull_request": pull_request_data(
+                    **{
+                        "number": 1,
+                        "base": {
+                            "ref": "test_branch",
+                            "repo": {
+                                "clone_url": "https://example.org/test-org/test-repo.git"
+                            },
+                        },
+                    }
+                ),
+            }
+
+            if overrides:
+                update_dict(data, overrides)
+
+            return data
 
         return _webhook_content
 
@@ -372,17 +378,24 @@ class TestViewsPullRequestUpdateWebHook:
         client,
         expected_body,
         webhook_gh_client,
+        webhook_content,
         hmac_headers,
     ):
         """Test that the webhook fails when called without correct headers."""
-        mock_github_api_client = webhook_gh_client(github_api_client, body)
+        content = webhook_content(overrides={"pull_request": {"body": body}})
+        mock_github_api_client = webhook_gh_client(github_api_client)
 
         generate_warnings_and_blockers.return_value = {
             "warnings": ["a warning"],
             "blockers": ["a blocker"],
         }
 
-        response = client.post("/api/pulls/git-repo/1/webhook")
+        response = client.post(
+            "/api/pulls/webhook",
+            content,
+            content_type="application/json",
+            headers={},
+        )
 
         assert mock_github_api_client.update_pull_request_body.call_count == 0
         assert response.status_code == 403
@@ -402,15 +415,16 @@ class TestViewsPullRequestUpdateWebHook:
         webhook_content,
     ):
         """Test that the webhook is calling the GitHub API with the correct parameters."""
-        mock_github_api_client = webhook_gh_client(github_api_client, body)
+        content = webhook_content(overrides={"pull_request": {"body": body}})
+        mock_github_api_client = webhook_gh_client(github_api_client)
 
         generate_warnings_and_blockers.return_value = {
             "warnings": ["a warning"],
             "blockers": ["a blocker"],
         }
         response = client.post(
-            "/api/pulls/git-repo/1/webhook",
-            content := webhook_content(),
+            "/api/pulls/webhook",
+            content,
             content_type="application/json",
             headers=hmac_headers(body=content),
         )
@@ -456,15 +470,16 @@ class TestViewsPullRequestUpdateWebHook:
         hmac_headers,
         webhook_content,
     ):
-        mock_github_api_client = webhook_gh_client(github_api_client, body)
+        content = webhook_content(overrides={"pull_request": {"body": body}})
+        mock_github_api_client = webhook_gh_client(github_api_client)
         generate_warnings_and_blockers.return_value = {
             "warnings": [],
             "blockers": ["a blocker"],
         }
 
         client.post(
-            "/api/pulls/git-repo/1/webhook",
-            content := webhook_content(),
+            "/api/pulls/webhook",
+            content,
             content_type="application/json",
             headers=hmac_headers(body=content),
         )
@@ -502,15 +517,16 @@ class TestViewsPullRequestUpdateWebHook:
         hmac_headers,
         webhook_content,
     ):
-        mock_github_api_client = webhook_gh_client(github_api_client, body)
+        content = webhook_content(overrides={"pull_request": {"body": body}})
+        mock_github_api_client = webhook_gh_client(github_api_client)
         generate_warnings_and_blockers.return_value = {
             "warnings": [],
             "blockers": [],
         }
 
         client.post(
-            "/api/pulls/git-repo/1/webhook",
-            content := webhook_content(),
+            "/api/pulls/webhook",
+            content,
             content_type="application/json",
             headers=hmac_headers(body=content),
         )
@@ -542,11 +558,13 @@ class TestViewsPullRequestUpdateWebHook:
         hmac_headers,
         webhook_content,
     ):
-        mock_github_api_client = webhook_gh_client(github_api_client, body)
-
+        content = webhook_content(
+            overrides={"pull_request": {"body": body}}, is_bot=True
+        )
+        mock_github_api_client = webhook_gh_client(github_api_client)
         response = client.post(
-            "/api/pulls/git-repo/1/webhook",
-            content := webhook_content(is_bot=True),
+            "/api/pulls/webhook",
+            content,
             content_type="application/json",
             headers=hmac_headers(body=content),
         )
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -1,8 +1,10 @@
 import json
+import logging
 from collections import defaultdict
 from datetime import datetime
 from functools import wraps
-from typing import Callable
+from json.decoder import JSONDecodeError
+from typing import Any, Callable
 
 from django import forms
 from django.conf import settings
@@ -47,6 +49,8 @@ from lando.utils.github_checks import (
 )
 from lando.utils.landing_checks import LandingChecks
 from lando.utils.phabricator import PHABRICATOR_API_KEY_HEADER, get_phabricator_client
+
+logger = logging.getLogger(__name__)
 
 
 class APIView(View):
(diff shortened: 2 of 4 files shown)


src/lando/api/views.py
```diff
@@ -353,6 +357,35 @@ def get(
 class PullRequestUpdateWebhook(PullRequestAPIView):
     """API method called by GitHub to update Lando data in pull request."""
 
+    @staticmethod
+    def _get_dispatch_args(request_body: str) -> tuple[str]:
```
cgsheeh: Type hint is incorrect: 
```suggestion
    def _get_dispatch_args(request_body: str) -> dict[str, Any]:
```

src/lando/api/views.py
```diff
@@ -353,6 +357,35 @@ def get(
...
+        """Return arguments that will be redirected to dispatch."""
+        POST = json.loads(request_body)
```
cgsheeh: The uppercase naming makes me think this overwrites a global/class-level `POST` variable somewhere. If it does, we should leave a comment explaining why. If it doesn't, perhaps `request_json` would be a better name, since all-caps variable names are reserved for constants.

src/lando/urls.py
```diff
@@ -127,7 +127,7 @@
         name="api-pull-request-checks",
     ),
     path(
-        "api/pulls/<str:repo_name>/<int:pull_number>/webhook",
+        "api/pulls/webhook",
```
cgsheeh, line 136: Is this future-proof if we need to add another Github->Lando webhook endpoint? The URL seems quite generic, perhaps it could be `api/pulls/pr_description/webhook` to keep the purpose scoped to the PR description box changes.

If this is gleaned from the body that's okay too, haven't set up a GH webhook before myself.

src/lando/api/views.py
```diff
@@ -353,6 +357,35 @@ def get(
...
+    def dispatch(self, request: WSGIRequest) -> JsonResponse:
+        """Fetch repo_name from data and call super."""
+        try:
+            dispatch_args = self._get_dispatch_args(request.body)
+        except (
+            JSONDecodeError,
+            IndexError,
+            KeyError,
+            Repo.DoesNotExist,
+        ) as e:
```
cgsheeh: `IndexError` isn't possible here, since all the accesses are dict-key with `str`. We should catch `TypeError` instead, in case something returns `None` and we try to do a dict-lookup on it.

For completeness we should also catch `Repo.MultipleObjectsReturned` as there's no DB-level guarantee that `.get()` on `url` and `default_branch` will return a single row. Perhaps we should add a `UniqueConstraint` on those two fields, which would also create an index for faster lookups.

src/lando/api/views.py
```diff
@@ -353,6 +357,35 @@ def get(
...
+    @staticmethod
+    def _get_dispatch_args(request_body: str) -> tuple[str]:
+        """Return arguments that will be redirected to dispatch."""
+        POST = json.loads(request_body)
+        pull_number = POST["pull_request"]["number"]
+        branch = POST["pull_request"]["base"]["ref"]
+        clone_url = POST["repository"]["clone_url"]
+        return {
+            "pull_number": pull_number,
+            "repo_name": Repo.objects.get(url=clone_url, default_branch=branch).name,
```
cgsheeh: This `repo_name` parameter is passed to dispatch, where it's used in a `Repo.objects.get(name=repo_name)` query. Can we pass the `Repo` object to avoid a double-lookup (maybe passing `"repo"` in this dict and skipping the lookup if it exists)?

src/lando/api/views.py
```diff
@@ -353,6 +357,35 @@ def get(
...
+        }
+
+    def dispatch(self, request: WSGIRequest) -> JsonResponse:
```
cgsheeh, line 407: Type hint should be `HttpResponse` here, since we don't return `JsonResponse` when something is wrong with the webhook body.

cgsheeh (dismissed):

src/lando/api/views.py
zzzeid: Oh yeah this is from the previous implementation which I modified.

src/lando/api/views.py
zzzeid: >there's no DB-level guarantee that .get() on url and default_branch will return a single row. Perhaps we should add a UniqueConstraint on those two fields, which would also create an index for faster lookups

Yeah with the current state of git in Lando this is a good idea.


>IndexError isn't possible here, since all the accesses are dict-key with str.

Happy to remove it, this entire block is not doing much other than preventing GitHub from getting 500s or 400s. It is technically possible, if the webhook returns valid JSON but one of the dictionary items contains a string instead of a dict.

>We should catch TypeError instead, in case something returns None and we try to do a dict-lookup on it.

This is kind of related to the generality of the webhook. I'll add it, everything we add here is really to suppress errors to GitHub so as long as we log them should be OK.

src/lando/urls.py
zzzeid: The webhook is repo-scoped. Currently only used for PRs. If we are to expand the webhook scope we'll have to basically do something in the dispatcher, because there are no dynamic parameters in the webhook config in GH, and the branch name (which along with the repo URL) corresponds to a unique repo in Lando, so that is also dynamic.

> If this is gleaned from the body that's okay too

Short answer, not future proof but it is gleaned from the body. I think it's out of scope for this stack, but most likely we'll have to rename the view to `RepoWebhook` at a point in the future, and delegate from there based on the type of event.

shtrom: It looks like GitHub adds additional `X-GitHub` headers to webhooks payloads https://docs.github.com/en/webhooks/webhook-events-and-payloads#example-webhook-delivery

We could perhaps check that `X-GitHub-Hook-Installation-Target-Type` is `pull-request` (value to be checked), and otherwise ignore it.

src/lando/api/views.py
shtrom: There's this https://docs.djangoproject.com/en/5.2/ref/request-response/#django.http.HttpRequest.POST, but I don't think it's used in the same way, particularly as the `POST` here is not on `self`.

src/lando/api/views.py
shtrom: I think we should pass the whole `request`, too, rather than just the body. This way we can inspect headers, as per my other comment on `X-GitHub` headers.

src/lando/api/views.py
shtrom: Maybe we should return a Json object on error paths?

src/lando/api/views.py
shtrom: Could we move the payload exception handling into `_dispatch_args`, and return/raise a more predictable value on error?

src/lando/api/views.py
zzzeid: Yeah that's the convention I was going with here, but I think I'll just rename this to `post_data`.

src/lando/api/views.py
zzzeid: >Maybe we should return a Json object on error paths?

Yes this should return `JsonResponse({} .status=202)` instead. Will be fixed more broadly as part of improving base classes.

We won't have a clean approach here until those bases classes are improved, since parts of the code raise `Http404` which will be `HttpResponse` objects even in an `APIView`. But that's part of the middleware and is not returned by `dispatch` so it's not relevant here.

src/lando/urls.py
zzzeid: We would need to check `X-Github-Event`. `X-Github-Hook-Installation-Target-Type` is still a repository. I'll file a separate bug for this.

```
X-Github-Event: pull_request
...
X-Github-Hook-Installation-Target-Type: repository
```

Filed [this](https://bugzilla.mozilla.org/show_bug.cgi?id=2053406).

src/lando/api/views.py
zzzeid: No longer relevant as I refactored the method to only return the PR data.

src/lando/api/views.py
zzzeid: Modified this so we are not making unnecessary lookups, or API calls (also reuses pull request data in webhook, rather than fetching pull request data from scratch).

src/lando/api/views.py
zzzeid: Modified the handling so that `ValueError` is raised when when fetching data, and caught/handled in `dispatch`.

src/lando/urls.py
shtrom: Ah, yep, typo; definitely meant `Event` but copied the wrong entry 😅

src/lando/api/views.py
```diff
@@ -354,37 +354,42 @@ def get(
 
 
 @method_decorator(csrf_exempt, name="dispatch")
-class PullRequestUpdateWebhook(PullRequestAPIView):
+class PullRequestUpdateWebhook(View, PrivateRepoPermissionMixin):
```
shtrom, line 391: nit: What's the motivation for this bypassing the `PullRequestAPIView`?

We still use the `target_repo`, `client` and `pull_request` attributes. I think we could just keep that inheritance, and `@override` the `dispatch` method

src/lando/api/views.py
```diff
@@ -350,9 +354,43 @@ def get(
...
+                default_branch=pull_request_data["base"]["ref"],
+            )
+        except (Repo.DoesNotExist, Repo.MultipleObjectsReturned) as e:
+            raise Http404 from e
+
+        self.client = GitHubAPIClient(self.target_repo.url)
+        self.raise_404_if_needed(request, self.client)
+        self.pull_request = PullRequest(self.client, pull_request_data)
+
+        return super().dispatch(request)
```
shtrom, line 426: Ah. Do we want to not call the `PullRequestAPIView.dispatch()` method, instead going for the `View`'s? I'd prefer to keep the `PullRequestAPIView` inheritance, but it may require too much surgery to be worth it...

The cleanest I can think of is to add an optional `Repo`, and `PullRequest` arguments to `PullRequestAPIView.dispatch()`, so it doesn't try to reconstruct its own (the `client` and `raise_404_and_needed` could be delegated to the `super()`)

src/lando/api/views.py
```diff
@@ -350,9 +354,43 @@ def get(
...
+        try:
+            self.target_repo = Repo.objects.get(
+                url=pull_request_data["base"]["repo"]["clone_url"],
+                default_branch=pull_request_data["base"]["ref"],
+            )
+        except (Repo.DoesNotExist, Repo.MultipleObjectsReturned) as e:
+            raise Http404 from e
+
+        self.client = GitHubAPIClient(self.target_repo.url)
+        self.raise_404_if_needed(request, self.client)
```
shtrom, line 423: (The` name of this method is a bit misleading (: `needed` here means `unauthorised`, but this is not relevant to this PR)

This is a webhook context here, is there a django user in context? I suspect there isn't, because that would be coming from GitHub, so it may always throw 404s for private repos (maybe that's what we want? though it may degrade the UX when working on secure bugs).

OTOH, it doesn't look like we do any authentication on this webhook endpoint. Should there be some sort of signature verification, to be sure this is coming from GitHub?

src/lando/api/tests/test_views.py
```diff
@@ -233,9 +221,25 @@ def wrapper(github_api_client, body):
...
+    def webhook_content(self, pull_request_data, update_dict):
+        def _webhook_content(is_bot=False, overrides=None):
+            data = {
+                "sender": {"type": "User" if not is_bot else "Bot"},
+                "pull_request": pull_request_data(
+                    **{
+                        "number": 1,
+                        "base": {
+                            "ref": "test_branch",
+                            "repo": {"clone_url": "https://github.com/test-repo.git"},
```
shtrom: nit: No `owner` in the URL, this may come back to bite us on some URL parsing.

src/lando/api/tests/test_views.py
```diff
@@ -204,26 +203,15 @@ def _headers(signature="", body=None):
...
             mock_github_api_client.repo_is_private = False
-
-            mock_pr = PullRequest(mock_github_api_client, mock_pr_data)
-            mock_github_api_client.build_pull_request.return_value = mock_pr
             github_api_client.return_value = mock_github_api_client
 
             repo = Repo.objects.create(
                 name="git-repo",
+                default_branch="test_branch",
+                url="https://github.com/test-repo.git",
```
shtrom: ditto re: no owner

src/lando/conftest.py
```diff
@@ -1681,3 +1681,78 @@ def _mock_response(
...
+def pull_request_data(update_dict) -> Callable:
+    def _pull_request_data(**overrides):
+        data = {}
+        data["base"] = {}
+        data["base"]["user"] = {}
+        data["head"] = {}
+        data["head"]["repo"] = {}
+        data["user"] = {}
+
+        data["url"] = "https://example.org/"
```
shtrom: nit: using example.org may trip some URL parsing that tries to see if it's a GitHub URL.

Perhaps this fixture should have a dedicated `tld` that defaults to `example.org`, but can be ovewritten wholesale, i.e., to `github.com`, when necessary?

shtrom (approved):
Couple of nits. See, particularly, the discussion about dropping the `PullRequestAPIView` to avoid it's `dispatch` logic, vs extending it instead.

But LGTM as is.

src/lando/api/views.py
zzzeid: The main reason is to reduce duplication of logic and API calls/db lookups, while maintaining the existing dispatcher functionality in PullRequestAPIView.

An alternate approach which I considered was to maintain the existing class inheritance, and basically overriding `dispatch` in the same way, but it would mean: (1) we would have to modify `PullRequestAPIView.dispatch` to have all arguments be optional, and (2) the logic that uses `PrivateRepoPermissionMixin` would be a little funky because in `PullRequestAPIView.dispatch`, it is done before `self.pull_request` is set, whereas in this override it would be after (since we already have that info). Definitely think there is room for improvement here, but it would require some changes to `PullRequestAPIView.dispatch`.

src/lando/api/views.py
zzzeid: > This is a webhook context here, is there a django user in context? I suspect there isn't, because that would be coming from GitHub, so it may always throw 404s for private repos (maybe that's what we want? though it may degrade the UX when working on secure bugs).

Known issue with regards to the 404s for private repos (not sure I filed a bug but will double check). This would have to be handled in the future (most likely by doing some authentication based on the HMAC secret). As it stands, all these requests are unauthenticated, but they are authorized via the `require_github_signature` decorator. This happens after the request is dispatched to the appropriate method.

>OTOH, it doesn't look like we do any authentication on this webhook endpoint. Should there be some sort of signature verification, to be sure this is coming from GitHub?

The authorization happens on the `post` method, in line with the view conventions. However you bring up a good point which would have been more of an issue with the dispatch method if we were still calling github to build the pull request (before verifying the signature).

src/lando/api/views.py
zzzeid: I basically went back and forth between the two. I agree we need a cleaner solution for the long term, but I think there's enough of a difference at this time between the webhook and non-webhook endpoints (mainly that the webhook gets a lot more information in the request, vs. everything else) that we can go with the separate inheritance.


>The cleanest I can think of is to add an optional Repo, and PullRequest arguments to PullRequestAPIView.dispatch(), so it doesn't try to reconstruct its own (the client and raise_404_and_needed could be delegated to the super())

I actually did try that but see my note in the other comment re: pull request building/setting. We could definitely consolidate that but I think it's out of scope. We can probably roll it into general improvements in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2053406).

src/lando/conftest.py
zzzeid: Adjusted the existing test to use example.org, and modified the fixture to be changeable.

PR #1274, opened 2026-06-26
landing_checks: add backend validation to warning acknowledgements for pull requests (bug 2000955)

Magnolia-Liu:
Currently, the Lando UI only enables/disables the post landing job button based on the checkbox, without validating if there are any additional checks that have been created since the page loaded.

Upon clicking the landing button, check if the warnings have changed since the page loaded. If so, alert user of the change, display new warnings, and prompt again for warning acknowledgement.

Video of UI:

https://github.com/user-attachments/assets/af8a7b1b-b290-4ad2-a013-7ab228a7cb0e

---
Lando: [link](https://lando.moz.tools/pulls/lando/1274/)
Bugzilla: [bug 2000955](https://bugzilla.mozilla.org/show_bug.cgi?id=2000955)

### DIFF
--- a/src/lando/api/tests/test_views.py
+++ b/src/lando/api/tests/test_views.py
@@ -162,6 +162,7 @@ def test__views__phabricator_auth_backend_invalid_token(
 @mock.patch("lando.api.views.GitHubAPIClient")
 @pytest.mark.django_db(transaction=True)
 def test__views__pull_request_api_view__private_repo(github_api_client, client):
+
     mock_github_api_client = mock.MagicMock()
     mock_pr = mock.MagicMock()
 
@@ -561,3 +562,78 @@ class TestViewsPullRequestUpdateWebHook:
         )
         assert response.status_code == 202
         assert mock_github_api_client.update_pull_request_body.call_count == 0
+
+
+@mock.patch("lando.api.views.generate_warnings_and_blockers")
+@mock.patch("lando.api.views.GitHubAPIClient")
+@pytest.mark.django_db(transaction=True)
+@pytest.mark.parametrize(
+    "warnings_1, warnings_2, expected_status, expected_response",
+    [
+        (["warning-1", "warning-2"], ["warning-1", "warning-2"], 201, b""),
+        (
+            ["warning-1", "warning-2"],
+            ["warning-3", "warning-4"],
+            400,
+            [
+                "The warnings present when the request was constructed have changed. Please acknowledge the new warnings and try again."
+            ],
+        ),
+    ],
+)
+def test__views_landing_job_pull_request_view__warnings(
+    github_api_client,
+    mock_warnings_and_blockers,
+    authenticated_client,
+    repo_mc_github_api_client,
+    repo_mc,
+    warnings_1,
+    warnings_2,
+    expected_status,
+    expected_response,
+):
+    repo = repo_mc(SCMType.GIT)
+    github_api_client.return_value = repo_mc_github_api_client
+
+    mock_pr = mock.MagicMock()
+    repo_mc_github_api_client.build_pull_request.return_value = mock_pr
+    mock_pr.author = ("Test Author", "test@email.com")
+    mock_pr.commit_message = "Test Commit Message"
+    mock_pr.number = 1
+    mock_pr.head_sha = "aaa123"
+    mock_pr.base_sha = "bbb123"
+    mock_pr.patch = "diff --git a/abc b/def\n"
+    mock_pr.reviews_summary = {}
+
+    mock_warnings_and_blockers.return_value = {
+        "warnings": warnings_1,
+        "blockers": [],
+    }
+
+    old_warnings = authenticated_client.get(
+        f"/api/pulls/{repo.name}/1/checks",
+        content_type="application/json",
+    ).json()["warnings"]
+
+    mock_warnings_and_blockers.return_value = {
+        "warnings": warnings_2,
+        "blockers": [],
+    }
+
+    response = authenticated_client.post(
+        f"/api/pulls/{repo.name}/1/landing_jobs",
+        data={
+            "head_sha": "aaa123",
+            "base_sha": "bbb123",
+            "pull_number": 1,
+            "old_warnings": old_warnings,
+        },
+        content_type="application/json",
+    )
+
+    assert response.status_code == expected_status
+    if expected_status == 400:
+        new_warnings = response.json()["new_warnings"]
+
+        assert new_warnings == mock_warnings_and_blockers.return_value["warnings"]
+        assert response.json()["errors"] == {"warnings": expected_response}
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -285,24 +285,44 @@ class LandingJobPullRequestAPIView(PullRequestAPIView):
         class Form(forms.Form):
             """Simple form to get clean some fields."""
 
+            def clean(self):
+
+                cleaned_data = self.cleaned_data
+                new_warnings = cleaned_data["new_warnings"]
+                old_warnings = cleaned_data["old_warnings"]
+                if sorted(new_warnings) != sorted(old_warnings):
+                    self.errors["warnings"] = [
+                        "The warnings present when the request was constructed have changed. "
+                        "Please acknowledge the new warnings and try again."
+                    ]
+
+                return cleaned_data
+
             head_sha = forms.CharField()
+            new_warnings = forms.JSONField()
+            old_warnings = forms.JSONField()
             # TODO: use this for verification later, see bug 1996571.
             # base_ref = forms.CharField()
 
         ldap_username = request.user.email
 
-        blockers = generate_warnings_and_blockers(
+        warnings_and_blockers = generate_warnings_and_blockers(
             self.target_repo, self.pull_request, request
-        )["blockers"]
+        )
+        new_warnings = warnings_and_blockers["warnings"]
 
-        if blockers:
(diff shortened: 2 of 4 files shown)


src/lando/api/views.py
```diff
@@ -268,24 +268,43 @@ def post(
...
+        warnings_and_blockers = generate_warnings_and_blockers(
             self.target_repo, self.pull_request, request
-        )["blockers"]
+        )
+        warnings, blockers = (
+            warnings_and_blockers["warnings"],
+            warnings_and_blockers["blockers"],
+        )
 
         if blockers:
```
zzzeid: nit: it may be cleaner to do this here, vs. up above.
```suggestion
        if blockers := warnings_and_blockers["blockers"]:
```

src/lando/api/views.py
```diff
@@ -268,24 +268,43 @@ def post(
         class Form(forms.Form):
             """Simple form to get clean some fields."""
 
+            def clean(self):
+
+                cleaned_data = self.cleaned_data
+                warnings = cleaned_data["warnings"]
+                old_warnings = cleaned_data["old_warnings"]
+                if warnings != old_warnings:
+                    self.errors["warnings"] = ["mismatch"]
```
zzzeid: Is there some sorting/ordering that could make this check a little more robust?

src/lando/api/views.py
```diff
@@ -268,24 +268,43 @@ def post(
...
+            warnings_and_blockers["blockers"],
+        )
 
         if blockers:
-            # Pull request has blockers that prevent it from landing.
             return JsonResponse({"errors": blockers}, status=400)
 
-        form = Form(json.loads(request.body))
+        data = json.loads(request.body)
+        data["warnings"] = warnings
```
zzzeid: Adding a comment here describing why we are including the new warnings in the form would be helpful.

src/lando/api/views.py
zzzeid: nit: It would be nice if `mismatch` was reworded to be a little more specific. In fact, we could return the full error message that will get displayed to the user when this error is triggered, and pick that up in the UI, instead of hiding/showing an existing element.

In this way, we would also be able to say how many new errors are there, or maybe even identify which ones are new.

src/lando/api/views.py
```diff
@@ -323,7 +342,7 @@ class Form(forms.Form):
         job.status = JobStatus.SUBMITTED
         job.save()
 
-        return JsonResponse({"id": job.id}, status=201)
+        return JsonResponse({"id": job.id, "warnings": warnings}, status=201)
```
zzzeid: What's the purpose of adding `warnings` here? It should always be an empty list.

zzzeid (changes requested):

src/lando/api/views.py
Magnolia-Liu: I initially returned the new warnings so that I could check in test__views_landing_job_pull_request_view__warnings if they matched the old warnings to make sure the mismatch check in the clean function wasn't passing false negatives. Now that you mention it though it seems unnecessary to include if it's only used for testing purposes.

src/lando/api/views.py
```diff
@@ -268,24 +268,43 @@ def post(
...
-        form = Form(json.loads(request.body))
+        data = json.loads(request.body)
+        # add new warnings to the data so that the form can validate that they match the old warnings
+        data["warnings"] = warnings
+        form = Form(data)
 
         if not form.is_valid():
-            return JsonResponse(form.errors, 400)
+            return JsonResponse(
+                {"errors": form.errors, "warnings": warnings}, status=400
```
shtrom: `errors` and `warnings` here feel like they are not at the same level: `errors` (and the nested `errors.warnings`) are validation errors on the request, while `warnings` here is additional information about warnings related to the PR. 

I wonder if renaming to, say `new_warnings` may make the intent clearer, as I see it's used to update warnings in the page on error.

shtrom: Perhaps we should send a 409 error code here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409; it would be easier to explicitly handle this error in the UI without risking processing unrelated errors.

shtrom: > Perhaps we should send a 409 error code

OTOH, looking at the UI code now, it handles all 400 gracefully, with an optional handling of `[new_]warnings` if they are present. Maybe changing the status code is  not worth it (:

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -217,22 +218,44 @@ $.fn.stack = function () {
...
-                        pull_request_button.prop("disabled", true);
+                        var result = await response.json();
                         pull_request_button
                             .removeClass("is-danger")
                             .removeClass("is-loading")
                             .addClass("is-warning");
-                        pull_request_button.html("Could not create landing job");
+                        pull_request_button.prop("disabled", true);
+
+                        if ("warnings" in result) {
```
shtrom: `new_warnings`

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -217,22 +218,44 @@ $.fn.stack = function () {
...
+                            $("#warnings-mismatch").append(
+                                `<li>${result["errors"]["warnings"]}</li>`,
+                            );
+                            $("#warnings").empty();
+                            $("#acknowledge-warnings").prop("checked", false);
+                            pull_request_button.html(
+                                "Acknowledge warnings to continue",
```
shtrom: Some of those message strings are duplicated. Could we factor them so only one instance of the string exists?

...

Looking further into this, this whole block is very similar https://github.com/mozilla-conduit/lando/pull/1274/changes#diff-4d791453e88ede62fd62f9e15ecccd2a546f9d0e6fe016091adb198d09a6ca67R187-R194 we should turn it into a separate method, say `need_warnings_acknowledgements` that we can call from the various call sites

shtrom (commented):
LGTM, but I think it needs a little bit of refactoring.

src/lando/static_src/legacy/js/components/Stack.js
```diff
@@ -1,5 +1,16 @@
 "use strict";
 
+// Label shown on the landing button when unacknowledged warnings block landing.
+const ACKNOWLEDGE_WARNINGS_LABEL = "Acknowledge warnings to continue";
```
shtrom, line 4: Probably not necessary now that it's only used once. However, don't remove it now! If anything, it's good practice to have those strings centralised in a single location, as the application gets bigger.

Anyway, nothing to do here (:

shtrom (approved):
lgtm!

src/lando/api/views.py
```diff
@@ -285,24 +285,43 @@ def post(
...
             """Simple form to get clean some fields."""
 
+            def clean(self):
+
+                cleaned_data = self.cleaned_data
+                new_warnings = cleaned_data["new_warnings"]
+                old_warnings = cleaned_data["old_warnings"]
+                if sorted(new_warnings) != sorted(old_warnings):
+                    self.errors["warnings"] = [
+                        "The warnings present when the request was constructed have changed. Please acknowledge the new warnings and try again."
```
zzzeid: nit / break string into two lines to avoid long line
```suggestion
                        "The warnings present when the request was constructed have changed. "
                        "Please acknowledge the new warnings and try again."
```

src/lando/api/tests/test_views.py
```diff
@@ -561,3 +562,105 @@ def test__views__pull_request_update_webhook_bot(
...
+@mock.patch("lando.api.views.generate_warnings_and_blockers")
+@mock.patch("lando.api.views.GitHubAPIClient")
+@pytest.mark.django_db(transaction=True)
+def test__views_landing_job_pull_request_view__warnings(
+    github_api_client, mock_warnings_and_blockers, authenticated_client, repo_mc
+):
+    repo = repo_mc(SCMType.GIT)
+    mock_github_api_client = mock.MagicMock()
+    github_api_client.return_value = mock_github_api_client
+    mock_github_api_client.repo_is_private = False
```
zzzeid: Could the `repo_mc_github_api_client` be used or adapted to be used here? (comment applies to other instances here)

src/lando/api/tests/test_views.py
```diff
@@ -561,3 +562,105 @@ def test__views__pull_request_update_webhook_bot(
...
+
+    mock_pr = mock_github_api_client.build_pull_request.return_value
```
zzzeid: Not sure this is correct here - normally you would set the `return_value` attribute to a value, rather than the other way around. In this case, `mock_pr` is basically a new `MagicMock`. (comment applies to other instances here)

zzzeid (changes requested):
Couple of small comments. Can you deploy this to develop before we merge + deploy?

src/lando/api/tests/test_views.py
```diff
@@ -561,3 +562,110 @@ def test__views__pull_request_update_webhook_bot(
...
+        content_type="application/json",
+    )
+
+    assert response.status_code == 201
+
+
+@mock.patch("lando.api.views.generate_warnings_and_blockers")
+@mock.patch("lando.api.views.GitHubAPIClient")
+@pytest.mark.django_db(transaction=True)
+def test__views_landing_job_pull_request_view__warnings_mismatch(
```
zzzeid: nit: There's very little difference between this test and `test__views_landing_job_pull_request_view__warnings`. I think this could benefit from parametrization.

zzzeid (approved):
lgtm with one nit.

PR #1275, opened 2026-06-26
treestatus: migrate Treestatus from old-Lando to new-Lando (bug 1984161)

cgsheeh:
Finish moving Treestatus into Lando by reading tree state from the local
database instead of the old external HTTP service. The legacy `TreeStatus`
client, its `TreeStatusDouble` test double, the `TREESTATUS_URL` setting, and
the worker startup ping are removed; `is_open` and the landing workers now
query the DB directly via the uncached `fetch_tree_by_name` so landing
decisions always reflect the latest committed state, while `get_tree_by_name`
keeps the cache for display surfaces. The Treestatus navbar button is un-hidden,
and a one-time `import_treestatus_data` management command is added to import
trees and their log history from an existing Treestatus instance (defaulting to
old-Lando).

---
Lando: [link](https://lando.moz.tools/pulls/lando/1275/)
Bugzilla: [bug 1984161](https://bugzilla.mozilla.org/show_bug.cgi?id=1984161)

:warning: This pull request has 9 warnings.


src/lando/treestatus/management/commands/import_treestatus_data.py
shtrom, line 1: I think we should land this first, separately.

src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -0,0 +1,96 @@
...
+
+        logger.debug(f"Fetching trees from {base_url}.")
+        trees_response = requests.get(f"{base_url}/trees")
+        trees_response.raise_for_status()
+        trees_data = trees_response.json()["result"]
+
+        if not trees_data:
+            raise CommandError(f"No trees returned from {base_url}.")
+
+        with transaction.atomic():
```
shtrom, line 56: Do we need a transaction over the loop? It feels a bit iffy. Maybe wrapping each tree import into the transaction would be sufficient?

shtrom: OTOH, I guess nothing else would be touching those trees during import, so it's probably not a big issue.

src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -0,0 +1,96 @@
...
+                self.import_tree(base_url, tree_info)
+
+        self.stdout.write(self.style.SUCCESS("Finished importing Treestatus data."))
+
+    def import_tree(self, base_url: str, tree_info: dict):
+        """Import a single tree and its complete log history."""
+        tree_name = tree_info["tree"]
+        if Tree.objects.filter(tree=tree_name).exists():
+            self.stdout.write(
+                self.style.WARNING(f"Tree {tree_name} already exists, skipping.")
```
shtrom: This may cause an issue if we need to catch-up from a previous import. Could we compare the timestamps of the last import to the current state before deciding?

src/lando/treestatus/utils.py
```diff
@@ -203,7 +206,17 @@ def tree_cache_key(tree_name: str) -> str:
 
 @cache_method(tree_cache_key, cache_alias=TREESTATUS_CACHE)
 def get_tree_by_name(tree_name: str) -> Optional[CombinedTree]:
-    """Retrieve a `CombinedTree` representation of a tree by name.
+    """Cached lookup of a tree by name.
```
shtrom, line 60: As we are now using a local DB, do we gain much from having a cache?

shtrom (changes requested):
LGTM, but I think we should split the PR in two:

1. a first with the import code, ideally able to work incrementally; to deploy and use prior to the migration
2. a second which replaces all the remote requests with DB queries; to deploy once we are satisfied the migration has stuck.

src/lando/treestatus/utils.py
zzzeid: Had part of this conversation offline. I agree that there's no need to use a cache and this is no longer relevant since there are no API calls. But minimizing initial code changes is fine, and using a DB cache optimized for that purpose won't cause any harm. I think ultimately this should not use a cache, but this could be done in a follow up.

src/lando/treestatus/utils.py
zzzeid, line 1: I think the name of this module is a little misleading. There are no views in this module and this is no longer an API in the conventional sense in this repo. Should be moved to something like `lando.treestatus.utils`. There are also things in `lando.treestatus.views.ui` which can be moved to `lando.treestatus.views`.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -39,6 +39,7 @@
     TreeClosed,
 )
 from lando.pushlog.pushlog import PushLog, PushLogForRepo
+from lando.treestatus.views.api import is_open
```
zzzeid: `is_open` should be moved outside of the `api` module (see comment in that module).

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -122,7 +123,7 @@ def run_job(self, job: LandingJob) -> bool:
                 "Pull Requests are not supported for this repository."
             )
 
-        if not self.treestatus_client.is_open(repo.tree):
+        if not is_open(repo.tree):
```
zzzeid, line 126: I think adding a property to the repo here would be helpful and simplify some of the other code (can be done in separate PR).
```suggestion
        if repo.tree_closed:
```

Can also add `repo.tree_open`.

zzzeid (changes requested):
Agree with @shtrom about splitting this PR. But also there are some quick wins we can achieve either as followups or putting this in a stack with some additional PRs (commented in place).

- Being able to deploy this and test it (maybe under a feature flag) before the migration would be good. 
- `suite` should be updated to accommodate these new changes so the workflow doesn't break for people using it.
- Are there migrations scripts that we need to add? These would be good to add to the stack.

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Handled in https://github.com/mozilla-conduit/lando/pull/1348.

src/lando/api/legacy/workers/landing_worker.py
cgsheeh: I took a crack at adding this but there is a circular import between `lando.main.models` and `lando.treestatus.models`. So we either have to import `is_open` in the instance method (not great), or resolve the circular import in the PR. I took a crack at the latter, but it added a lot of noise to the PR. Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2062488 to do this as a follow-up instead.

cgsheeh:
> Agree with @shtrom about splitting this PR. But also there are some quick wins we can achieve either as followups or putting this in a stack with some additional PRs (commented in place).
> 
>     * Being able to deploy this and test it (maybe under a feature flag) before the migration would be good.

I'm going to deploy and test this on dev this week.

>     * `suite` should be updated to accommodate these new changes so the workflow doesn't break for people using it.

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2062491 for this.

>     * Are there migrations scripts that we need to add? These would be good to add to the stack.

The data import management command PR should cover everything in terms of migrations. Otherwise the DB already includes all the new models. I won't be landing this until I've tested everything on dev, if I find other changes during testing I'll update the PR and re-request review.

shtrom:
Can we land https://github.com/mozilla-conduit/lando/pull/1348 now?

cgsheeh:
> This is looking pretty good.
> 
> The main thing that's missing (forgot to flag it in the original review though maybe we discussed offline?) is to add a maintenance exception to the treestatus namespace, which would match the legacy behaviour. Ideally we block POST requests during deployments that have migrations affecting treestatus models, but we can probably file a separate bug for that and just be mindful of those changes until then.

Done! :)

src/lando/tests/test_middlewares.py
```diff
@@ -23,6 +28,38 @@ def test_url_origin(url: str, expected: str):
...
+):
+    """The Treestatus API stays available while Lando is under maintenance."""
+    ConfigurationVariable.set(
+        ConfigurationKey.API_IN_MAINTENANCE, VariableTypeChoices.BOOL, "1"
+    )
+
+    response = client.get(path)
+
+    if available_in_maintenance:
+        assert response.status_code == 200, (
```
zzzeid: Not sure we should include this here, since technically it is also the case in `else` (I thought there was a bug filed for this somewhere but can't find it now). Basically we should be returning a `503` when maintenance is ON but that's not the case today).

src/lando/tests/test_middlewares.py
```diff
@@ -23,6 +28,38 @@ def test_url_origin(url: str, expected: str):
...
+@pytest.mark.parametrize(
+    "path,available_in_maintenance",
+    (
+        ("/treestatus/", False),
+        ("/api/treestatus/trees", True),
+        ("/api/treestatus/stack", True),
+    ),
+)
+@pytest.mark.django_db
+def test_maintenance_mode_treestatus_api_exception(
```
zzzeid: I think this test is generic enough to test all `excepted_namespaces` -- it would basically be the same test but with a different name and slightly modified parameters. This would also take care of [bug 2041036](https://bugzilla.mozilla.org/show_bug.cgi?id=2041036).

cgsheeh: Killing two bugs with one PR -- done. :)

src/lando/tests/test_middlewares.py
```diff
@@ -31,26 +31,36 @@ def test_url_origin(url: str, expected: str):
...
-    """The Treestatus API stays available while Lando is under maintenance."""
+    """Excepted namespaces and URL names stay available while Lando is under maintenance."""
     ConfigurationVariable.set(
         ConfigurationKey.API_IN_MAINTENANCE, VariableTypeChoices.BOOL, "1"
     )
 
     response = client.get(path)
 
     if available_in_maintenance:
-        assert response.status_code == 200, (
```
shtrom: Shouldn't we keep this?

src/lando/tests/test_middlewares.py
```diff
@@ -31,26 +31,36 @@ def test_url_origin(url: str, expected: str):
 @pytest.mark.parametrize(
     "path,available_in_maintenance",
     (
+        # Paths outside of the excepted namespaces and URL names.
+        ("/", False),
         ("/treestatus/", False),
```
shtrom, line 36: Should this one also be excluded to support backward compat?

src/lando/treestatus/tests/test_treestatus.py
```diff
@@ -84,12 +126,25 @@ def test_get_combined_tree(new_treestatus_tree):
...
+def test_api_is_served_under_api_treestatus(client, new_treestatus_tree):
+    """The API is mounted under `/api/treestatus/`, not at the base endpoint."""
+    new_treestatus_tree(tree="mozilla-central")
+
+    assert client.get("/api/treestatus/trees").status_code == 200, (
+        "The API should be served under `/api/treestatus/`."
+    )
+    assert client.get("/trees").status_code == 404, (
+        "The API should no longer be served from the base endpoint."
+    )
```
shtrom, line 139: Should we redirect rather than error out, for backward compat?

shtrom (commented):
Questions about backward compat, but LGTM otherwise.

shtrom:
Ah, and I think we should perform the migration prior to landing this PR.

PR #1276, opened 2026-06-30
lando: run the prettier style check from the Vitest suite (bug 2048013)

cgsheeh:
Move the prettier formatting check out of the Python tests and into the
Vitest suite so the JS test run reports whether the sources are correctly
formatted. Along the way, point the prettier lint path at the repository
root, excluding files it should not own via `.prettierignore`, and reformat
the now-covered files.

---
Lando: [link](https://lando.moz.tools/pulls/lando/1276/)
Bugzilla: [bug 2048013](https://bugzilla.mozilla.org/show_bug.cgi?id=2048013)

:warning: This pull request has 2 warnings.


shtrom (approved):
LGTM! 

Ideally the automatic changes should be in a separate commit for review, but GitHub did a good job at hiding whitespace changes.

It seems like a bunch of non-js files (md, yaml) got reformatted too, so perhaps prettier overshot a bit. OTOH, it doesn't hurt to have something looking after those (:

PR #1277, opened 2026-06-30
repos: add ff-test trunk and uplift repos to dev environment (Bug 2051749)

cgsheeh:
`ff-test-dev` is the autoland/trunk equivalent; `ff-test-beta` and
`ff-test-release` are approval-required uplift targets.

### DIFF
--- a/src/lando/utils/management/commands/create_environment_repos.py
+++ b/src/lando/utils/management/commands/create_environment_repos.py
@@ -160,6 +160,41 @@ REPOS = {
             "required_permission": SCM_CONDUIT,
             "push_target": "@",
         },
+        #
+        # `ff-test` branches stand in for Firefox's trunk and uplift branches.
+        # `ff-test-dev` is the autoland/trunk equivalent, while `ff-test-beta`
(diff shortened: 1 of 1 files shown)


shtrom:
The initial thought was to have ff-test-dev and ff-test-stage to target different repos depending on the lando environment, but we don't seem to have codified ff-test-stage, though it exists (likely manually provisioned) https://lando.allizom.org/admin/main/repo/3/change/

Could you add it for the stage env while we're at it? This then raises the question as to whether we should have `ff-test-dev-beta`, rather than `ff-test-beta`, but I won't lose sleep over it.

shtrom (approved):

zzzeid (approved):

shtrom (approved):
lgtm. Do we want ff-test-stage, and the uplift repos in staging, too?

cgsheeh:
> lgtm. Do we want ff-test-stage, and the uplift repos in staging, too?

Not sure as I don't use staging much, we can do those as follow-ups or as part of the main-on-staging work. :)

PR #1321, opened 2026-07-02
try backing

cgsheeh:
- **landing_job: add `git_branch` flag to push to a job-specific branch**
- **try_api: create a Git backing landing job alongside the Hg Try push**

---
Lando: [link](https://lando.moz.tools/pulls/lando/1321/)

:no_entry_sign: This pull request has 4 blockers.


src/lando/try_api/tests/test_api.py
```diff
@@ -46,6 +48,23 @@ def _post(
...
 
+@pytest.fixture
+def mocked_repo_config_try_git_backing(mocked_repo_config_try) -> Repo:
+    """Create the `try-backing` Git repo alongside the Hg `try` repo."""
+    return Repo.objects.create(
+        scm_type=SCMType.GIT,
+        name="try-backing",
+        url="http://git.test/try-backing",
+        push_path="http://git.test/try-backing",
+        pull_path="http://git.test/try-backing",
```
ahal, line 59: I think this will need to pull from `mozilla-firefox/firefox` right? So that it's up-to-date with latest main and includes whatever base revision the user pushes with.

Does Lando fetch from this repo prior to attempting to apply the patch?

src/lando/try_api/api.py
```diff
@@ -74,6 +85,171 @@ def get_commit_hash(
...
+) -> LandingJob:
+    """Create the Hg Try `LandingJob`, referencing the paired Git backing job.
+
+    When `git_job` is set, a reference revision recording its ID is added on top
+    of the push.
+    """
+    try_job = LandingJob.objects.create(
+        target_repo=try_repo,
+        requester_email=requester_email,
+        target_commit_hash=target_commit_hash,
```
ahal, line 247: This is the Hg hash right? We still need to somehow pass the Git hash into the push so CI knows which revision to clone from. Can that also be added to the GIT_BACKING_JOB_FILE?

src/lando/try_api/tests/test_api.py
cgsheeh: This config is just for the test fixture, where we have a `docker compose` setup and `git.test` is the hostname for the local Git container. In practice we will have a `Repo` object in the DB with the config you described (pull from `mozilla-firefox/firefox`, push to `mozilla-releng/try-backing`). That's the purpose of the `get_git_backing_repo` function in `src/lando/try_api/api.py`, which maps the `try` repo to the configured `try-backing` repo record.

To answer your question more directly: yes, we will need to pull from `mozilla-firefox/firefox`, and Lando does fetch the latest changes before attempting to apply the patch. :)

src/lando/try_api/api.py
cgsheeh: Good point, updated the backing file to include the Git base SHA.

src/lando/try_api/api.py
```diff
@@ -74,6 +85,179 @@ def get_commit_hash(
...
+
+    Adds a small JSON file on top of the Hg Try push referencing the Git backing
+    job, so the two pushes can be correlated and CI knows which Git revision and
+    branch to clone from. We add a new commit rather than rewriting the push's
+    existing `try_task_config.json`, whose opaque diff cannot be reliably edited
+    here.
+    """
+    contents = json.dumps(
+        {
+            "git_landing_job_id": git_job.id,
```
ahal, line 175: One thing I'm a bit confused on, what prints this ID in the Treeherder url to the terminal? Is it the Hg server itself?

ahal:
I really appreciate you putting this up!

But I think I'm leaning towards using native pushes again. Mainly due to the fact that it would require the Decision task to handle synchronization via polling, but also because I think this would take a few weeks to get tested, landed, deployed and a new worker spun up (at least if I'm driving it).

On the other side the native push implementation is already finished, and I think I came up with an acceptable solution for the people using native Hg repos to push to project branches.

cgsheeh:
Closing in favour of the native pushes route. :)

PR #1325, opened 2026-07-07
automation_worker: update pull request with backout information (Bug 2024925)

Magnolia-Liu:
In Phabricator, when a revision is backed out, it is updated with the commits and is reopened so it's clear the change was reverted.

Implement something similar: The AutomationWorker now detects reverts among the commits it lands and comments on the affected pull requests.

This is comment-only for now. Reopening the reverted pull request can be implemented in a follow-up bug.

---

Lando: [link](https://lando.moz.tools/pulls/lando/1325/)
Bugzilla: [bug 2024925](https://bugzilla.mozilla.org/show_bug.cgi?id=2024925)

:warning: This pull request has 16 warnings.


src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -25,13 +26,177 @@
...
 
+REVERT_SUMMARY_RE = re.compile(r"^Revert \"?(?P<summary>.*)\"?", re.MULTILINE)
+REVERT_RE = re.compile(r"This reverts commit (?P<commit>[0-9a-f]{40})")
+PULL_REQUEST_RE = re.compile(
+    r"Pull request: https?:\/\/github\.com\/(?P<owner>[^\/]+)\/(?P<repo>[^\/]+)\/pull\/(?P<number>\d+)",
+    re.MULTILINE,
+)
+PUSH_PATH_RE = re.compile(
+    r"https?:\/\/github\.com\/(?P<owner>[^\/]+)\/(?P<repo>[^\/]+)"
+)
```
shtrom: We have `GitHub.GITHUB_URL_RE`, could we use that?

https://github.com/mozilla-conduit/lando/blob/main/src/lando/utils/github.py#L41-L43

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -25,13 +26,177 @@
...
 from lando.utils.landing_checks import LandingChecks
 from lando.utils.tasks import phab_trigger_repo_update
 
 logger = logging.getLogger(__name__)
 
 
+REVERT_SUMMARY_RE = re.compile(r"^Revert \"?(?P<summary>.*)\"?", re.MULTILINE)
+REVERT_RE = re.compile(r"This reverts commit (?P<commit>[0-9a-f]{40})")
+PULL_REQUEST_RE = re.compile(
+    r"Pull request: https?:\/\/github\.com\/(?P<owner>[^\/]+)\/(?P<repo>[^\/]+)\/pull\/(?P<number>\d+)",
```
shtrom: This could build on the `PUSH_PATH_RE`/`GitHub.GITHUB_URL_RE`

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -25,13 +26,177 @@
...
+
+
+def find_revert_commits(commit_data: list[CommitData]) -> list[str]:
+    """Return the full commit messages of any commits that are reverts."""
+    revert_commits = []
+    for data in commit_data:
+        commit_message = data.desc
+        if REVERT_SUMMARY_RE.search(commit_message):
+            revert_commits.append(commit_message)
+    return revert_commits
```
shtrom: I think this, and supporting REs, should go into the `main.scm` namespace, either `CommitData` or `AbstractSCM`.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -25,13 +26,177 @@
...
+
+
+def parse_pr_url(commit_message: str) -> dict | None:
+    """Return the owner/repo/number from a commit's PR trailer, or `None`."""
+    pr_match = PULL_REQUEST_RE.search(commit_message)
+    return pr_match.groupdict() if pr_match else None
```
shtrom: Ideally, we should have a parser for this in the `utils.GitHub` object, rather than here, so it's in the best place for discoverability and reuse, maybe a static method to the `PullRequest` class?

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -25,13 +26,177 @@
...
+
+
+def parse_push_path(push_path: str) -> tuple[str, str]:
+    """Return the `(owner, repo)` named in a repo push path."""
+    match = PUSH_PATH_RE.search(push_path)
+    return match.group("owner"), match.group("repo")
```
shtrom: We have `GitHub.parse_url` https://github.com/mozilla-conduit/lando/blob/main/src/lando/utils/github.py#L67-L73

We could even add a build method that takes that URL, and returns a constructed `PullRequest`

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -25,13 +26,177 @@
...
+
+
+def find_reverted_commit_hashes(revert_commit_message: str) -> list[str]:
+    """Return the full SHAs named in `This reverts commit <sha>.` lines."""
+    return REVERT_RE.findall(revert_commit_message)
```
shtrom: Probably better in the `main.scm` namespace, too.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,6 +341,19 @@ def run_job(self, job: AutomationJob) -> bool:
...
+        if revert_commits:
+            reverted_pr_numbers = []
+            github_client = GitHubAPIClient(repo.push_path)
+            for commit in revert_commits:
+                pr_numbers = find_reverted_pr_numbers(
+                    commit, repo.push_path, repo.default_branch, scm, github_client
+                )
+                if pr_numbers:
+                    reverted_pr_numbers.extend(pr_numbers)
+            comment_on_reverted_prs(reverted_pr_numbers, github_client)
```
shtrom: As a personal preference, I like the Stepdown Rule https://dzone.com/articles/the-stepdown-rule.

This means high-level entry points appear earlier in the code than utility methods, so you can read the code from top to bottom and get increasing amounts of details. It's particularly good for reviewers!

This is only a preference, and not mandated, but flagging for interest (and hoping you'll like it too!).

shtrom (changes requested):
The logic looks good, but most of the utility functions should go to `main.scm` or `utils.github`, where they would be more easily found, reused, and maintained. This may also help reduce some duplication introduced here.

src/lando/api/legacy/workers/automation_worker.py
shtrom: We may also need to keep the `Pull request: ` trailer in sync with https://github.com/mozilla-conduit/lando/blob/28622091aff45bc65711b55001b20d6163798af8/src/lando/utils/github.py#L674, or at least close to one another so we don't forget to update one when we do the other.

src/lando/main/scm/abstract_scm.py
```diff
@@ -355,3 +359,18 @@ def tag(self, name: str, target: str | None):
...
+        for data in commit_data:
+            commit_message = data.desc
+            if REVERT_SUMMARY_RE.search(commit_message):
+                revert_commits.append(commit_message)
+        return revert_commits
+
+    @staticmethod
+    def find_reverted_commit_hashes(revert_commit_message: str) -> list[str]:
+        """Return the full SHAs named in `This reverts commit <sha>.` lines."""
+        return REVERT_RE.findall(revert_commit_message)
```
shtrom: Ok, seeing this now, which only manipulates CommitData, and doesn't do any SCM interaction, I think this should be staticmethods on the CommitData class instead of the AbstractSCM (FWIW, I was tipped off by the call to `AbstractSCM.find_revert_commits`, which looked weird to me, to call a static method on an abstract class).

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+
         # Trigger update of repo in Phabricator so patches are closed quicker.
         # Especially useful on low-traffic repositories.
         if repo.phab_identifier:
             self.call_task(phab_trigger_repo_update, repo.phab_identifier)
 
         return True
+
+
+def find_reverted_pr_numbers(
```
shtrom: Ooooh stepdown ❤️ Thank you!

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+    push_path: str,
+    default_branch: str,
+    scm: AbstractSCM,
+    github_client: GitHubAPIClient,
+) -> list[str]:
+    """Return PR numbers named in a revert commit that should be commented on."""
+    original_commit_hashes = AbstractSCM.find_reverted_commit_hashes(
+        revert_commit_message
+    )
+    repo_owner, repo_name = parse_push_path(push_path)
```
shtrom: The `github_client` is built from the `repo.push_path` in https://github.com/mozilla-conduit/lando/pull/1325/changes#diff-cb2cabd24a324fa5ca838e0586246f5e526d56eaf6290f0b2346863cb977596bR185,

`GitHubAPIClient` has methods that allow to get the `repo_ower` and `repo_name` without re-parsing the URL https://github.com/mozilla-conduit/lando/blob/2c40908bece39dc5e7a500406fb50b185dce1b40/src/lando/utils/github.py#L189-L195

So I'd say you can get rid of this, and use `github_client.repo_owner` and `github_client.repo_name` directly when calling `reverted_pr_number_for_commit` below.

This also means you don't need the dedicated `parse_push_path` method.

src/lando/main/scm/abstract_scm.py
```diff
@@ -12,6 +13,9 @@
 
 logger = logging.getLogger(__name__)
 
+REVERT_SUMMARY_RE = re.compile(r"^Revert \"?(?P<summary>.*)\"?", re.MULTILINE)
```
shtrom: Do we need `re.MULTILINE` here? The `Revert` line should be the first of the commit.

src/lando/main/scm/abstract_scm.py
```diff
@@ -12,6 +13,9 @@
...
+REVERT_RE = re.compile(r"This reverts commit (?P<commit>[0-9a-f]{40})")
```
shtrom: I think those two strings are very Git-specific. OTOH, we can probably live with them not being in the GitSCM for now, as it would be unnecessarily messy to move there.

Maybe we should just name them with a `GIT_` prefix to acknowledge this.

src/lando/utils/github.py
```diff
@@ -29,6 +29,7 @@
 
 logger = logging.getLogger(__name__)
 
+PR_TRAILER_PREFIX = "Pull request: "
```
shtrom: This could be `PullRequest.TRAILER_PREFIX`

src/lando/main/scm/abstract_scm.py
shtrom: The only thing that would maybe justify having that as class methods in the SCM, would be to handle SCM-specific `REVERT_SUMMARY_RE` and `REVERT_RE`.

HOWEVER, because we may know commits sometime migrate across SCMs, I think we need to be looser in in finding reverts, and we may find reverts that come from another SCM, so it's better to have the logic in an SCM-agnostic class. This confirms (to me) the thought of putting it in `CommitData` instead.

src/lando/main/scm/abstract_scm.py
```diff
@@ -355,3 +359,18 @@ def tag(self, name: str, target: str | None):
 
         If `target` is `None`, use the currently checked out commit.
         """
+
+    @staticmethod
+    def find_revert_commits(commit_data: list[CommitData]) -> list[str]:
+        """Return the full commit messages of any commits that are reverts."""
```
shtrom: Maybe instead of returning a `list[str]`, you could filter the `commit_data`, and return a `list[CommitData]` of only the revert commits.

Downstream users can still use `commit.desc` to get the `commit_message`, BUT you get the rest of the metadata for free.

We could even add a `is_revert_commit` method to `CommitData` to do the filtering on, as well as `reverted_commits` to power `find_reverted_commit_hashes`.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+
+    reverted_pr_numbers = []
```
shtrom: You could make this a `set` upfront, so you don't need to do it at the end to force the uniqueness.

```suggestion
    reverted_pr_numbers = set()
```

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+    expected_owner: str,
+    expected_repo: str,
+    expected_branch: str,
+    github_client: GitHubAPIClient,
+) -> str | None:
+    """Return the PR number to comment on for one reverted commit, or `None` to skip it."""
+
+    pr_url_data = PullRequest.parse_pr_url(original_commit_message)
+    if not pr_url_data:
+        logger.warning(
```
shtrom: `warning` may be a bit too high, as it could be a common occurrence when dealing with non-GitHub contributions such as Phabricator. I'd suggest INFO at most, or even DEBUG.

There used to be an article at https://labs.ig.com/logging-level-wrong-abstraction that gave a good heuristic about choosing which log level to use. The key was this:
> * DEBUG: information that is useful during development. Usually very chatty, and will not show in production.
> * INFO: information you will need to debug production issues.
> * WARN: someone in the team will have to investigate what happened, but it can wait until tomorrow.
> * ERROR: Oh-oh, call the fireman! This needs to be investigated now!

Note that we DO have DEBUG logging enabled in production, so we can filter it in or out (I generally filter it out, unless I'm tracing an issue).

In any case, I don't think this message is such that “someone in the team will have to investigate what happened” (:

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+            f"parseable PR URL in commit message."
+        )
+        return None
+
+    pr_number = pr_url_data["number"]
+
+    try:
+        pull_request = github_client.build_pull_request(pr_number)
+    except Exception:
+        logger.warning(
```
shtrom: This one, I think, is a legitimate warning.

As you are catching `Exception` widely, though, it would be worth using `logger.exception`, which reports the exception to Sentry, so we can debug further.

It may also be worth adding details to the message about what the original exception was.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+) -> str | None:
+    """Return a validation error for a reverted pull request, or `None` if valid."""
+    pr_owner = pr_url_data["owner"]
+    pr_repo = pr_url_data["repo"]
+    pr_number = pr_url_data["number"]
+
+    # The trailer URL points at a different repo than the worker is acting on.
+    if pr_owner != expected_owner or pr_repo != expected_repo:
+        return (
+            f"PR URL in commit message points to unexpected repo: "
```
shtrom: For all this method: it would be worth saying which commit (hash, message) is causing the error.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+            f"{expected_owner}/{expected_repo}."
+        )
+
+    # The PR is on a different branch than the worker is acting on.
+    if pull_request.head_ref != expected_branch:
+        return (
+            f"PR URL in commit message points to PR #{pr_number} which is on "
+            f"branch {pull_request.head_ref}, but automation worker expected PRs "
+            f"to be on branch {expected_branch}."
+        )
```
shtrom: I'm not sure we want this. I suspect a commit me have landed on `autoland`, but may get reverted from `main` later on.

@zzzeid @cgsheeh Worth confirming.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+    # The PR exists but its commit message does not appear in the reverted commit.
+    pr_commit_title = f"{pull_request.title}"
+    if pr_commit_title not in original_commit_message:
+        return (
+            f"PR URL in commit message points to PR #{pr_number} but the commit "
+            f"title of that PR does not appear in the revert commit message."
+        )
```
shtrom: This seems too specific. I know we now allow editing the commit message before landing, so they should be the same, but I'd say relying on what's in the SCM as the source of truth should be sufficient: if a commit that gets reverted says it's coming from a given PR, then I'd ping that PR without this check.

@zzzeid @cgsheeh wdyt?

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+        pr_url_data,
+        pull_request,
+        original_commit_message,
+        expected_owner,
+        expected_repo,
+        expected_branch,
+    )
+
+    if error:
+        logger.warning(
```
shtrom: Maybe `info`?

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,154 @@ def run_job(self, job: AutomationJob) -> bool:
...
+    return None
+
+
+def comment_on_reverted_prs(
+    reverted_pr_numbers: list[str], github_client: GitHubAPIClient
+):
+    """Post a 'has been reverted' comment on each reverted pull request."""
+    for pr_number in reverted_pr_numbers:
+        github_client.add_comment_to_pull_request(
+            pr_number, f"This pull request (#{pr_number}) has been reverted."
```
shtrom: Can we say by which commit?

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1978,518 @@ def test_automation_job_processing(automation_job):
...
+
+    test_file = Path(seed_dir) / "test.txt"
+    test_file.write_text(test_file.read_text() + "added line\n")
+
+    subprocess.run(["git", "add", "test.txt"], check=True, cwd=seed_dir)
+    subprocess.run(
+        ["git", "commit", "-m", commit_message_1],
+        check=True,
+        cwd=seed_dir,
+    )
```
shtrom: It may be worth reusing/extending the `create_git_commit` (or `create_scm_commit`) fixture to shorten the length of this test 

https://github.com/mozilla-conduit/lando/blob/2c40908bece39dc5e7a500406fb50b185dce1b40/src/lando/conftest.py#L1140-L1156

The fixtures would need optionally configurable file name, content to append and commit message, but I think this would be useful in the long run for other tests, too!

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1978,518 @@ def test_automation_job_processing(automation_job):
...
+
+    assert response.status_code == 202, f"Unexpected status code: {response.content}"
+
+    job_id = response.json()["job_id"]
+    job = AutomationJob.objects.get(id=job_id)
+    automation_worker.worker_instance.applicable_repos.add(repo)
+    job.target_repo.scm.push = mock.MagicMock()
+
+    assert automation_worker.run_job(job)
```
shtrom: It may be a good idea to show the error, if the job fails:

```suggestion
    assert automation_worker.run_job(job), job.error
```

(I haven't checked the syntax here, but we have it in other tests throughout the codebase, and it's handy to debug occasional test failures).

src/lando/conftest.py
```diff
@@ -960,6 +960,20 @@ def factory(
     return factory
 
 
+@pytest.fixture
+def git_repo_automation(
+    git_repo: pathlib.Path,
+    tmp_path: pathlib.Path,
+) -> Repo:
```
shtrom, line 966: Could you add a docstring here explaining why this fixture is useful (namely, because the `push_path` is different than the `url`)?

Maybe it would be better named to suggest that it has a push GitHub URL in some way, rather than based on where it's used at the moment.

shtrom (changes requested):
Mainly comments on good practices and making the code reusable and debuggable in prod.

This is looking good, though (:

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,140 @@ def run_job(self, job: AutomationJob) -> bool:
...
+
+    # The PR is on a different branch than the worker is acting on.
+    if pull_request.head_ref != expected_branch:
+        return (
+            f"PR URL in commit {original_commit_hash} message {original_commit_message} "
+            f"points to PR #{pr_number} which is on branch {pull_request.head_ref}, but "
+            f"automation worker expected PRs to be on branch {expected_branch}."
+        )
+    # The PR exists but its commit message does not appear in the reverted commit.
+    pr_commit_title = f"{pull_request.title}"
```
zzzeid: May be unnecessary f-string here?

src/lando/api/legacy/workers/automation_worker.py
zzzeid: Couple of things here. On line 210:

```python
original_commit_message = scm.describe_commit(original_commit_hash).desc
```

`scm` is operating within the current repo on disk, which for Lando purposes is a particular branch of a repo (`default_branch`). So a commit outside this branch may not exist on this branch.

> but may get reverted from main later on. 

I don't think backouts happen on main, generally speaking, but they could happen on different branches.

I think more importantly we should be checking the repo-branch combo, since we do not want cross-chatter (which we had to deal with in a [hacky](https://github.com/mozilla-conduit/phabricator/blob/master/hgext/differentiator.py#L127) way in Phabricator albeit with a different context). We definitely don't want a commit message to trigger a comment in an unrelated repo (e.g., if a pull request URL is referenced in a commit message in a different context, or a malicious actor).

TL;DR there should be a check that ensures the repo referenced in the pull request URL is the same repo where the commit is found.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,140 @@ def run_job(self, job: AutomationJob) -> bool:
...
+def get_pr_errors(
+    pr_url_data: dict,
+    pull_request: PullRequest,
+    original_commit_hash: str,
+    original_commit_message: str,
+    expected_owner: str,
+    expected_repo: str,
+    expected_branch: str,
+) -> str | None:
+    """Return a validation error for a reverted pull request, or `None` if valid."""
```
zzzeid: Would be good to add some descriptions to each argument in the docstring.

src/lando/api/legacy/workers/automation_worker.py
zzzeid: @Magnolia-Liu what does this check intend to prevent? I think the most important thing to check is that the referenced PR is indeed the actual PR that was reverted (and not just any PR referenced in a commit message). The commit message itself may not match anyway because we don't update it with `r=`, etc...

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,140 @@ def run_job(self, job: AutomationJob) -> bool:
...
+
+def reverted_pr_number_for_commit(
+    original_commit_message: str,
+    original_commit_hash: str,
+    expected_branch: str,
+    github_client: GitHubAPIClient,
+) -> str | None:
+    """Return the PR number to comment on for one reverted commit, or `None` to skip it."""
+
+    pr_url_data = PullRequest.parse_pr_url(original_commit_message)
```
zzzeid, line 224: Will this capture any reference to `Pull request: <pull request URL>`? We should make sure it is only capturing the actual PR (Which is at the bottom of the commit message) and not any reference to a PR that user-controlled text might include.

src/lando/utils/github.py
```diff
@@ -441,6 +441,12 @@ def pr_cache_key(self: "PullRequest", *args, **kwargs) -> str:
 class PullRequest:
     """A class that parses data returned from the GitHub API for pull requests."""
 
+    PR_TRAILER_PREFIX = "Pull request: "
+    PULL_REQUEST_RE = re.compile(
+        rf"{PR_TRAILER_PREFIX}{GitHub.GITHUB_URL_RE.pattern}pull/(?P<number>\d+)",
+        re.MULTILINE,
+    )
```
zzzeid, line 447: Mentioned this elsewhere, but can we make this more robust, so that it captures only one instance and only if it's the last line of a commit message?

zzzeid (changes requested):

src/lando/api/legacy/workers/automation_worker.py
Magnolia-Liu: Now that I've thought about it more, I think this came from my meeting notes. I had a note that simply said "check the commit message," without any context about what specifically needed to be validated. When I wrote the test, I added that check without really grounding it in a concrete scenario.

I also think I wrote this before I had a solid understanding of how we extract the PR URL trailer. At the time, I assumed it made sense to guard against a hand-edited/incorrect PR URL, but looking back, that assumption wasn't really based on an actual use case. 

I'll go ahead and remove the check.

src/lando/api/legacy/workers/automation_worker.py
shtrom: > TL;DR there should be a check that ensures the repo referenced in the pull request URL is the same repo where the commit is found.

`repo`, or `repo-branch`? My instinct says the former?

src/lando/api/legacy/workers/automation_worker.py
shtrom: +1

Also suggested name clarifications:
* get_pr_errors -> check_pr_error
* pull_request -> pr_to_revert

(Though, I think we should get rid of this method altogether, see my other comments.)

src/lando/api/legacy/workers/automation_worker.py
shtrom: And the check for matching repo is already here, at the beginning of this method.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,140 @@ def run_job(self, job: AutomationJob) -> bool:
...
+    try:
+        pull_request = github_client.build_pull_request(pr_number)
+    except Exception as e:
+        logger.exception(
+            f"Skipping commit {original_commit_hash}: PR #{pr_number} could not "
+            f"be found via the GitHub API. \n{e}"
+        )
+        return None
+
+    error = get_pr_errors(
```
shtrom: While I appreciate the idea of doing all the error checking in one place, I think, in this particular instance, this makes the logic less clear.

There are many parameters, and some of them seem to be redundant (e.g., `pr_url_data` and `pull_request`). However they are not (as the `pull_request` is built on the `github_client`, which has an owner and repo set, but which are not (yet) validated to be the same in the `pr_url_data`.

Instead, I'd suggest to split out the `get_pr_errors` altogether, and check new data as soon as there is enough:
1. parse PR URL
2. check that owner and repo match
3. build PullRequest object
4. check that head ref and commit message match*

* wrt. my previous comment, neither of those checks may be necessary, so maybe the only needed check is the one from 2, which would simplify the logic altogether by removing step 4`

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,140 @@ def run_job(self, job: AutomationJob) -> bool:
...
+        github_client.repo_name,
+        expected_branch,
+    )
+    if error:
+        logger.info(
+            f"Skipping commit {original_commit_hash} because validation failed: {error}"
+        )
+        return None
+
+    return pr_number
```
shtrom: Why not return the `PullRequest` object here?

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,140 @@ def run_job(self, job: AutomationJob) -> bool:
...
+
+    return None
+
+
+def comment_on_reverted_prs(
+    reverted_pr_numbers: list[str], github_client: GitHubAPIClient, commit_hash: str
+):
+    """Post a 'has been reverted' comment on each reverted pull request."""
+    for pr_number in reverted_pr_numbers:
+        github_client.add_comment_to_pull_request(
```
shtrom: I know this is pre-existing code, but it feel like `add_comment_to_pull_request` is on the wrong object: it should be `PullRequest.add_comment`.

It might be worth adding a stub for it now, even if it uses `github_client.add_comment_to_pull_request` under the hood. This would allow to trade `reverted_pr_numbers: list[str]` for a richer `reverted_prs: list[PullRequest]`, and not require the `github_client` as a separate argument.

src/lando/main/scm/commit.py
```diff
@@ -12,3 +18,16 @@ class CommitData:
...
+        """Return any commits that are reverts."""
+        return [commit for commit in commit_data if commit.is_revert_commit()]
+
+    def is_revert_commit(self) -> bool:
+        """Return whether this commit's message marks it as a revert."""
+        return bool(GIT_REVERT_SUMMARY_RE.match(self.desc))
+
+    def reverted_commit_hashes(self) -> list[str]:
+        """Return the full SHAs this commit reverts."""
+        return GIT_REVERT_RE.findall(self.desc)
```
shtrom, line 33: Lovely (:

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,87 @@ def run_job(self, job: AutomationJob) -> bool:
...
+    pr_repo = pr_url_data["repo"]
+    expected_owner = github_client.repo_owner
+    expected_repo = github_client.repo_name
+
+    if pr_owner != expected_owner or pr_repo != expected_repo:
+        return (
+            f"Skipping commit {original_commit_hash} because PR URL in commit message "
+            f"[{original_commit_message}] points to unexpected repo: {pr_owner}/{pr_repo}, "
+            f"but automation worker expected PRs to be from {expected_owner}/{expected_repo}."
+        )
```
shtrom: We can't return a string here.


```suggestion
        logger.warning (
            f"Skipping commit {original_commit_hash} because PR URL in commit message "
            f"[{original_commit_message}] points to unexpected repo: {pr_owner}/{pr_repo}, "
            f"but automation worker expected PRs to be from {expected_owner}/{expected_repo}."
        )
        return None
```

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,87 @@ def run_job(self, job: AutomationJob) -> bool:
...
+        return None
+    return pr_to_revert
+
+
+def comment_on_reverted_prs(reverted_prs: list[PullRequest], commit_hash: str):
+    """Post a 'has been reverted' comment on each reverted pull request."""
+    for pr in reverted_prs:
+        pr.add_comment(
+            pr.number,
+            f"This pull request (#{pr.number}) has been reverted by commit {commit_hash}.",
```
shtrom: `pr.number` is probably not needed here, as we comment an that PR.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,87 @@ def run_job(self, job: AutomationJob) -> bool:
...
+    if pr_owner != expected_owner or pr_repo != expected_repo:
+        return (
+            f"Skipping commit {original_commit_hash} because PR URL in commit message "
+            f"[{original_commit_message}] points to unexpected repo: {pr_owner}/{pr_repo}, "
+            f"but automation worker expected PRs to be from {expected_owner}/{expected_repo}."
+        )
+
+    try:
+        pr_to_revert = github_client.build_pull_request(pr_number)
+    except Exception as e:
```
shtrom: This may be a bit broad. It's good as a catch-all, but perhaps we want to be more selective about errors, e.g., if we get a 500 from GitHub, that should defer the job to retry it (so we get a good chance to mark all the reverts correctly)

404s, we should just eat up and continue like done here.

Other 4xx from GitHub, not sure. We don't want to make the job perma-deferred, but we should definitely log something (likely a warning).

Catching all Exceptions as a fallback here is fine, as we don't want unintended exceptions to block the main task of the automation job. Perhaps the same apply to any GitHub interaction, in which case you can ignore this whole comment 😅

src/lando/utils/github.py
```diff
@@ -708,6 +713,15 @@ def serialize(self) -> dict[str, str]:
             "user_login": self.user_login,
         }
 
+    @staticmethod
+    def parse_pr_url(commit_message: str) -> dict | None:
+        """Return the owner/repo/number from a commit's PR trailer, or `None`."""
+        pr_match = PullRequest.PULL_REQUEST_RE.search(commit_message)
+        return pr_match.groupdict() if pr_match else None
```
shtrom: To Zeid's point: what happens if their are multiple PR URLs trailers in the commit?

It may be safer to do a `pr_urls=re.findall` first, then do a `re.match` on `pr_urls[-1]`, to parse the last one as authoritative.

src/lando/utils/github.py
```diff
@@ -708,6 +713,15 @@ def serialize(self) -> dict[str, str]:
...
+
+    def add_comment(self, pull_number: int, comment: str) -> dict:
```
shtrom: Here, you can now use `self.number` here, rather than passing `pull_number` separately.

shtrom (approved):
A few nits, but I think that's the last stretch!

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,87 @@ def run_job(self, job: AutomationJob) -> bool:
...
+            reverted_prs.add(pr)
+    return list(reverted_prs)
+
+
+def get_reverted_pr(
+    original_commit_message: str,
+    original_commit_hash: str,
+    github_client: GitHubAPIClient,
+) -> PullRequest | None:
+    """Return the PR number to comment on for one reverted commit, or `None` to skip it."""
```
zzzeid: nit
```suggestion
    """Return the PR to comment on for one reverted commit, or `None` to skip it."""
```

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,87 @@ def run_job(self, job: AutomationJob) -> bool:
...
         job.transition_status(JobAction.LAND, commit_id=commit_id)
 
+        # If any of the new commits are reverts, comment on the reverted PRs.
+        revert_commits = CommitData.find_revert_commits(new_commits)
+        if revert_commits:
+            github_client = GitHubAPIClient(repo.push_path)
+            for commit in revert_commits:
+                pull_requests = find_reverted_prs(commit, scm, github_client)
+                if pull_requests:
+                    comment_on_reverted_prs(pull_requests, commit.hash)
```
zzzeid: nit: would be nice to flatten this a bit. We could have a mapping of a commit to PRs to accomplish this.

There would be no need to check `if pull_requests` since `comment_on_reverted_prs` accepts an empty list.

```suggestion
            github_client = GitHubAPIClient(repo.push_path)
            reverts = {commit.hash: find_reverted_prs(commit, scm, github_client) for commit in revert_commits}
            for commit_hash, pull_requests in reverts:
                comment_on_reverted_prs(pull_requests, commit_hash)
```

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,87 @@ def run_job(self, job: AutomationJob) -> bool:
...
+        original_commit_message = scm.describe_commit(original_commit_hash).desc
+
+        pr = get_reverted_pr(
+            original_commit_message,
+            original_commit_hash,
+            github_client,
+        )
+        if pr:
+            reverted_prs.add(pr)
+    return list(reverted_prs)
```
zzzeid: nit: this could be flattened/simplified a little, also with the help of a mapping.
```suggestion
    original_commits = {
        commit_hash: scm.describe_commit(commit_hash).desc
        for commit_hash in revert_commit.reverted_commit_hashes()
    }
    reverted_prs = [
        get_reverted_pr(commit_hash, commit_message, github_client)
        for commit_hash, commit_message in original_commits.items()
    ]
    return [pr for pr in reverted_prs if pr]
```

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
     assert job_from_db.duration_seconds > 0, (
         "`processing` should set and save the job duration."
     )
+
+
+@pytest.mark.parametrize(
+    "commit_message,expected",
```
zzzeid, line 1921: I think this could use more test cases. For example, having multiple references to PRs in the commit message and having a full commit message. E.g.,

```
Bug 1234567 - do something different in some module

This PR fixes a problem that was introduced in
Pull request https://github.com/mozilla-conduit/test-repo/pull/12345
where something something.

Pull request: https://github.com/mozilla-conduit/test-repo/pull/98765
```

```
Bug 1234567 - do something

Something something

Pull request: https://github.com/mozilla-conduit/test-repo/pull/11111
```

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+
+def make_pull_request(
+    head_ref: str = "main", title: str = "Bug 1234 - add a line"
+) -> mock.MagicMock:
+    """Build a stand-in pull request exposing the attributes `get_pr_errors` reads."""
+    pull_request = mock.MagicMock()
+    pull_request.head_ref = head_ref
+    pull_request.title = title
+    return pull_request
+
```
zzzeid: Is this used anywhere?

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+    client,
+    git_repo_github_push_path,
+    treestatusdouble,
+    automation_worker,
+    mock_phab_trigger_repo_update_apply_async,
+    automation_job,
+    headless_user,
+    create_git_commit,
+):
+
```
zzzeid, line 2035: Would be good to add a docstring here to elaborate on what this test is doing. Can also be added in a comment.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+    job.target_repo.scm.push = mock.MagicMock()
+
+    assert automation_worker.run_job(job), job.error
+
+    job.refresh_from_db()
+    assert job.status == JobStatus.LANDED, f"Job failed with error: {job.error}"
+
+    mock_pr.add_comment.assert_called_once()
+
+    print(mock_pr.add_comment.call_args_list)
```
zzzeid: I assume this is left behind from testing? Should probably change this to an assertion with exact args.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+
+
+@mock.patch("lando.api.legacy.workers.automation_worker.GitHubAPIClient")
+@pytest.mark.django_db
+def test_automation_job_pipeline_2_commits_reverted(
```
zzzeid: Rather than numbering these tests, it would be good to (1) add more specific names in the test name and (2) add a docstring or comments to describe the test.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+
+    seed_dir = repo.pull_path
+
+    mock_github_api_client = mock.MagicMock()
+    pr_1 = mock.MagicMock()
+    pr_1.head_ref = "main"
+    pr_1.title = "Bug 1234 - add a line"
+    pr_1.body = "test description"
+    pr_1.number = 1
+
```
zzzeid: This may be a good candidate for a simple fixture here and above.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+    assert automation_worker.run_job(job), job.error
+
+    job.refresh_from_db()
+    assert job.status == JobStatus.LANDED, f"Job failed with error: {job.error}"
+
+    pr_1.add_comment.assert_called_once()
+    pr_2.add_comment.assert_called_once()
+
+    print(pr_1.add_comment.call_args_list)
+    print(pr_2.add_comment.call_args_list)
```
zzzeid: Prints here should be converted to assertions.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+    job.target_repo.scm.push = mock.MagicMock()
+
+    assert automation_worker.run_job(job), job.error
+
+    job.refresh_from_db()
+    assert job.status == JobStatus.LANDED, f"Job failed with error: {job.error}"
+
+    mock_pr.add_comment.assert_called_once()
+
+    print(mock_pr.add_comment.call_args_list)
```
zzzeid: Print here.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1973,3 +1976,426 @@ def test_automation_job_processing(automation_job):
...
+
+    commit_message_2 = "\n\nBug 5678 - add another line\n\nPull request: https://github.com/mozilla-conduit/test-repo/pull/2\n\n\n"
+    create_git_commit(
+        Path(seed_dir),
+        message=commit_message_2,
+        name="test.txt",
+        content="added line\n added another line\n",
+    )
+
+    subprocess.run(
```
zzzeid: I think this should go into either its own fixture or a small method (say `generate_revert_patch`) both to improve readability and to be reused (perhaps with different parameters).

zzzeid (changes requested):
Looking pretty good! Just a few small comments.

src/lando/api/legacy/workers/automation_worker.py
```diff
@@ -175,9 +178,82 @@ def run_job(self, job: AutomationJob) -> bool:
...
+        return None
+
+    try:
+        pr_to_revert = github_client.build_pull_request(pr_number)
+    except Exception as e:
+        logger.exception(
+            f"Skipping commit {original_commit_hash}: PR #{pr_number} could not "
+            f"be found via the GitHub API. \n{e}"
+        )
+        return None
```
shtrom: FWIW, I don't think we need to include the text of the exception with `logger.exception`. It does it by default https://docs.python.org/3/library/logging.html#logging.Logger.exception


```suggestion
    except Exception as e:
        logger.exception(
            f"Skipping commit {original_commit_hash}: PR #{pr_number} could not "
            f"be found via the GitHub API."
        )
        return None
```

src/lando/api/legacy/workers/automation_worker.py
shtrom:
```suggestion
            f"This pull request has been reverted by commit {commit_hash}.",
```

src/lando/main/scm/commit.py
```diff
@@ -1,6 +1,12 @@
+from __future__ import annotations
```
shtrom, line 1: Python 3.14 added lazy annotations evaluation https://docs.python.org/3/reference/compound_stmts.html#annotations

I think we'd be better off upgrading Lando to Python 3.14.

@zzzeid @cgsheeh WDYT? 

...

It's simple enough I'll just go ahead and do it. If we merge that ahead of this PR, this will no longer be needed. https://bugzilla.mozilla.org/show_bug.cgi?id=2062839

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1940,3 +1944,455 @@ def test_automation_job_processing(automation_job):
     assert job_from_db.duration_seconds > 0, (
         "`processing` should set and save the job duration."
     )
+
+
+MULTITRAILER_COMMIT_MESSAGE = """Bug 1234567 - do something
+
+This PR fixes a problem that was introduced in
+Pull request https://github.com/mozilla-conduit/test-repo/pull/12345
```
shtrom: I'd leave a colon here, just to be extra tricky to the RE.
```suggestion
Pull request: https://github.com/mozilla-conduit/test-repo/pull/12345
```

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1940,3 +1944,455 @@ def test_automation_job_processing(automation_job):
...
+        ('Revert "Bug 1 - a thing"', True),
+        ("Bug 1 - a normal commit", False),
+    ],
+    ids=[
+        "revert",
+        "no_revert",
+    ],
+)
+def test_is_revert_commit(description, is_revert, make_scm_commit):
+    """`find_revert_commits` returns the full message of each commit that is a revert."""
```
shtrom: The docstring doesn't seem to match the test: there is no assertion that checks the commit message, just a boolean check.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1940,3 +1944,455 @@ def test_automation_job_processing(automation_job):
...
+    commit = make_scm_commit(1)
+    commit.desc = description
+    assert commit.is_revert_commit() == is_revert, (
+        "`is_revert_commit` should match expected value"
+    )
+
+
+@pytest.fixture
+def mock_pull_request() -> Callable:
```
shtrom: nit: we might as well move this into a conftest now (maybe as `mock_github_pull_request`) where other GitHub fixtures are. This will be useful in many other places.

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1940,3 +1944,455 @@ def test_automation_job_processing(automation_job):
...
+        pull_request.number = number
+        pull_request.title = title
+        pull_request.body = body
+        pull_request.head_ref = head_ref
+        return pull_request
+
+    return mock_pull_request
+
+
+def assert_reverted_pr_comment(pull_request: mock.MagicMock, pr_number: int) -> str:
```
shtrom: nit: I'd recommend putting non-test utility function towards the bottom of the file (or at least to stepdown them).

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1940,3 +1944,455 @@ def test_automation_job_processing(automation_job):
...
+    mock_pr = mock_pull_request(
+        number=1, title="Bug 1234 - add a line", body="test description"
+    )
+    mock_github_api_client.build_pull_request.return_value = mock_pr
+    mock_github_api_client.repo_owner = "mozilla-conduit"
+    mock_github_api_client.repo_name = "test-repo"
+
+    github_api_client.return_value = mock_github_api_client
+
+    commit_message = "\n\nBug 1234 - add a line\n\ntest description\n\nPull request: https://github.com/mozilla-conduit/test-repo/pull/1\n\n\n"
```
shtrom: Why are there two leading `\n` here?

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -1940,3 +1944,455 @@ def test_automation_job_processing(automation_job):
...
+        number=1, title="Bug 1234 - add a line", body="test description"
+    )
+    pr_2 = mock_pull_request(number=2, title="Bug 5678 - add another line")
+
+    prs_by_number = {"1": pr_1, "2": pr_2}
+    mock_github_api_client.build_pull_request.side_effect = lambda pr_number: (
+        prs_by_number[pr_number]
+    )
+
+    commit_message_1 = "\n\nBug 1234 - add a line\n\ntest description\nPull request: https://github.com/mozilla-conduit/test-repo/pull/1\n\n\n"
```
shtrom: It may be worth stripping those strings on both sides altogether?

src/lando/main/scm/commit.py
zzzeid: seams good

Magnolia-Liu:
@shtrom @zzzeid Regarding [this line](https://github.com/mozilla-conduit/lando/pull/1434/changes#diff-87ae2cf8c225691dd0aa45b9a3b54d7ff0e0dcee5b846f8d6fd11433287ce156R66) in another PR: Right now github revert commit messages get captured by this regex but then don't match any of the subsequent expressions. Do you guys think I should try to bypass this specific check or would it make more sense to modify the rest of the regular expressions to work with github reverts too?

src/lando/main/scm/commit.py
Magnolia-Liu: Seems like this was merged ahead of this PR, should I remove that line then?

shtrom:
> @shtrom @zzzeid Regarding [this line](https://github.com/mozilla-conduit/lando/pull/1434/changes#diff-87ae2cf8c225691dd0aa45b9a3b54d7ff0e0dcee5b846f8d6fd11433287ce156R66) in another PR: Right now github revert commit messages get captured by this regex but then don't match any of the subsequent expressions. Do you guys think I should try to bypass this specific check or would it make more sense to modify the rest of the regular expressions to work with github reverts too?

Ah, that's a good call. The “Revert” form is the one that is now used when sheriffs back-out commits via lando. I think it is necessary to support git-reverts, too.

zzzeid:
> > @shtrom @zzzeid Regarding [this line](https://github.com/mozilla-conduit/lando/pull/1434/changes#diff-87ae2cf8c225691dd0aa45b9a3b54d7ff0e0dcee5b846f8d6fd11433287ce156R66) in another PR: Right now github revert commit messages get captured by this regex but then don't match any of the subsequent expressions. Do you guys think I should try to bypass this specific check or would it make more sense to modify the rest of the regular expressions to work with github reverts too?
> 
> Ah, that's a good call. The “Revert” form is the one that is now used when sheriffs back-out commits via lando. I think it is necessary to support git-reverts, too.

Also see [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2063219), which is related.

shtrom:
> > > @shtrom @zzzeid Regarding [this line](https://github.com/mozilla-conduit/lando/pull/1434/changes#diff-87ae2cf8c225691dd0aa45b9a3b54d7ff0e0dcee5b846f8d6fd11433287ce156R66) in another PR: Right now github revert commit messages get captured by this regex but then don't match any of the subsequent expressions. Do you guys think I should try to bypass this specific check or would it make more sense to modify the rest of the regular expressions to work with github reverts too?
> > 
> > 
> > Ah, that's a good call. The “Revert” form is the one that is now used when sheriffs back-out commits via lando. I think it is necessary to support git-reverts, too.
> 
> Also see [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2063219), which is related.

I added more tests for this in #1458

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -2101,3 +2105,435 @@ def test_automation_job_processing(automation_job):
...
+    automation_worker.worker_instance.applicable_repos.add(repo)
+    job.target_repo.scm.push = mock.MagicMock()
+
+    assert automation_worker.run_job(job), job.error
+
+    job.refresh_from_db()
+    assert job.status == JobStatus.LANDED, f"Job failed with error: {job.error}"
+
+    # Only the reverted PR is commented on; the two surrounding non-revert commits
+    # are ignored.
```
zzzeid: nit: We are not explicitly testing that the other PRs are not commented on here. To make this more accurate we should assert those methods are called 0 times.

zzzeid (approved):
This is showing some test failures,

PR #1329, opened 2026-07-10
uplift_worker: disable auto try push for secure uplifts (bug 2053811)

Magnolia-Liu:
Automatic try-pushes for for security uplift requests currently give an error sent in a email notification. 

Avoid automatically pushing to try for security uplifts by adding a new is_secure boolean field in `UpliftSubmission` model, setting the status in the uplift post request, detecting this status in the uplift worker, and skipping the try push based on this detection.

---

Lando: [link](https://lando.moz.tools/pulls/lando/1329/)
Bugzilla: [bug 2053811](https://bugzilla.mozilla.org/show_bug.cgi?id=2053811)

:warning: This pull request has 7 warnings.

### DIFF
--- a/src/lando/api/legacy/workers/uplift_worker.py
+++ b/src/lando/api/legacy/workers/uplift_worker.py
@@ -27,6 +27,7 @@ from lando.main.scm import GitSCM
 from lando.main.scm.commit import CommitData
 from lando.main.scm.helpers import PatchHelper
 from lando.try_api.api import get_commit_hash, get_commit_map
+from lando.utils.landing_checks import BugReferencesCheck
 from lando.utils.tasks import (
     send_uplift_failure_email,
     send_uplift_success_email,
@@ -166,12 +167,12 @@ class UpliftWorker(Worker):
         # `LANDED` is the same as "success".
         job.status = JobStatus.LANDED
         job.save()
-
         if ConfigurationVariable.get(ConfigurationKey.UPLIFT_TRY_ENABLED, True):
             try:
                 try_job = self.create_uplift_try_push(
                     base_revision, repo.scm_type, job, scm, new_commits
                 )
+
             except Exception:
                 logger.exception(
                     "Failed to create try push for uplift job.",
@@ -317,6 +318,11 @@ class UpliftWorker(Worker):
         new_commits: Iterable[CommitData],
     ) -> LandingJob:
         """Create a Try `LandingJob` for the commits landed by an uplift job."""
+        patch_helpers = list(scm.get_patch_helpers_for_commits(new_commits))
+        result = self.check_uplift_bug_references(patch_helpers)
+        if result:
+            raise ValueError(result)
+
         try_repo = Repo.objects.get(name="try")
 
         if try_repo.scm_type != repo_scm_type:
@@ -342,13 +348,12 @@ class UpliftWorker(Worker):
                 raise
 
         with transaction.atomic():
-            revisions = []
-            patch_helpers = scm.get_patch_helpers_for_commits(new_commits)
-            for patch_helper in patch_helpers:
-                revisions.append(self.create_revisions_from_patch_helper(patch_helper))
+            revisions = [
+                self.create_revisions_from_patch_helper(patch_helper)
+                for patch_helper in patch_helpers
+            ]
             try_revision = self.create_try_revision(job.requester_email)
             revisions.append(try_revision)
-
             try_job = LandingJob.objects.create(
                 target_repo=try_repo,
                 requester_email=job.requester_email,
@@ -358,6 +363,17 @@ class UpliftWorker(Worker):
             add_revisions_to_job(revisions, try_job)
             try_job.save()
         return try_job
+
+    def check_uplift_bug_references(
+        self, patch_helpers: list[PatchHelper]
+    ) -> str | None:
+        """Check if uplift job contains references to non-public bugs.
+
+        Return the error message when a referenced bug is not public, and `None` otherwise."""
+        secure_check = BugReferencesCheck()
+        for patch_helper in patch_helpers:
+            secure_check.next_diff(patch_helper)
+        return secure_check.result()
 
     def create_try_diff_from_json(self) -> str:
         try_config_path = (
--- a/src/lando/api/tests/test_uplift.py
+++ b/src/lando/api/tests/test_uplift.py
@@ -10,6 +10,7 @@ from packaging.version import (
     Version,
 )
 
+from lando.api.legacy.commit_message import parse_bugs
 from lando.api.legacy.uplift import (
     create_uplift_bug_update_payload,
     parse_milestone_version,
@@ -105,7 +106,6 @@ def test_uplift_creation_uses_existing_revisions_and_links_jobs(
         create_patch_revision(123, patch=normal_patch(0)),
     ]
     revisions_ordered = reversed(revisions_created)
-
     url = reverse("uplift-page")
     form_data = {
         "source_revisions": [revision.revision_id for revision in revisions_ordered],
@@ -723,10 +723,12 @@ def test_link_assessment_replaces_existing_form(
 
 
 @pytest.mark.django_db
+@pytest.mark.parametrize("secure", [True, False])
 def test_uplift_worker_applies_patches_and_creates_uplift_revision_success_git(
     repo_mc,
     user,
     uplift_worker,
+    secure,
     create_patch_revision,
     normal_patch,
     monkeypatch,
@@ -775,6 +777,22 @@ def test_uplift_worker_applies_patches_and_creates_uplift_revision_success_git(
         mock_task,
     )
 
+    mock_bug_search = mock.MagicMock()
+    mock_status_code = mock.MagicMock()
+    if secure:
+        mock_bug_search.return_value = set()
+        mock_status_code.return_value = 401
+
+    else:
+        bug_ids = parse_bugs(revisions[0].commit_message)
+        mock_bug_search.return_value = set(bug_ids)
+        mock_status_code.return_value = 200
+
+    monkeypatch.setattr(
+        "lando.utils.landing_checks.get_status_code_for_bug",
+        mock_status_code,
+    )
+    monkeypatch.setattr("lando.utils.landing_checks.search_bugs", mock_bug_search)
     # Let update_repo/apply_patch run for real and only mock moz-phab uplift to return new tip D-IDs.
     monkeypatch.setattr(
         uplift_worker,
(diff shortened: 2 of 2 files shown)


src/lando/api/tests/test_uplift.py
```diff
@@ -1309,3 +1335,54 @@ def test_uplift_context_for_revision_returns_original_and_uplifted_requests(
...
+    )
+
+    # @pytest.mark.django_db
+    # def test_secure_uplift_skips_try_push(uplift_worker, make_uplift_job_with_revisions, ...):
+
+    #     job = make_uplift_job_with_revisions(repo, user, revisions)
+    #     job.submission.is_secure = True
+    #     job.submission.save()
+    #     # ... run as in the existing success test ...
+    #     assert uplift_worker.run_job(job)
```
zzzeid: Is this left here by mistake?

src/lando/api/tests/test_uplift.py
```diff
@@ -723,9 +729,13 @@ def test_link_assessment_replaces_existing_form(
 
 
 @pytest.mark.django_db
+@pytest.mark.parametrize("secure", [True, False])
 def test_uplift_worker_applies_patches_and_creates_uplift_revision_success_git(
```
zzzeid, line 727: I think the name of this test may be outdated now.

src/lando/api/tests/test_uplift.py
```diff
@@ -1309,3 +1335,54 @@ def test_uplift_context_for_revision_returns_original_and_uplifted_requests(
...
+    # def test_secure_uplift_skips_try_push(uplift_worker, make_uplift_job_with_revisions, ...):
+
+    #     job = make_uplift_job_with_revisions(repo, user, revisions)
+    #     job.submission.is_secure = True
+    #     job.submission.save()
+    #     # ... run as in the existing success test ...
+    #     assert uplift_worker.run_job(job)
+    assert not LandingJob.objects.filter(target_repo__name="try").exists(), (
+        "Secure uplift should not create a public Try push."
+    )
```
zzzeid: I don't think this would ever be `True` before running a job?

zzzeid (changes requested):

cgsheeh (changes requested):
I think we should use `BugReferencesCheck` in the worker here instead of using `revision_is_secure` to check the Phabricator secure-project tag. This gate is really about whether a commit references a private bug, so reusing the check skips try only when the push would actually break/leak, preserves CI coverage for secure-tagged revisions that reference public bugs (common when phab-bot is stalled), and survives the GitHub migration since it only needs commit messages + BMO.

shtrom:
> I think we should use `BugReferencesCheck` in the worker here instead of using `revision_is_secure` to check the Phabricator secure-project tag. This gate is really about whether a commit references a private bug, so reusing the check skips try only when the push would actually break/leak, preserves CI coverage for secure-tagged revisions that reference public bugs (common when phab-bot is stalled), and survives the GitHub migration since it only needs commit messages + BMO.

Perhaps we should apply all of the landing checks enabled for the target repo?

zzzeid:
> > I think we should use `BugReferencesCheck` in the worker here instead of using `revision_is_secure` to check the Phabricator secure-project tag. This gate is really about whether a commit references a private bug, so reusing the check skips try only when the push would actually break/leak, preserves CI coverage for secure-tagged revisions that reference public bugs (common when phab-bot is stalled), and survives the GitHub migration since it only needs commit messages + BMO.
> 
> Perhaps we should apply all of the landing checks enabled for the target repo?

Yeah good point, this would also prevent hard-coding the check name, which would be good in case we ever add any other additional checks that affect security patches.

It would be good to update the docstrings and helptext of the hooks to make it more obvious that those are no longer only running at landing time.

Edit: removed comment (see more detail in [bug 2056365](https://bugzilla.mozilla.org/show_bug.cgi?id=2056365))

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -167,21 +167,20 @@ def apply_uplift_revision(revision: Revision):
         job.status = JobStatus.LANDED
         job.save()
 
-        if ConfigurationVariable.get(ConfigurationKey.UPLIFT_TRY_ENABLED, True):
```
zzzeid, line 170: We can decouple removing the feature flag from this bug (i.e., it can be done in a separate branch). There's no harm in keeping a feature flag in until we're confident with the feature.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -167,21 +167,20 @@ def apply_uplift_revision(revision: Revision):
...
-                )
+        try:
+            try_job = self.create_uplift_try_push(
+                base_revision, repo.scm_type, job, scm, new_commits
+            )
+        except Exception:
+            logger.exception(
+                "Failed to create try push for uplift job.",
+                extra={"job_id": job.id},
+            )
```
zzzeid: I know this is pre-existing code (but it's being touched here). This except statement should be more specific.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -340,24 +339,34 @@ def create_uplift_try_push(
...
-                requester_email=job.requester_email,
-                target_commit_hash=target_commit_hash,
-                status=JobStatus.SUBMITTED,
-            )
-            add_revisions_to_job(revisions, try_job)
-            try_job.save()
-        return try_job
+            patch_helpers = list(scm.get_patch_helpers_for_commits(new_commits))
+
+            if not self.uplift_is_private(patch_helpers):
```
zzzeid: Could check `if self.uplift_is_private(patch_helpers)` and raise early, then dedent the rest.

src/lando/main/models/uplift.py
```diff
@@ -183,6 +183,9 @@ class UpliftSubmission(BaseModel):
         default=UpliftTargetSelectionMethod.SERVER_RENDERED,
     )
 
+    # Whether the uplift request is for a secure revision
+    is_secure = models.BooleanField(default=False)
```
zzzeid: This appears unused at this point.

src/lando/ui/legacy/revisions.py
zzzeid, line 1: Unintended change?

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -340,24 +339,34 @@ def create_uplift_try_push(
...
+                for patch_helper in patch_helpers:
+                    revisions.append(
+                        self.create_revisions_from_patch_helper(patch_helper)
+                    )
```
zzzeid: This could be a list comprehension.

src/lando/main/migrations/0063_upliftsubmission_is_secure.py
zzzeid, line 1: File no longer needed after removing the new field.

zzzeid (changes requested):

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -166,13 +167,13 @@ def apply_uplift_revision(revision: Revision):
...
         job.save()
-
         if ConfigurationVariable.get(ConfigurationKey.UPLIFT_TRY_ENABLED, True):
             try:
                 try_job = self.create_uplift_try_push(
                     base_revision, repo.scm_type, job, scm, new_commits
                 )
-            except Exception:
+
+            except (ValueError, Repo.DoesNotExist):
```
cgsheeh: Narrowing this `except` is usually good practice, but weigh the tradeoff here: the job is already marked `LANDED` and saved above, so this Try push is a best-effort side effect.

Previously *any* failure was swallowed and logged, keeping the successful uplift successful. Now an unexpected error propagates out after the uplift already committed, which could turn a succeeded uplift into an apparent failure/retry. 

We should add a catch-all `except Exception` and log that an unexpected error occurred.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -340,15 +341,18 @@ def create_uplift_try_push(
                     extra={"job_id": job.id},
                 )
                 raise
-
         with transaction.atomic():
-            revisions = []
-            patch_helpers = scm.get_patch_helpers_for_commits(new_commits)
-            for patch_helper in patch_helpers:
-                revisions.append(self.create_revisions_from_patch_helper(patch_helper))
+            patch_helpers = list(scm.get_patch_helpers_for_commits(new_commits))
```
cgsheeh: This check does network I/O (`search_bugs`/`get_status_code_for_bug` hit BMO), but it runs inside `with transaction.atomic():`. This will cause this code to hold a DB transaction and its connection open for the whole duration of a potentially slow HTTP call. The check needs nothing from the transaction and nothing is created before it, so lift it out above the `with` block. 

Rule of thumb: keep transactions short and never wrap blocking network I/O in one. :)

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -340,15 +341,18 @@ def create_uplift_try_push(
...
+
+            if self.uplift_is_private(patch_helpers):
+                raise ValueError("Uplift job references a private bug.")
```
cgsheeh: A truthy result from this check does not only mean "private" — it also fires for a 404, a BMO outage, and unexpected statuses. Hardcoding "references a private bug" means a BMO outage will log the wrong cause and send whoever's debugging down the wrong path. The check already returns a precise message; surface it instead of flattening it.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -357,7 +361,13 @@ def create_uplift_try_push(
             )
             add_revisions_to_job(revisions, try_job)
             try_job.save()
-        return try_job
+            return try_job
+
+    def uplift_is_private(self, patch_helpers: PatchHelper) -> str | None:
```
cgsheeh: Three small things on this signature:

1. A name starting with `is_` reads as a predicate, so I would expect a `bool` to be returned here. We should rename to `check_uplift_bug_references()` and keep the message return (pairs with the discarded-message note above).
2. The `patch_helpers` arg is a `list[PatchHelper]`, not a single `PatchHelper` — fix the type hint.
3. Missing a docstring; every other method here has one, so we should add one per the repo conventions.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -357,7 +361,13 @@ def create_uplift_try_push(
             )
             add_revisions_to_job(revisions, try_job)
             try_job.save()
-        return try_job
+            return try_job
```
cgsheeh: Dedent this — the `with` block commits when it exits, which happens on the way out either way, so returning after the block is equivalent and reads cleaner.

src/lando/api/tests/test_uplift.py
```diff
@@ -775,6 +777,22 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
...
     )
 
+    mock_bug_search = mock.MagicMock()
+    mock_status_code = mock.MagicMock()
+    if secure:
+        mock_bug_search.return_value = set()
+        mock_status_code.return_value = 401
+
+    else:
+        bug_id = parse_bugs(revisions[0].commit_message)
```
cgsheeh: `parse_bugs` returns a list, so the variable name should be a plural `bug_ids`.

cgsheeh (changes requested):

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -341,14 +342,18 @@ def create_uplift_try_push(
                 )
                 raise
 
+        patch_helpers = list(scm.get_patch_helpers_for_commits(new_commits))
```
cgsheeh: I'd move these four lines to the top of `create_uplift_try_push`, above the `Repo.objects.get(name="try")` lookup.

Right now the check runs after the Try repo lookup and the CommitMap conversion. For a secure uplift where the CommitMap lookup happens to fail first, we log `CommitMap not found` and never learn the Try push was going to be skipped anyway. Nothing above the check is needed by it, and running it first also avoids the pointless lookup work for uplifts we're not going to push.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -359,6 +364,15 @@ def create_uplift_try_push(
             try_job.save()
         return try_job
 
+    def check_uplift_bug_references(
+        self, patch_helpers: list[PatchHelper]
+    ) -> str | None:
+        """Check if uplift job contains references to non-public bugs"""
```
cgsheeh: This docstring is missing the trailing period, and it doesn't say what the return value means. Worth stating that it returns the issue message when a referenced bug isn't public and `None` when everything checks out.

cgsheeh (approved):
Looks great! r+ with a few nits. :)

cgsheeh (approved):

cgsheeh:
@Magnolia-Liu could you update this with the latest changes from `main` so we can merge it? :)

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -359,6 +364,16 @@ def create_uplift_try_push(
             try_job.save()
         return try_job
 
+    def check_uplift_bug_references(
+        self, patch_helpers: list[PatchHelper]
+    ) -> str | None:
+        """Check if uplift job contains references to non-public bugs.
+        Return the error message when a referenced bug is not public, and `None` otherwise."""
```
zzzeid, line 372: nit: Missing [blank line](https://peps.python.org/pep-0257/#multi-line-docstrings) after first line, in multi-line docstring.

zzzeid (approved):

zzzeid (approved):

PR #1332, opened 2026-07-14
migrate_data: add migrate_3_bug2054804_enable_PreventSignedCommitsCheck (bug 2054804)

shtrom:
This is only relevant for Git repos that get synced to HgMO, but it doesn't hurt to set it for all repos for the time-being, as it is the default for new repos.

This data migration was missed when bug 1996492 was deployed.

### DIFF
--- a/src/lando/utils/management/commands/migrate_data.py
+++ b/src/lando/utils/management/commands/migrate_data.py
@@ -130,3 +130,26 @@ class Command(BaseCommand):
             CommitMap.objects.bulk_update(batch, ["git_repo_name"])
 
         self.stdout.write("done.")
+
+    def migrate_3_bug2054804_enable_PreventSignedCommitsCheck(
+        self, ask_confirm: bool = True
+    ):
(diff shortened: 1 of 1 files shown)


src/lando/utils/management/commands/migrate_data.py
```diff
@@ -97,3 +97,26 @@ def migrate_1_bug1971103_add_firefox_required_automation_permissions(
                 self.stdout.write(f"{repo.name} ", ending="")
 
         self.stdout.write("done.")
+
+    def migrate_3_bug2054804_enable_PreventSignedCommitsCheck(self, ask_confirm: bool = True)
```
shtrom: 3, because 2 is in https://github.com/mozilla-conduit/lando/pull/1322

zzzeid (approved):
Failing formatting check, otherwise 👍

shtrom:
Rebased, squashed, and force-pushed due to landing conflict https://lando.moz.tools/landings/69640/

PR #1334, opened 2026-07-17
headless_api: remove headless_api.add_automationjob permission gate (bug 2055757)

shtrom:
We now have repo.required_automation_permission, but we add a test case to make sure the system fails closed if it's unset.

 * headless_api: remove headless_api.add_automationjob permission gate
 * headless_api: improve error reporting for missing repo.required_automation_permission

---
Lando: [link](https://lando.moz.tools/pulls/lando/1334/)
Bugzilla: [bug 2055757](https://bugzilla.mozilla.org/show_bug.cgi?id=2055757)

### DIFF
--- a/src/lando/headless_api/tests/test_automation_api.py
+++ b/src/lando/headless_api/tests/test_automation_api.py
@@ -280,18 +280,32 @@ def test_automation_job_create_hg_repo_rejected(
     )
 
 
+@pytest.mark.parametrize(
+    "automation_enabled,empty_automation_permission_required",
+    (
+        (False, False),
+        (False, True),
+        (True, True),
+    ),
+)
 @pytest.mark.django_db
 def test_automation_job_create_repo_automation_disabled(
-    client,
-    headless_user,
-    repo_mc,
+    client: Client,
+    headless_user: tuple[User, str],
+    repo_mc: Callable,
+    automation_enabled: bool,
+    empty_automation_permission_required: bool,
 ):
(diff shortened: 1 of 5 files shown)


src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -323,62 +337,14 @@ def test_automation_job_create_repo_automation_disabled(
     )
     assert (
         response.json()["details"]
-        == "Repo mozilla-central-git is not enabled for automation."
+        == "Repo mozilla-central-git is not enabled for automation, or the required permission is not set."
     ), "Details should indicate automation API is disabled for repo."
```
zzzeid, line 341: This is outdated now.

zzzeid (approved):

src/lando/headless_api/tests/test_automation_api.py
shtrom: The django permission was replaced with an SCM permission as a text field. If that text field is empty, we fail closed, and don't allow access.

I updated the error message because it caused some confusion in the past when a user did have the expected correct permission, but the repo's field was incorrectly empty.

shtrom: Ah, I see what you mean. In a way, not having the required permission set should also count as disabling the repo, even if less voluntarily. I think this is accurate enough. Worse come to worse, we can `grep` for the message (:

PR #1335, opened 2026-07-17
create_environment_repo: add releng staging esr153 repo (bug 2046470)

jcristau:
Remove esr128 (EOL), and set required_automation_permission so that the automation API can actually be used.

---
Lando: [link](https://lando.moz.tools/pulls/lando/1335/)
Bugzilla: [bug 2046470](https://bugzilla.mozilla.org/show_bug.cgi?id=2046470)

### DIFF
--- a/src/lando/utils/management/commands/create_environment_repos.py
+++ b/src/lando/utils/management/commands/create_environment_repos.py
@@ -215,7 +215,7 @@ for branch in ["main", "autoland", "beta", "release", "esr115", "esr140", "esr15
             "push_path": "https://github.com/mozilla-releng/staging-firefox.git",
             "short_name": f"staging-firefox-{branch}",
             "required_permission": SCM_LEVEL_1,
-            "required_automation_permission": SCM_LEVEL_1,
+            "required_automation_permission": SCM_LEVEL_3,
             "automation_enabled": True,
         }
     )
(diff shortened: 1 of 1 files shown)


src/lando/utils/management/commands/create_environment_repos.py
```diff
@@ -215,6 +215,7 @@
             "push_path": "https://github.com/mozilla-releng/staging-firefox.git",
             "short_name": f"staging-firefox-{branch}",
             "required_permission": SCM_LEVEL_1,
+            "required_automation_permission": SCM_LEVEL_1,
```
shtrom: Let's make this SCM_LEVEL_3, as there's no code review required on this path.

```suggestion
            "required_automation_permission": SCM_LEVEL_3,
```

shtrom: Some offline convo pointed out that
1. this staging repo is akin to pushing to Try, which doesn't require any more than SCM1
2. but this repo is more visible/discoverable as on GitHub, even though the code is never used
3. the strong requirement is that the `stage-ffxbld` is allowed to push, but some other staff should be able too
4. we are just about to remove the add_automation_job gate https://bugzilla.mozilla.org/show_bug.cgi?id=2055757 in favour of automation_permission (#1334)
5. access is also gated by having a manually-issued long-lived token, but we are considering moving this to short-lived OAuth-obtainable token

So it feels like, mainly once 4 is addressed, and if we go ahead with 5, this would allow too broad an unconstrained write access to the target GitHub repos. Note that as it's possible to land commits without a review via Phabricator, there is also a similar risk with SCM1 there.

shtrom (approved):

PR #1337, opened 2026-07-17
revisions: block hackbot commits (bug 2054545)

zzzeid:
- add blocker_diff_author_is_hackbot
- add tests
- fix phabdouble to account for user email (and raw data)

---
Lando: [link](https://lando.moz.tools/pulls/lando/1337/)
Bugzilla: [bug 2054545](https://bugzilla.mozilla.org/show_bug.cgi?id=2054545)

### DIFF
--- a/src/lando/api/legacy/revisions.py
+++ b/src/lando/api/legacy/revisions.py
@@ -166,7 +166,7 @@ def blocker_diff_author_is_hackbot(*, diff: dict, **kwargs) -> Optional[str]:
     if hackbot_email not in emails:
         return None
 
-    return "Diff is authored by Hackbot."
+    return "Diff contains commit authored by Hackbot."
 
 
 def revision_has_needs_data_classification_tag(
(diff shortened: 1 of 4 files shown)


zzzeid:
screenshot 
image

src/lando/api/legacy/revisions.py
```diff
@@ -155,6 +155,20 @@ def blocker_diff_author_is_known(*, diff: dict, **kwargs) -> Optional[str]:
...
+    hackbot_email = "hackbot@mozilla.tld"
+    commits = PhabricatorClient.expect(diff, "attachments", "commits", "commits")
+    if not commits:
+        return None
+
+    emails = (c.get("author", {}).get("email", "").lower() for c in commits)
+    if hackbot_email not in emails:
+        return None
+
+    return "Diff is authored by Hackbot."
```
suhaibmujahid: This might indicate that the problem is that it was originally authored by Hackbot, instead it is still the author.
```suggestion
    return "Diff author is still Hackbot."
```

suhaibmujahid (approved):
LGTM!

src/lando/api/legacy/revisions.py
zzzeid: Maybe "Diff contains commit authored by Hackbot" would be the most accurate here. The fix is to resubmit the amended commit with proper authorship information.

Magnolia-Liu (approved):

PR #1338, opened 2026-07-21
try_api: catch and report Exceptions in LandingChecks (Bug 2056454)

shtrom:
* tests: fix pytest.mark.parametrize non-Collection deprecation
 
    See https://docs.pytest.org/en/stable/deprecations.html\#parametrize-iterators.

 * tests: remove test_try from lando.api

    It is skipped since 2024, and we have a new implementation with a
    dedicated testsuite in lando.try_api.

 * try_api: catch and report Exceptions in LandingChecks

---
Lando: [link](https://lando.moz.tools/pulls/lando/1338/)
Bugzilla: [bug 2056454](https://bugzilla.mozilla.org/show_bug.cgi?id=2056454)

:warning: This pull request has 7 warnings.


src/lando/api/tests/test_try.py
```diff
@@ -1,481 +0,0 @@
...
-import pytest
-
-from lando.api.legacy.workers.landing_worker import LandingWorker
-from lando.main.models import JobStatus, LandingJob, Repo
-from lando.main.scm.helpers import (
-    get_timestamp_from_git_date_header,
-    parse_git_author_information,
-)
-
-pytest.skip(allow_module_level=True)
```
shtrom: Skipped since 2024.

src/lando/try_api/api.py
```diff
@@ -225,10 +225,16 @@ def patches(
...
-    errors = landing_checks.run(
-        repo.hooks,
-        patch_helpers,
-    )
+    try:
+        errors = landing_checks.run(
+            repo.hooks,
+            patch_helpers,
+        )
+    except Exception:
```
zzzeid, line 242: Is there a list of possible exceptions we can use here, instead of a catch-all?

src/lando/try_api/api.py
```diff
@@ -225,10 +225,16 @@ def patches(
...
+        message = "Unexpected error while performing pre-submission patch checks."
+        logger.exception(message)
+        # Prepare message to be handled as a normal error just below.
+        errors = [message]
```
zzzeid, line 246: I'll say my usual blurb here about trying to to expose internals in an error message shown externally (in this case, could be visible publicly).

zzzeid (dismissed):

src/lando/try_api/api.py
shtrom: Ah! But I did hear the blurb q:

Note that `message` here is set on line 234, which is just a non-descript error. There is a `logger.exception`, with the `message` too, that logs and report to sentry, but the original error message is not shown to the user (:

src/lando/try_api/api.py
shtrom: No. We do the same in the landing worker 

https://github.com/mozilla-conduit/lando/blob/351a74286d69ef0d7945fb0e1e9215c478269c9e/src/lando/api/legacy/workers/landing_worker.py#L314-L317

to be able to terminate cleanly, rather than die with a 502 (which is the current behaviour)

shtrom: (In the case of this particular bug, that was an exception from Rust, in rs_parsepatch, that bubbled up all the way).

src/lando/try_api/tests/test_api.py
```diff
@@ -336,6 +336,60 @@ def test_try_api_patches_failed_checks(
...
+    assert response.status_code == 400, (
+        f"Request to Try API with checks raising exceptions should result in 400: {response.text}"
+    )
+
+    rj = response.json()
+    assert rj["title"] == "Errors found in pre-submission patch checks."
+    assert "Patch failed checks:" in rj["detail"]
+    assert (
+        "Unexpected error while performing pre-submission patch checks" in rj["detail"]
+    )
```
zzzeid: nit: also usual blurb about trying to be more explicit in the comparisons. Unless there's some fuzzy value that we can't predict, might as well capture the whole output and compare to an exact string, otherwise this assertion could still pass even if there are unexpected values (along with the expected one) here.

zzzeid (approved):

zzzeid:
Also, I think the PR description/commit message are out of date (I don't see anything relevant to the first two points in this PR).

src/lando/try_api/tests/test_api.py
shtrom: 😅

shtrom:
> Also, I think the PR description/commit message are out of date (I don't see anything relevant to the first two points in this PR).

Ah, yep. The commits are still in the history, but they are noops now. Good catch!

shtrom:
Force pushing due to landing conflicts https://lando.moz.tools/pulls/lando/1338/

PR #1340, opened 2026-07-21
jinja: improve job views for non-Phabricator revisions (bug 2012875)

shtrom:
This replaces the `embedded` boolean with a `page_context` parameter. This allows for finer control of what details of the jobs to show, depending on context:

  * In `timeline` pages: multiple jobs may be present, with shared info presented at the stack level.
  * In Landing `job` pages: a single job, with no info anywhere else.
  * On the `home` page: multiple jobs, but with no shared info.

Feature matrix:

|           | Stack/Pull link | Job link | Tip title | Treestatus link |
| --------- | --------------- | -------- | --------- | --------------- |
|  timeline |                 |    X     |           |           X     |
|       job |           X     |          |    X      |           X     |
|      home |           X     |    X     |    X      |                 |

Stacks and PR views use the `timeline` context via the timeline partial.
Defaults to `home`, which shows most information.
  
Additional page improvements:

  * make the rendering of each type of revision almost similar (but not exactly the same just yet), paving the way for rendering of Stacked PRs;
  * show the Target repository in all contexts;
  * fix a long-standing bug where the error breakdown `<pre>` was not adequately wrapped, and long lines would push the boundary of the page out of the viewport.

---

 * revision: remove deprecated use of Optional (bug 1986844)
 * revision: make __str__ and __repr__ more informative depending on source
 * revision: make url() a PR-aware property
 * jinja: pre-wrap errors in job partials
 * jinja: add page_control to jobs partial
 * jinja: drop phabricator_revision_url for revision list
 * jinja: show target repository on jobs

---

Lando: [link](https://lando.moz.tools/pulls/lando/1340/)
Bugzilla: [bug 2012875](https://bugzilla.mozilla.org/show_bug.cgi?id=2012875)


shtrom:
Some eye-candy:

homepage

try push

pull

narrow stack

src/lando/api/tests/test_transplants.py
```diff
@@ -881,7 +881,8 @@ def test_integrated_transplant_updated_diff_id_reflected_in_landed_phabricator_r
     assert len(result["revisions"]) == 1, (
         f"Revisions list of unexpected length in GET /landing_jobs/{job.id} response"
     )
-    assert result["revisions"][0]["url"] == r1["uri"], (
+    # The diff_id gets appended to the query string
```
shtrom:
```suggestion
    # The diff_id gets appended to the query string.
```

src/lando/main/models/revision.py
```diff
@@ -82,41 +82,64 @@ class Revision(BaseModel):
...
         """Return a human-readable representation of the instance."""
         # Add an identifier for the Phabricator revision if it exists.
-        phab_identifier = (
-            f" [D{self.revision_id}-{self.diff_id}]>"
-            if self.is_phabricator_revision
-            else ""
-        )
-        return f"<{self.__class__.__name__}: {self.id}{phab_identifier}>"
+        descriptor = self.descriptor
+        return f"<{self.__class__.__name__}: {self.id}[{descriptor}]>"
```
shtrom:
```suggestion
        return f"<{self.__class__.__name__}: {self.id}[{self.descriptor}]>"
```

src/lando/main/models/revision.py
```diff
@@ -82,41 +82,64 @@ class Revision(BaseModel):
...
+    @property
+    def descriptor(self) -> str:
+        """Returns a short string identifying the revision by type and ID."""
+        if self.is_phabricator_revision:
+            return f"D{self.revision_id} (diff {self.diff_id})"
 
+        if self.is_pull_request:
+            return f"PR#{self.pull_number}"
+        else:
+            return f"Lando rev. {self.id}"
```
shtrom:
```suggestion

        return f"Lando rev. {self.id}"
```

src/lando/main/models/revision.py
```diff
@@ -82,41 +82,63 @@ class Revision(BaseModel):
...
-    _patch_helper: Optional[HgPatchHelper] = None
+    _patch_helper: HgPatchHelper | None = None
 
     def __str__(self) -> str:
-        if self.is_phabricator_revision:
-            return f"D{self.revision_id} (diff {self.diff_id})"
-        s = f"Revision {self.id}"
-        if desc := self.patch_data.get("commit_message"):
-            s = f"{s} ({desc})"
+        """Return a pretty string representing the revision, with commit information."""
```
zzzeid, line 88: This feels a little overloaded. I think `__str__` should be more deterministic, but we could add the `if desc` logic elsewhere and call something like `revision.commit_info` to get the more detailed description.

src/lando/main/models/revision.py
```diff
@@ -82,41 +82,63 @@ class Revision(BaseModel):
...
+    @property
+    def descriptor(self) -> str:
+        """Returns a short string identifying the revision by type and ID."""
+        if self.is_phabricator_revision:
+            return f"D{self.revision_id} (diff {self.diff_id})"
+
+        if self.is_pull_request:
+            return f"PR#{self.pull_number}"
+
+        return f"Lando rev. {self.id}"
```
zzzeid: Nice. But wondering if we should be a little more verbose, especially since the last return is pretty verbose.

```suggestion
        if self.is_phabricator_revision:
            return f"Revision D{self.revision_id} (diff {self.diff_id})"

        if self.is_pull_request:
            return f"Pull Request #{self.pull_number}"

        return f"Lando Revision {self.id}"
```

On a different topic, it may be time to rename `Revision` in Lando. Will file a bug for this.

src/lando/ui/jinja2/partials/job.html
```diff
@@ -1,44 +1,82 @@
-{% set embedded = embedded|default(True) %}
+{% if false %}
```
zzzeid, line 1: What's this for here? Maybe `{#` and `#}` would be better as a jinja comment?

zzzeid (commented):
Small general comments, but I think that the job.html partial is getting a little too complex with the context variable and nested if statements. It may be cleaner to set the context within the template based on available variables instead.

Even better would be to create macros that are separate from the template, then use those macros in two different templates that are inserted based on the context (e.g., `job_main` and `job_timeline`).

src/lando/ui/jinja2/partials/job.html
shtrom: Ah, yeah, that was to comment out.

src/lando/main/models/revision.py
shtrom: The `str` is used for things like lando admin tables. That's why I wanted to make them humanly readable/distinguishable. The lack of deterministicality (if that's a word) is because Revisions can be Phab (`D...`), PRs (`#...`) or Try pushes (only the commit message, `desc` is really useful).

In comparison, the `repr` is a lot more deterministic in that way, but less user-friendly.

PR #1342, opened 2026-07-22
uplift_worker: generate `try_task_config.json` for uplift try-push dynamically with `mach try` (bug 2050634)

Magnolia-Liu:
Currently, when auto-generating try pushes from uplift requests, we are using a hard-coded try_task_config.json. 

Generate that with mach try instead.

---

Lando: [link](https://lando.moz.tools/pulls/lando/1342/)
Bugzilla: [bug 2050634](https://bugzilla.mozilla.org/show_bug.cgi?id=2050634)


src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -359,11 +360,13 @@ def create_uplift_try_push(
...
-            settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"
+    def create_try_diff_from_json(self, repo: Repo) -> str:
+        """Create a raw diff for the `try_task_config.json` file."""
+        config_contents = self.run_mach_command(
+            repo_path=repo.path,
+            args=["try", "fuzzy", "-q", "^build-", "--no-push"],
+            extra_env={"MOZBUILD_STATE_PATH": repo.mozbuild_state_path},
         )
-        config_contents = try_config_path.read_text()
         config_lines = config_contents.splitlines()
```
cgsheeh: The full stdout of `mach try fuzzy ... --no-push` is spliced verbatim into the `try_task_config.json` diff, but `mach try` output is *not* pure JSON. As written, the generated `try_task_config.json` will be malformed and the try push will fail downstream in Taskcluster.

The JSON needs to be extracted from the mach output before building the diff. Consider locating the JSON payload in the output and round-tripping it through `json.loads`/`json.dumps` so a bad capture fails loudly here rather than producing a broken try push.

src/lando/api/tests/test_uplift.py
```diff
@@ -747,6 +748,13 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
 
     repo = repo_mc(SCMType.GIT, name="firefox-beta", approval_required=True)
 
+    dir = tmp_path / "firefox-beta"
```
cgsheeh: Two things here:

1. `dir` shadows the builtin function, we should use `repo_dir` instead.
2. Once the JSON-extraction fix lands, please make the fake `mach` emit realistic output (JSON embedded in surrounding text) and assert that the generated try revision's diff contains the expected parsed config. Currently the test would still pass if the config came from somewhere else entirely, so it doesn't actually exercise the mach integration.

src/lando/api/legacy/workers/base.py
```diff
@@ -454,3 +455,65 @@ def call_task(task: Task, *args):
...
+        `extra_env` is merged on top of the current process environment rather
+        than replacing it. If it contains `MOZBUILD_STATE_PATH`, the directory
+        is created before the subprocess runs so `mach` can write bootstrapped
+        toolchains and state there instead of the worker's homedir.
+        """
+        if not self.mach_path(repo_path):
+            raise Exception("No `mach` found in local repo!")
+
+        # Convert to `str` here so we can log the mach path.
+        command_args = [str(self.mach_path(repo_path))] + args
```
cgsheeh: `self.mach_path(repo_path)` is called twice (here and on line 469). Since this is now the shared base implementation, worth computing it once.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -359,11 +360,13 @@ def create_uplift_try_push(
             try_job.save()
         return try_job
 
-    def create_try_diff_from_json(self) -> str:
-        try_config_path = (
```
cgsheeh, line 379: We should remove the hard-coded `try_task_config.json` since it's unused after this change. Alternatively, we could fall back to the hard-coded config file if we hit an error in generating from `mach try` dynamically.

cgsheeh (changes requested):

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -348,11 +347,13 @@ def create_uplift_try_push(
...
         with transaction.atomic():
-            revisions = [
-                self.create_revisions_from_patch_helper(patch_helper)
-                for patch_helper in patch_helpers
-            ]
-            try_revision = self.create_try_revision(job.requester_email)
+            revisions = []
+            patch_helpers = scm.get_patch_helpers_for_commits(new_commits)
+            for patch_helper in patch_helpers:
+                revisions.append(self.create_revisions_from_patch_helper(patch_helper))
```
cgsheeh: This recomputes `patch_helpers`, which we already built on line 320, and swaps the list comprehension out for an append loop. Neither change is related to the `mach try` work, so let's revert this back to the comprehension over the existing `patch_helpers`.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -375,12 +376,19 @@ def check_uplift_bug_references(
...
         return secure_check.result()
 
-    def create_try_diff_from_json(self) -> str:
-        try_config_path = (
-            settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"
+    def create_try_diff_from_json(self, repo: Repo) -> str:
+        """Create a raw diff for the `try_task_config.json` file."""
+        config_contents = self.run_mach_command(
+            repo_path=repo.path,
+            args=["try", "fuzzy", "-q", "^build-", "--no-push"],
```
cgsheeh: The hard-coded config we're deleting set `"disable-pgo": true`, but `mach try fuzzy` only emits that key when `--disable-pgo` is passed. Without it we'd silently start running PGO builds on every uplift try push, so let's add the flag here.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -375,12 +376,19 @@ def check_uplift_bug_references(
...
+            extra_env={"MOZBUILD_STATE_PATH": repo.mozbuild_state_path},
         )
-        config_contents = try_config_path.read_text()
-        config_lines = config_contents.splitlines()
+
+        json_start = config_contents.index("{")
+        json_text = config_contents[json_start:]
+        config_json = json.dumps(json.loads(json_text))
```
cgsheeh: `--no-push` prints `Commit message:` followed by the full commit message (which embeds the `mach try` command line) before it prints `Calculated try_task_config.json:`, so `index("{")` is relying on nothing in that preamble ever containing a brace. Let's anchor on the `Calculated try_task_config.json:` marker instead and raise a descriptive error when it's missing, rather than a bare `ValueError`. Also worth passing `indent=4, sort_keys=True` to `json.dumps` so the committed file matches what `mach try` itself would write.

src/lando/api/tests/test_uplift.py
```diff
@@ -749,6 +750,17 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
 
     repo = repo_mc(SCMType.GIT, name="firefox-beta", approval_required=True)
 
+    try_task_config = {"parameters": [], "version": 1}
+    config_json = json.dumps(try_task_config)
+
+    repo_dir = tmp_path / "firefox-beta"
+    repo_dir.mkdir()
+    mach_file = repo_dir / "mach"
+    mach_file.write_text(f"#!/bin/sh\necho '{config_json}'\n")
```
cgsheeh: This is the half of my earlier comment that's still outstanding: the fake `mach` emits bare JSON, so the extraction from surrounding output is never exercised. Please have it print the `Commit message:` / `Calculated try_task_config.json:` preamble ahead of the JSON, and use a realistic config (`parameters` as a dict, `version: 2`).

src/lando/api/tests/test_uplift.py
```diff
@@ -749,6 +750,17 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
...
+
+    mach_file.chmod(0o755)
+    repo.system_path = str(repo_dir)
```
cgsheeh: Two nits here. `system_path` is never saved, so this only works because `job` happens to hold a cached `Repo` instance; a `repo.save()` would make it robust.

Also, `MOZBUILDS_ROOT` is left pointing at the real `/files/mozbuilds`, so the test creates state directories outside `tmp_path` and fails on a non-root local run — the other `mach` tests either point it at `tmp_path` or mock `run_mach_command`.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -375,12 +383,23 @@ def check_uplift_bug_references(
...
 
-    def create_try_diff_from_json(self) -> str:
-        try_config_path = (
-            settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"
+    def create_try_diff_from_json(self, repo: Repo) -> str:
+        """Create a raw diff for the `try_task_config.json` file."""
+        config_contents = self.run_mach_command(
+            repo_path=repo.path,
+            args=["try", "fuzzy", "-q", "^build-", "--no-push", "--disable-pgo"],
+            extra_env={"MOZBUILD_STATE_PATH": repo.mozbuild_state_path},
```
cgsheeh: `mach try fuzzy` needs the `fzf` binary, and our `Dockerfile` only installs `mercurial`. Without `fzf` present, `fzf_bootstrap` will try to prompt for a download and this will fail in the worker. Can you confirm the worker image has `fzf`, or add it?

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -375,12 +383,23 @@ def check_uplift_bug_references(
...
         return secure_check.result()
 
-    def create_try_diff_from_json(self) -> str:
-        try_config_path = (
-            settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"
+    def create_try_diff_from_json(self, repo: Repo) -> str:
+        """Create a raw diff for the `try_task_config.json` file."""
+        config_contents = self.run_mach_command(
+            repo_path=repo.path,
+            args=["try", "fuzzy", "-q", "^build-", "--no-push", "--disable-pgo"],
```
cgsheeh: `^build-` matches every task in the train's target set beginning with `build-`, which is considerably more than the nine builds the hardcoded config pinned (asan, ccov, shippable, signing, etc). Has the extra try load been checked with releng? If we do want the wider set, it's worth saying so in a comment here.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -375,12 +383,23 @@ def check_uplift_bug_references(
...
+        _, _, json_text = config_contents.partition(
+            "Calculated try_task_config.json:\n"
         )
-        config_contents = try_config_path.read_text()
-        config_lines = config_contents.splitlines()
+        if not json_text:
+            raise ValueError(
+                "Could not find 'Calculated try_task_config.json:' marker."
+            )
+        config_json = json.dumps(json.loads(json_text), indent=4, sort_keys=True)
```
cgsheeh: `json.loads` requires that the marker is the last thing `mach` prints. If anything is emitted after the config block we'll fail with `Extra data`. `json.JSONDecoder().raw_decode(json_text)` would let us stop at the end of the object and ignore trailing output.

cgsheeh: Nothing validates the parsed config, so an empty `tasks` list gets turned into a try push that runs nothing. Can we raise if `parameters.try_task_config.tasks` is empty?

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -352,7 +351,16 @@ def create_uplift_try_push(
                 self.create_revisions_from_patch_helper(patch_helper)
                 for patch_helper in patch_helpers
             ]
-            try_revision = self.create_try_revision(job.requester_email)
+            try:
+                try_revision = self.create_try_revision(
```
cgsheeh: `create_try_revision` now shells out to `mach try`, which generates the taskgraph and can take a long time. Running that inside `transaction.atomic()` holds a database transaction open for the whole subprocess. Can we build the try revision before entering the atomic block?

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -352,7 +351,16 @@ def create_uplift_try_push(
...
+                    job.target_repo, job.requester_email
+                )
+            except ValueError:
+                logger.exception(
+                    "Error creating try revision",
+                    extra={"job_id": job.id},
+                )
```
cgsheeh: `run_mach_command` raises a bare `Exception` when `mach` is missing and `CalledProcessError` on failure, so the interesting new failure modes aren't `ValueError` and slip past this. `process` already catches `Exception` around `create_uplift_try_push` and logs with `job_id`, so I'd either widen this to `Exception` or drop it.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -375,12 +383,23 @@ def check_uplift_bug_references(
             secure_check.next_diff(patch_helper)
         return secure_check.result()
 
-    def create_try_diff_from_json(self) -> str:
-        try_config_path = (
-            settings.BASE_DIR / "api" / "legacy" / "workers" / "try_task_config.json"
+    def create_try_diff_from_json(self, repo: Repo) -> str:
```
cgsheeh: The name is stale now that the config comes from `mach try` rather than a JSON file in the repo. Something like `create_try_task_config_diff` would read better.

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -375,12 +383,23 @@ def check_uplift_bug_references(
...
+            extra_env={"MOZBUILD_STATE_PATH": repo.mozbuild_state_path},
+        )
+        _, _, json_text = config_contents.partition(
+            "Calculated try_task_config.json:\n"
         )
-        config_contents = try_config_path.read_text()
-        config_lines = config_contents.splitlines()
+        if not json_text:
+            raise ValueError(
+                "Could not find 'Calculated try_task_config.json:' marker."
```
cgsheeh: Wrap the marker in backticks here for consistency with the rest of our messages.

src/lando/api/legacy/workers/base.py
```diff
@@ -631,3 +632,66 @@ def call_task(task: Task, *args):
             # Log the exception but continue gracefully.
             logger.exception(f"Failed sending {task.__name__} task to Celery.")
             logger.exception(e)
+
+    def run_mach_command(
```
cgsheeh: Sharing this between the two workers makes sense, but `Worker` is our generic abstract base and `mach` is Firefox-specific. Would a small `mach` helper module (or a mixin) be a better home, so `Worker` stays SCM/tool agnostic?

src/lando/api/tests/test_uplift.py
```diff
@@ -749,6 +772,19 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
...
+    mach_file.write_text(
+        "#!/bin/sh\ncat <<'MACH_EOF'\n"
+        + MOCK_MACH_TRY_PREAMBLE
+        + json.dumps(MOCK_TRY_TASK_CONFIG)
+        + "\nMACH_EOF\n"
+    )
+    mach_file.chmod(0o755)
+
+    # `mach_path` lookup is mocked to find the fake `mach` that is not in the git working directory
+    monkeypatch.setattr(uplift_worker, "mach_path", lambda repo_path: mach_file)
```
cgsheeh: The fake `mach` ignores its arguments, so nothing here asserts we actually invoke `try fuzzy -q ^build- --no-push --disable-pgo`. Having the script write `"$@"` to a file and asserting on it would catch an accidental change to the query.

src/lando/api/tests/test_uplift.py
```diff
@@ -749,6 +772,19 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
...
+    mach_file = tmp_path / "mach"
+    mach_file.write_text(
+        "#!/bin/sh\ncat <<'MACH_EOF'\n"
+        + MOCK_MACH_TRY_PREAMBLE
+        + json.dumps(MOCK_TRY_TASK_CONFIG)
+        + "\nMACH_EOF\n"
+    )
+    mach_file.chmod(0o755)
+
+    # `mach_path` lookup is mocked to find the fake `mach` that is not in the git working directory
```
cgsheeh: Nit: end the comment with a period and wrap it at 88 characters.

src/lando/api/tests/test_uplift.py
```diff
@@ -69,6 +70,27 @@
...
+Calculated try_task_config.json:
+"""
+
+MOCK_TRY_TASK_CONFIG = {
+    "parameters": {
+        "optimize_target_tasks": False,
+        "try_task_config": {
+            "disable-pgo": True,
+            "env": {"TRY_SELECTOR": "fuzzy"},
+            "tasks": [],
```
cgsheeh: An empty `tasks` list isn't what `mach try fuzzy -q ^build-` produces, and it means the diff assertion below never exercises any task names. Please use a couple of realistic `build-*` entries.

src/lando/api/tests/test_uplift.py
```diff
@@ -910,6 +946,17 @@ def _write_uplift_commits(job_arg, base_rev, env, output_path):
...
 
+        try_revision = try_job.revisions.last()
+        expected_config_json = json.dumps(
+            MOCK_TRY_TASK_CONFIG, indent=4, sort_keys=True
+        )
+        expected_diff_lines = "\n".join(
+            f"+{line}" for line in expected_config_json.splitlines()
+        )
+        assert expected_diff_lines in try_revision.diff, (
+            "Try revision diff should contain the parsed `try_task_config.json` contents."
```
cgsheeh: The `ValueError` path for a missing marker is untested. Rather than growing this already very long integration test, could we add a focused unit test that calls `create_try_diff_from_json` directly and covers both the happy path and the missing-marker case?

cgsheeh (changes requested):

src/lando/api/legacy/workers/uplift_worker.py
```diff
@@ -431,12 +440,27 @@ def check_uplift_bug_references(
...
+    def create_try_task_config_diff(self, repo: Repo) -> str:
+        """Create a raw diff for the `try_task_config.json` file."""
+        config_contents = self.run_mach_command(
+            repo_path=repo.path,
+            args=["try", "fuzzy", "-q", "^build-", "--no-push", "--disable-pgo"],
+            extra_env={"MOZBUILD_STATE_PATH": repo.mozbuild_state_path},
         )
-        config_contents = try_config_path.read_text()
-        config_lines = config_contents.splitlines()
+        _, _, json_text = config_contents.partition(
```
shtrom: I wonder if we'd be better off adding something like a `--json-only` option to `mach try fuzzy`, rather than having to do the parsing here.

PR #1345, opened 2026-07-23
github_helpers: split Lando-specific GitHub logic (bug 2057141)

shtrom:
This commit separates Lando- and Django-specific GitHub adapters from
the project-agnostic GitHub logic. This mainly moves methods to new
files, though some tidbits of logic are moved to the call-sites, with the goal
to not import any project module from the util.github component.
    
This does the same for tests, separating them to match the split, and
moving GitHub fixtures to a dedicated conftest, which is explicitly
re-imported.

---
Lando: [link](https://lando.moz.tools/pulls/lando/1345/)
Bugzilla: [bug 2057141](https://bugzilla.mozilla.org/show_bug.cgi?id=2057141)


src/lando/utils/github/api.py
```diff
@@ -1,35 +1,38 @@
...
-from django.views import View
 from simple_github import AppAuth, AppInstallationAuth
-from typing_extensions import override
 
-from lando.main.models.configuration import ConfigurationKey, ConfigurationVariable
-from lando.main.scm.helpers import PatchHelper, PatchHelperMetadata
-from lando.utils.cache import cache_method
-from lando.utils.const import URL_USERINFO_RE
+from .cache import cache_method
+from .const import URL_USERINFO_RE
```
shtrom, line 27: I'm cheating a bit here, but both those imports are generic functionality, and can probably tag along or be separately reimplemented as needed depending on the host project.

Magnolia-Liu (approved):
lgtm!

src/lando/utils/github.py
```diff
@@ -1,35 +1,38 @@
+"""This file contains project-independent GitHub utils.
+
+WARNING: Do not import and lando- or django-namespaced modules here.
```
cgsheeh: Typo: `Do not import and lando-` should be `any lando-`.

src/lando/utils/github.py
```diff
@@ -1,35 +1,38 @@
...
-from lando.utils.cache import cache_method
-from lando.utils.const import URL_USERINFO_RE
+from .cache import cache_method
+from .const import URL_USERINFO_RE
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class settings:
```
cgsheeh: Can you add a docstring for this to clarify the intent? Seems odd to have a class with the name `settings`. Is there another way we could implement this, maybe with Django's `AppConfig`?

src/lando/settings.py
```diff
@@ -337,6 +338,11 @@
 
 HTTP_USER_AGENT = f"Lando/{version} ({ENVIRONMENT})"
 
+# Forward our own config to the GitHub utils'.
+github_settings.GITHUB_APP_ID = GITHUB_APP_ID
```
cgsheeh: Two concerns with push-forwarding from here. Importing `lando.utils.github` pulls `requests`, `simple_github` and `django.core.cache` in while the settings module is still being evaluated, which is a fragile place to grow a dependency chain. And it snapshots the values, so `override_settings(GITHUB_APP_ID=...)` in a test would silently not reach the util; a `configure()` call from an `AppConfig.ready()` would avoid both.

src/lando/utils/github_helpers.py
```diff
@@ -0,0 +1,111 @@
+"""This file contains lando-specific GitHub logic."""
+
+import functools
+import io
+import json
+import math
+from datetime import datetime
+from json.decoder import JSONDecodeError
+from typing import Callable
```
cgsheeh: `Callable` should be imported from `collections` instead of `typing`.

src/lando/utils/github_helpers.py
```diff
@@ -0,0 +1,111 @@
...
+        """Return the patch diff.
+
+        WARNING: As of 2025-10-13, this doesn't include any binary data.
+        """
+        return self._diff
+
+    @override
+    def write(self, f: io.StringIO):
+        """Implement the PatchHelper interface; not relevant for GitHub PRs."""
+        raise NotImplementedError("`from_bytes_io` not implemented.")
```
cgsheeh, line 89: Copy-paste error: `from_bytes_io` should be `write`.

src/lando/utils/tests/conftest_github.py
```diff
@@ -0,0 +1,336 @@
...
+                    "content-type": "application/json; charset=utf-8",
+                },
+            ),
+        }
+        # Unpaged request gives out the first page.
+        response_map[f"repos/{repo}/pulls/{pr_no}/commits"] = response_map[
+            f"repos/{repo}/pulls/{pr_no}/commits?page=0"
+        ]
+
+        def _mock_api_get(url: str, headers: dict = dict, **kwargs) -> Response:
```
cgsheeh: Should this be `headers: dict | None = None`?

conftest.py
```diff
@@ -0,0 +1,4 @@
+# pytest_plugins needs to be defined in the root conftest.py.
+pytest_plugins = [
+    "lando.utils.tests.conftest_github",
```
cgsheeh: Registering this globally makes the fixtures visible everywhere, but `github_api_client` depends on `mock_response`, which still lives in `src/lando/conftest.py` — so they only actually resolve for tests under `src/lando/`. Either move `mock_response` alongside them or note the dependency here.

cgsheeh (changes requested):

src/lando/utils/github/tests/conftest.py
zzzeid, line 1: Would it be feasible to move this along with `test_github` and `test_github_helpers` to `tests/github/conftest.py` etc.? That would simplify naming a little and go with the existing conventions (i.e., `conftest` as a filename).

src/lando/utils/github/api.py
zzzeid, line 1: On second thought, we could move everything into `lando.utils.github`. I.e.:

```
lando/utils/github/__init__.py (this file)
lando/utils/github/tests/* (conftest and other tests)
lando/utils/github/helpers.py (github_helpers.py)
```

Unless this breaks things without too many more changes.

src/lando/utils/github/api.py
zzzeid: I think this is a little funny -- is there a better way to do this? Seems like the only benefit is syntactic (i.e., keeping `lando` out of this) -- an inline comment could be sufficient here, or perhaps delegating this to a third module that we import from (e.g., `lando.utils.github.external`).

src/lando/utils/github.py
zzzeid: Agree re: renaming, but given that this is meant to be agnostic of Django/etc. maybe we can just rename this to `GitHubSettings`, which can be consumed by other things as needed. I'm not opposed to also simply keeping these as individual variables without putting them in a class.

src/lando/settings.py
```diff
@@ -337,6 +338,11 @@
...
+github_settings.GITHUB_APP_PRIVKEY = GITHUB_APP_PRIVKEY
+github_settings.HTTP_USER_AGENT = HTTP_USER_AGENT
```
zzzeid: This is a little funky here. What do you mean by "Forward" here? What purpose does it serve within Lando's settings concept? Since we're not instantiating this class I think this could cause unexpected behaviour.

zzzeid (changes requested):
First pass

src/lando/utils/github/api.py
shtrom: Yep. The goal is 100% to keep lando out of it. I'll add a comment.

src/lando/utils/github/tests/conftest.py
shtrom: I can move test_github and conftest. I'm not sure about test_github_helpers, as that's the lando glue which is supposed to stay behind.

src/lando/utils/github/api.py
shtrom: Ah, that works!

But same reluctance about the helper.

src/lando/settings.py
shtrom: I want to keep all the lando settings configured in one place, this file. However, the GitHub settings are now read from a different place (that uninstantiated class). So this allows to keep the settings where we expect them, while making sure they are where the lib actually expects them.

I'm not too shocked by the uninstantiated class. We could make it a singleton instance and/or provide a setup method. But this all feels like adding more complexity to achieve the same result.

WDYT?

shtrom:
Ok, I fixed all the bits, and moved everything but the lando `github_helper` into a separate module, and merged `main` in.

src/lando/utils/github/api.py
```diff
@@ -1,35 +1,40 @@
...
 from simple_github import AppAuth, AppInstallationAuth
-from typing_extensions import override
 
-from lando.main.models.configuration import ConfigurationKey, ConfigurationVariable
-from lando.main.scm.helpers import PatchHelper, PatchHelperMetadata
-from lando.utils.cache import cache_method
-from lando.utils.const import URL_USERINFO_RE
+# We deliberately import those without a fully-qualified import to allow portability of
+# the file away from Lando.
+from ..cache import cache_method
```
cgsheeh: `lando.utils.cache` does `from django.core.cache import caches`, so this module still pulls Django in transitively and the `WARNING` at the top of the file isn't actually true yet. Could we move `cache_method` and `URL_USERINFO_RE` into the `github` package (or inject the cache) so the package is genuinely standalone?

src/lando/utils/github/api.py
```diff
@@ -1,35 +1,40 @@
...
+from ..const import URL_USERINFO_RE
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass
```
cgsheeh: `@dataclass` is misleading here since we never instantiate this and every read/write goes through the class attributes.

src/lando/settings.py
```diff
@@ -338,6 +339,11 @@
 
 HTTP_USER_AGENT = f"Lando/{version} ({ENVIRONMENT})"
 
+# Forward our own config to the GitHub utils'.
+GitHubSettings.GITHUB_APP_ID = GITHUB_APP_ID
```
cgsheeh: Forwarding at settings-import time means anything that changes these later — `test_settings`, `override_settings` — silently never reaches the GitHub code. Importing `lando.utils.github` from `settings.py` also drags `django.core.cache` in while settings are still being defined, which is a fragile place to be. Could `UtilsConfig.ready()` do the forwarding instead, reading from `django.conf.settings`?

src/lando/utils/github/tests/conftest.py
```diff
@@ -194,9 +104,7 @@ def github_pr_response() -> str:
             https://api.github.com/repos/mozilla-conduit/test-repo/pulls/1 \
             > src/lando/utils/tests/data/github_api_response_pull.json
```
cgsheeh: These `curl` output paths are now stale — the data files moved to `src/lando/utils/github/tests/data/`. Same on line 122.

conftest.py
```diff
@@ -0,0 +1,4 @@
+# pytest_plugins needs to be defined in the root conftest.py.
+pytest_plugins = [
```
cgsheeh: This registers the GitHub fixtures globally for every test session, and it's only needed because `test_github_helpers.py` sits outside the `github` package. Given `src/lando/conftest.py` is already a session-wide conftest, could the re-export live there, or the glue test move under `src/lando/utils/github/tests/`?

src/lando/utils/github_helpers.py
cgsheeh: Still the copy-paste error from my last pass: this should be `write` rather than `from_bytes_io`.

cgsheeh (approved):

src/lando/utils/github/api.py
shtrom: Yes, it's a bit of a dirty trick: I don't mind pulling some Django stuff BUT I don't want it hardcoded in this file. As it stands, this means I can take that file and drop it somewhere else. If it has a `cache.py` implementation in the parent, it will be used.

Maybe I could gate that with a `try` around the import, and provide a noop `cache_method` decorator otherwise, but as it's a temporary stop-gap, I think it's probably OK.

The injection idea may be the way to go, perhaps in a similar way as we do with the settings.

In any case, I don't think I want to address this right now, but I agree it's something we'll have to revisit.

shtrom: -> https://bugzilla.mozilla.org/show_bug.cgi?id=2070807

zzzeid (approved):
lgtm

PR #1346, opened 2026-07-23
scm: add ScmErrorBreakdown dataclass (bug 1941245)

shtrom:
---

Lando: [link](https://lando.moz.tools/pulls/lando/1346/)
Bugzilla: [bug 1941245](https://bugzilla.mozilla.org/show_bug.cgi?id=1941245)

### DIFF
--- a/src/lando/main/scm/abstract_scm.py
+++ b/src/lando/main/scm/abstract_scm.py
@@ -3,7 +3,7 @@ import random
 import string
 from abc import ABC, abstractmethod
 from contextlib import AbstractContextManager
-from dataclasses import dataclass, field
+from dataclasses import asdict, dataclass, field
 from pathlib import Path
 from typing import Any, Iterable, Self, override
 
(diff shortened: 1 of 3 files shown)


src/lando/main/scm/abstract_scm.py
```diff
@@ -13,6 +16,86 @@
...
+            return self.__dict__
+
+    @dataclass
+    class RejectPath:
+        path: str
+        content: str
+
+        def to_dict(self) -> dict[str, str]:
+            return self.__dict__
+
```
cgsheeh: Any reason we can't un-nest these classes?

cgsheeh (commented):
Wanted to do a quick drive by to say _hell yeah_! 🎉

src/lando/main/scm/abstract_scm.py
shtrom: Unless unnesting refers to a Python technique I don't know of, the reason why we have multiple dataclasses is that the error breakdown contains collections of those two.

That said, I could use `dataclass.asdict` instead, that would be cleaner.

cgsheeh: No no, I meant can you move `FailedPath` and `RejectPath` so they are defined outside of `ScmErrorBreakdown`.

shtrom: Ah, right. I wanted to keep the class a bit more self-contained, but it doesn't make much of a difference. Will do.

PR #1348, opened 2026-07-24
treestatus: add re-runnable import_treestatus_data command (bug 1984161)

cgsheeh:
Add a management command that imports trees and their log history from an
existing Treestatus instance (defaulting to old-Lando). The command is safe to
re-run: existing trees are reused and only logs missing from the local database
are created, so subsequent runs pick up rows added to the source since the last
import.

---
Lando: [link](https://lando.moz.tools/pulls/lando/1348/)
Bugzilla: [bug 1984161](https://bugzilla.mozilla.org/show_bug.cgi?id=1984161)


src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def import_log(self, tree: Tree, log_entry: dict) -> bool:
+        """Create a `Log` entry unless a matching one already exists.
+
+        Deduplicates on the tree and the source `when` timestamp so re-runs do
+        not recreate logs. Returns `True` when a new `Log` is created and `False`
+        when a matching entry is already present.
+        """
+        when = parse_datetime(log_entry["when"])
+        if Log.objects.filter(tree=tree, created_at=when).exists():
```
shtrom: Are we sure we don't have weird time-rounding boundaries here?

shtrom (approved):
:shipit:

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Switched to using the `id` field to ensure de-duplication. IDs should align between the two services so we don't need to worry about timestamp weirdness.

src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -78,37 +82,59 @@ def import_tree(self, base_url: str, tree_info: dict):
...
+                "changed_by": log_entry["who"],
+                "status": TreeStatus(log_entry["status"]),
+                "reason": log_entry["reason"],
+                "tags": log_entry["tags"],
+            },
         )
 
+        # A row already stored under this id must describe the same source entry;
+        # otherwise the two databases have diverged and references are corrupt.
+        if log.tree_id != tree.tree:
```
shtrom: At this point, should we check the other values (status, reason, ...)?

shtrom (approved):

src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -0,0 +1,140 @@
...
+                f"{tree.tree} in the source; the databases are misaligned."
+            )
+
+        if not created:
+            return False
+
+        # `created_at`/`updated_at` use `auto_now_add`/`auto_now`, so a queryset
+        # update is needed to retain the historical timestamp from the source.
+        when = parse_datetime(log_entry["when"])
+        Log.objects.filter(pk=log.pk).update(created_at=when, updated_at=when)
```
zzzeid: We should keep `updated_at` as-is, as within the meaning of the field, the existing values are relevant (i.e., migration time.)

src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -0,0 +1,140 @@
...
+            },
+        )
+
+        # A row already stored under this id must describe the same source entry;
+        # otherwise the two databases have diverged and references are corrupt.
+        if log.tree_id != tree.tree:
+            raise CommandError(
+                f"Log {source_id} maps to tree {log.tree_id} locally but "
+                f"{tree.tree} in the source; the databases are misaligned."
+            )
```
zzzeid: Unclear what would cause this to happen. Can you elaborate as to what could cause this misalignment?

>otherwise the two databases have diverged and references are corrupt

src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -0,0 +1,140 @@
...
+        log, created = Log.objects.get_or_create(
+            id=source_id,
+            defaults={
+                "tree": tree,
+                "changed_by": log_entry["who"],
+                "status": TreeStatus(log_entry["status"]),
+                "reason": log_entry["reason"],
+                "tags": log_entry["tags"],
+            },
+        )
```
zzzeid: Ideally we don't set the `id` field here. Rather than relying on the original primary keys for references, it would be more robust to determine the relationships between the objects more concretely (i.e., have a mapping between the old and the new objects) and use the mapping to set the foreign keys where they need to be set. This would avoid having to force the primary keys in the log table, and avoids having to reset sequences etc...

This might also have some impact on the audit log, and it will also cause a mess if new entries are locally added between runs of this script, somewhat rendering the re-runnability of this script moot.

src/lando/treestatus/management/commands/import_treestatus_data.py
```diff
@@ -0,0 +1,140 @@
...
+        if created:
+            self.stdout.write(f"Created tree {tree_name}.")
+        else:
+            self.stdout.write(f"Tree {tree_name} already exists, importing new logs.")
+
+        logger.debug(f"Fetching logs for {tree_name}.")
+        logs_response = requests.get(f"{base_url}/trees/{tree_name}/logs_all")
+        logs_response.raise_for_status()
+        logs = logs_response.json()["result"]
```
zzzeid: Would be more consistent /cleaner to put this in `import_log`, and move `self.import_log` below to `handle`, i.e., 

```python
for tree_info in trees_data.values():
    self.import_tree(base_url, tree_info)
    self.import_log(base_url, tree_info)
```

src/lando/treestatus/tests/test_treestatus.py
```diff
@@ -958,3 +962,218 @@ def test_api_get_stack(client, new_treestatus_tree):
...
+            call_command("import_treestatus_data", IMPORT_BASE_URL)
+
+    assert Tree.objects.count() == 2, "Re-running should not duplicate trees."
+    assert Log.objects.filter(tree="autoland").count() == 2, (
+        "Re-running should not duplicate log entries."
+    )
+
+
+@pytest.mark.django_db
+def test_import_treestatus_data_imports_new_logs_on_rerun():
```
zzzeid: This test seems like a dupe / special case of `test_import_treestatus_data_imports_logs_for_existing_tree`.

src/lando/treestatus/tests/test_treestatus.py
```diff
@@ -958,3 +962,218 @@ def test_api_get_stack(client, new_treestatus_tree):
...
+    logs = list(Log.objects.filter(tree="autoland").order_by("created_at"))
+    assert len(logs) == 3, "Only the newly added source log should be imported."
+    assert logs[-1].id == 3, "The new log should reuse its source id."
+    assert logs[-1].reason == "burning", (
+        "The most recent log should be the newly added source entry."
+    )
+
+
+@pytest.mark.django_db
+def test_import_treestatus_data_aborts_on_id_misalignment(new_treestatus_tree):
```
zzzeid: Unclear still when this would happen? If this is a possibility, it should be accounted for in the script, or the original data should be parsed/cleaned before being imported, since it means we can't migrate the data fully on first import.

If this is intended to handle the case when we re-run the import after we've already modified local data, we should simply delete all records before re-importing a fresh version of the data, since presumably the imported data is no longer valid (and presumably at that point we have not migrated to the new treestatus yet.) This would also get rid of a lot of the special handling in this PR of those cases.

src/lando/treestatus/management/commands/import_treestatus_data.py
zzzeid, line 1: We should file a bug to remove this script and related tests after the migration.

zzzeid (changes requested):

src/lando/treestatus/management/commands/import_treestatus_data.py
shtrom: Should we have a `legacy_updated_at` field to retain the `when` here?

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Done: https://bugzilla.mozilla.org/show_bug.cgi?id=2062480

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Left as-is. `Log.to_dict()` maps `when` to `created_at`, not `updated_at`, so the source `when` is fully preserved by only setting `created_at`; `updated_at` now keeps its local meaning of "when this row was last written", which for an imported log is the import time. No need for a `legacy_updated_at` field. There is a test asserting both.

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Switched the import to a full re-import: all local logs are deleted before each run, then recreated under the source ids. That guarantees the alignment rather than detecting divergence after the fact.

We do want the ids to align rather than mapping old ids to new ones: log ids are part of the Treestatus API surface (a tree's `log_id`, and the id clients pass back to amend a log), so keeping them identical means references stay valid across both services. The command is documented as only usable while the source instance is authoritative — after the cutover it would discard changes made in Lando.

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Removed. With the delete-and-re-import approach the misalignment case cannot arise, so the check and its test are gone.

src/lando/treestatus/tests/test_treestatus.py
cgsheeh: Agreed — went with your suggestion: all records are deleted before re-importing a fresh copy of the source data. The misalignment test is gone, and the re-run test now asserts that a locally-created log is dropped and the source logs are re-created under their source ids.

src/lando/treestatus/tests/test_treestatus.py
cgsheeh: Merged into a single `test_import_treestatus_data_rerun_replaces_logs_with_source`, which covers both the re-run and the new-source-log cases.

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Done — `import_logs` now fetches and imports the logs, and `handle` calls `import_tree` then `import_logs` per tree.

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Resolved by the re-import: the logs are deleted and recreated from the source every run, so all fields always match the source. Trees now use `update_or_create` so their fields are re-synced too.

src/lando/treestatus/management/commands/import_treestatus_data.py
cgsheeh: Obsolete — no timestamp comparison is done any more; logs are deleted and re-created from the source each run.

PR #1352, opened 2026-07-24
Bug 2055604: Add blocker that ensures status flags are set on sec-high and sec-critical bugs

tomrittervg:
I have not figured out how to test this locally, but I'm hoping the repo's checks will at least run some set of tests for me...

---

Lando: [link](https://lando.moz.tools/pulls/lando/1352/)
Bugzilla: [bug 2055604](https://bugzilla.mozilla.org/show_bug.cgi?id=2055604)


src/lando/api/legacy/transplants.py
```diff
@@ -139,6 +161,8 @@ class StackAssessmentState:
     secure_project_phid: str
     testing_tag_project_phids: list[str]
     testing_policy_phid: str
+    security_bug_flags: dict[int, dict] = field(default_factory=dict)
```
cgsheeh: Name is a bit off: this holds every bug in the stack, not just security ones, and full bug dicts, not flags. `bugs_by_id` would read truer.

src/lando/api/legacy/transplants.py
```diff
@@ -180,6 +204,28 @@ def from_assessment(
             for phid, revision in stack_data.revisions.items()
         }
 
+        # Fetch keywords and status flags for every bug referenced by the stack in a
+        # single privileged BMO request. This lets the security status-flag checks
+        # determine which bugs are sec-high/sec-critical and whether their Firefox
+        # status flags are set. If BMO is unreachable, `bmo_reachable` is recorded as
+        # False so the checks can degrade from a hard blocker to an acknowledgeable
+        # warning.
+        bug_ids = {
```
cgsheeh: This puts network IO in `from_assessment`, whose docstring says it builds "fields that are shared between checks but are derived from existing fields". Every other external fetch (`user_search`, `project_search`, `get_secure_project_phid`) happens in `build_stack_assessment_state` and gets passed in. Doing the same here keeps the constructor pure and makes it trivially mockable.

src/lando/api/legacy/transplants.py
```diff
@@ -180,6 +204,28 @@ def from_assessment(
...
+            for revision in stack_data.revisions.values()
+            if (bug_id := get_bugzilla_bug(revision)) is not None
+        }
+        security_bug_flags: dict[int, dict] = {}
+        bmo_reachable = True
+        if bug_ids:
+            try:
+                params = {"id": ",".join(str(bug) for bug in sorted(bug_ids))}
+                bugs = uplift_get_bug(params)["bugs"]
+                security_bug_flags = {int(bug["id"]): bug for bug in bugs}
```
cgsheeh: `["bugs"]` and `int(bug["id"])` raise `KeyError`/`TypeError`, which aren't caught below — a malformed payload 500s the stack page. Since the surrounding code deliberately degrades gracefully, it should cover this too.

src/lando/api/legacy/transplants.py
```diff
@@ -180,6 +204,28 @@ def from_assessment(
...
+            except requests.exceptions.RequestException:
+                logger.exception("Failed to fetch bug status flags from BMO.")
+                bmo_reachable = False
```
cgsheeh: `raise_for_status` makes 401/403 a `RequestException`, so an expired or misconfigured `BUGZILLA_API_KEY` silently downgrades the hard block to a click-through warning while looking like a transient outage. Worth separating connection errors/5xx from 4xx and alerting on the latter.

src/lando/api/legacy/transplants.py
```diff
@@ -458,6 +506,25 @@ def warning_revision_secure(
     )
 
 
+@RevisionWarningCheck(
+    "Security status flags could not be verified because Bugzilla is unavailable."
+)
+def warning_security_bug_status_flags_unverified(
+    revision: dict, diff: dict, stack_state: StackAssessmentState
```
cgsheeh, line 529: This fires for every revision with any bug, not just security ones. During a BMO outage every landing in Lando gains an unacknowledged warning — and since warnings feed `confirmation_token`, BMO flapping between page render and submit produces "the warnings present when the request was constructed have changed". Gating on `revision_is_secure(revision, stack_state.secure_project_phid)` would cut the blast radius considerably.

src/lando/api/legacy/transplants.py
```diff
@@ -862,6 +929,36 @@ def blocker_try_task_config(
         return issues[0]
 
 
+def blocker_security_bug_status_flags(
+    revision: dict, diff: dict, stack_state: StackAssessmentState
+) -> str | None:
+    """Block sec-high/sec-critical revisions whose Firefox status flags are unset.
+
+    If BMO was unreachable this returns `None`, deferring to
+    `warning_security_bug_status_flags_unverified`."""
```
cgsheeh: The check isn't scoped to Firefox repos — it runs on every Phabricator repo, so a Thunderbird or NSS sec-high bug gets blocked on `cf_status_firefox*` flags that don't apply. We don't have an explicit "is this a Firefox repo" flag, but you could check if `Repo.name.startswith("firefox-")` as a proxy.

Nit: closing `"""` goes on its own line elsewhere in this file.

src/lando/api/legacy/transplants.py
```diff
@@ -862,6 +929,36 @@ def blocker_try_task_config(
...
+    if not stack_state.bmo_reachable:
+        return None
+
+    bug_id = get_bugzilla_bug(revision)
+    if bug_id is None:
+        return None
+
+    bug = stack_state.security_bug_flags.get(bug_id)
```
cgsheeh: A bug missing from the BMO response passes with no blocker and no warning. That's tjr's "one wrinkle is that Lando would need access to read security bugs" case (bug 2055604 comment 2): a sec bug already in a `core-security` group Lando's key can't see gets waved through. Should warn at minimum.

src/lando/api/legacy/transplants.py
```diff
@@ -862,6 +929,36 @@ def blocker_try_task_config(
...
+    if not bug or not is_security_bug(bug):
+        return None
+
+    missing_flags = unset_status_flags(bug)
+    if not missing_flags:
+        return None
+
+    return (
+        f"Bug {bug_id} is marked sec-high but is missing Firefox "
```
cgsheeh: Nit: hardcoded "sec-high" means a sec-critical bug gets told it's marked sec-high. Return the matched keyword from `is_security_bug`, or just say "sec-high/sec-critical".

src/lando/api/tests/test_transplants.py
```diff
@@ -484,6 +487,109 @@ def test_warning_revision_secure_is_not_secure(
     assert warning_revision_secure(revision, {}, stack_state) is None
 
 
+def _security_flags_state(
```
cgsheeh: The `side_effect=... if down else None, return_value=None if down else ...` pairing is hard to follow; two explicit branches would read better.

src/lando/api/tests/test_transplants.py
```diff
@@ -484,6 +487,109 @@ def test_warning_revision_secure_is_not_secure(
...
+                "cf_status_firefox130": "affected",
+                "cf_status_firefox129": "---",
+            }
+        ],
+    )
+
+    result = blocker_security_bug_status_flags(revision, {}, stack_state)
+    assert result is not None
+    assert "cf_status_firefox129" in result
+    # A set flag is not reported as missing.
```
cgsheeh: Nit: adding assert strings to the test cases would be a nice addition.

src/lando/api/tests/test_transplants.py
```diff
@@ -484,6 +487,109 @@ def test_warning_revision_secure_is_not_secure(
...
+def test_blocker_security_flags_defers_when_bmo_down(phabdouble, create_state):
+    revision, stack_state = _security_flags_state(phabdouble, create_state, down=True)
+
+    # BMO unreachable: the blocker defers to the warning instead of hard-blocking.
+    assert stack_state.bmo_reachable is False
+    assert blocker_security_bug_status_flags(revision, {}, stack_state) is None
+
+
+@pytest.mark.django_db
+def test_warning_security_flags_unverified_when_bmo_down(phabdouble, create_state):
```
cgsheeh: Coverage gaps worth filling: sec-critical message text (would catch the hardcoded "sec-high"), a bug absent from the BMO response, a revision with no bug ID, `?` and `None` flag values, a non-Firefox repo, and an integration test through `run_landing_checks` confirming the blocker surfaces in the assessment and removes the node from the landable stack.

cgsheeh (changes requested):
Two more things we should do:

- Add a similar check for GitHub PRs, in `src/lando/utils/landing_checks.py`.
- Double-check that the Lando-uplift endpoint in BMO returns the values we expect. IIRC that endpoint was scoped to only return and provide access to a small subset of relevant fields.

src/lando/api/legacy/transplants.py
```diff
@@ -68,6 +71,36 @@
     defaults=(None, None, None, False),
 )
 
+SECURITY_KEYWORDS = ("sec-critical", "sec-high")
+STATUS_FLAG_PREFIX = "cf_status_firefox"
+UNSET_STATUS_FLAG_VALUES = frozenset({"---", ""})
```
shtrom: What's the purpose of the frozenset here?

src/lando/api/legacy/transplants.py
```diff
@@ -160,12 +195,16 @@ def from_assessment(
         secure_project_phid: str,
         testing_tag_project_phids: list[str],
         testing_policy_phid: str,
+        bugs_by_id: dict[int, dict] | None = None,
+        bmo_data_available: bool = True,
```
shtrom: Should it be false by default, if `bugs_by_id` is None, by default.

Or perhaps we can just rely on `bugs_by_id` being None, and don't need this additional value?

src/lando/api/legacy/transplants.py
```diff
@@ -463,6 +504,59 @@ def warning_revision_secure(
...
+    Firefox status flags (`cf_status_firefox*`) only apply to Firefox repos, so
+    the security status-flag checks are scoped to them. Non-Firefox repos
+    (Thunderbird, NSS, ...) are out of scope. We have no explicit "is Firefox"
+    flag, so the repo name (`firefox`, `firefox-beta`, `firefox-esr...`, ...) is
+    used as a proxy.
+    """
+    repo = stack_state.get_repo_for_revision(revision)
+    if repo is None:
+        return False
+    return repo.name == "firefox" or repo.name.startswith("firefox-")
```
shtrom: I wonder if we should instead add a flag on the Repo object to state whether status flags are required (or maybe a list of which ones are).

Maybe we could add an optional text field that would contain the `STATUS_FLAG_PREFIX`, and condition the blocker based on it being set on the target repo.

src/lando/api/legacy/transplants.py
```diff
@@ -1022,6 +1158,52 @@ def get_parsed_diffs(
...
+        return {}, True
+
+    params = {"id": ",".join(str(bug) for bug in sorted(bug_ids))}
+    try:
+        bugs = uplift_get_bug(params)["bugs"]
+        return {int(bug["id"]): bug for bug in bugs}, True
+    except requests.exceptions.HTTPError as exc:
+        status_code = exc.response.status_code if exc.response is not None else None
+        if status_code is not None and 400 <= status_code < 500:
+            logger.error(
```
shtrom: This should be a `logger.warning`.

src/lando/api/legacy/transplants.py
```diff
@@ -1022,6 +1158,52 @@ def get_parsed_diffs(
...
+                "BMO rejected the Lando status-flag request with HTTP %s; check "
+                "BUGZILLA_API_KEY. Security status-flag checks will degrade to a "
+                "warning until this is resolved.",
+                status_code,
+            )
+        else:
+            logger.exception("BMO returned an error fetching bug status flags.")
```
shtrom:
```suggestion
            logger.warning("BMO returned an error fetching bug status flags.")
```

src/lando/api/tests/test_transplants.py
```diff
@@ -484,6 +488,259 @@ def test_warning_revision_secure_is_not_secure(
     assert warning_revision_secure(revision, {}, stack_state) is None
 
 
+def _security_flags_state(
```
shtrom: Could this be made into a fixture?

shtrom (commented):
LGTM in general, though I think we should have the `STATUS_FLAG` value set on the `Repo` object.

It may also make sense to encapsulate all the new logic into a class, even if as staticmethods, as it would make it easier to consider them as a unit later on.

src/lando/api/legacy/transplants.py
tomrittervg: No real purpose, just making it constant, I'll switch it to match

tomrittervg:
I added a commit that adds the status flags to the repo object and adds a migration

src/lando/main/migrations/0065_repo_status_flag_prefix.py
shtrom, line 1: Ah, we'll also need a data migration script function similar to https://github.com/mozilla-conduit/lando/blob/main/src/lando/utils/management/commands/migrate_data.py#L73-L101, to update the existing Fx repos.

src/lando/api/legacy/transplants.py
```diff
@@ -68,6 +71,34 @@
...
+    `prefix` is the target repo's configured status-flag prefix (e.g.
+    `cf_status_firefox`).
+    """
+    return sorted(
+        name
+        for name, value in bug.items()
+        if name.startswith(prefix)
+        and (value is None or value in UNSET_STATUS_FLAG_VALUES)
+    )
+
```
shtrom: Sorry, I'm realising now that those would be better in `src/lando/api/legacy/bmo.py`

shtrom (commented):
Looking pretty good. We'll need a data migration script to set the flag on existing Fx repo. Is this for all of them? Or maybe all-but-autoland?

I just realised that some of the BMO-only logic would better be moved in the legacy.bmo file.

Finally, it would be good to have matching warning and checks for GitHub (in github_checks.py). They don't rely on a Phabricator for the metadata, but should be essentially similar, inspecting the PullRequest object, rather than the stack.

src/lando/utils/github_checks.py
```diff
@@ -14,6 +16,14 @@
 logger = logging.getLogger("__name__")
 
 
+def pull_request_bug_ids(pull_request: PullRequest) -> set[int]:
+    """Return the set of bug ids referenced by the PR's commit messages."""
+    bug_ids: set[int] = set()
+    for commit in pull_request.commits:
+        bug_ids.update(parse_bugs(commit["commit"]["message"]))
+    return bug_ids
```
shtrom: Let's make this a property on the `PullRequest` object, say, `PullRequest.bug_ids`.

src/lando/api/tests/test_transplants.py
```diff
@@ -484,6 +488,236 @@ def test_warning_revision_secure_is_not_secure(
...
+        repo = phabdouble.repo(name=repo_name)
+        projects = [secure_project] if secure_project is not None else []
+        revision = phabdouble.api_object_for(
+            phabdouble.revision(repo=repo, bug_id=bug_id, projects=projects),
+            attachments={"reviewers": True, "reviewers-extra": True, "projects": True},
+        )
+
+        if bmo_down:
+            patched = mock.patch(
+                "lando.api.legacy.transplants.uplift_get_bug",
```
cgsheeh: This patches `lando.api.legacy.transplants.uplift_get_bug`, but `transplants` imports `fetch_bugs` from `bmo` and never imports `uplift_get_bug`. All 15 new tests in this file error in CI with `AttributeError: <module 'lando.api.legacy.transplants'> does not have the attribute 'uplift_get_bug'`. Patch `lando.api.legacy.bmo.uplift_get_bug` instead (or mock `transplants.fetch_bugs` directly).

src/lando/api/legacy/transplants.py
```diff
@@ -1068,6 +1185,8 @@ def build_stack_assessment_state(
     testing_tag_project_phids = get_testing_tag_project_phids(phab)
     testing_policy_phid = get_testing_policy_phid(phab)
 
+    bugs_by_id = fetch_bugs_for_stack(stack_data)
```
cgsheeh: This adds an uncached BMO round-trip to every stack assessment, including stacks targeting repos with no `status_flag_prefix` (Thunderbird, NSS, etc). The GitHub blocker short-circuits on the prefix before fetching; can we do the same here and skip the fetch when no repo in the stack has a prefix configured?

src/lando/utils/github_checks.py
```diff
@@ -600,6 +669,62 @@ def _authors_str(cls, authors: Iterable[str]) -> str:
...
+        """
+        prefix = target_repo.status_flag_prefix
+        if not prefix:
+            return []
+
+        bug_ids = pull_request_bug_ids(pull_request)
+        if not bug_ids:
+            return []
+
+        bugs_by_id = fetch_bugs(bug_ids)
```
cgsheeh: The blocker and this warning each call `fetch_bugs` independently, so we make two BMO requests per check run and the two can disagree if one of them fails (blocker sees the bug, warning doesn't, or vice versa). Worth fetching once and sharing the result, the way `StackAssessmentState.bugs_by_id` does on the Phabricator side.

src/lando/utils/github_checks.py
```diff
@@ -600,6 +669,62 @@ def _authors_str(cls, authors: Iterable[str]) -> str:
...
+
+        messages = []
+        for bug_id in sorted(bug_ids):
+            if bugs_by_id is not None and bug_id in bugs_by_id:
+                # Data was available and the bug was found; the blocker assessed it.
+                continue
+
+            messages.append(
+                f"Lando could not verify the status flags on bug {bug_id} in "
```
cgsheeh: When BMO is unreachable this emits one warning per referenced bug, including non-security bugs, and each one needs acknowledging. Please collapse them into a single message listing the bug ids.

src/lando/utils/github_checks.py
```diff
@@ -276,6 +286,65 @@ def run(
...
+                continue
+
+            keyword = security_keyword(bug)
+            if keyword is None:
+                continue
+
+            missing_flags = unset_status_flags(bug, prefix)
+            if missing_flags:
+                messages.append(
+                    f"Bug {bug_id} is marked {keyword} but is missing status flags: "
```
cgsheeh: This message is duplicated verbatim from `blocker_security_bug_status_flags` in `transplants.py`, as is the "could not verify" warning text. Please move both into `bmo.py` next to `security_keyword`/`unset_status_flags` so the two flows can't drift apart.

src/lando/api/tests/test_transplants.py
tomrittervg: oops

src/lando/main/migrations/0064_repo_status_flag_prefix.py
```diff
@@ -0,0 +1,24 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("main", "0063_alter_repo_approval_required"),
```
cgsheeh: You'll have to redo the migrations as `0063` is no longer the tip.

src/lando/api/tests/test_transplants.py
```diff
@@ -484,6 +488,236 @@ def test_warning_revision_secure_is_not_secure(
...
+@pytest.mark.django_db
+def test_security_flags_blocker_surfaces_in_assessment(security_flags_state):
+    revision, stack_state = security_flags_state(
+        bugs=[{"id": 123, "keywords": ["sec-high"], "cf_status_firefox130": "---"}],
+    )
+
+    assessment = run_landing_checks(stack_state)
+
+    assert any(
+        "missing Firefox status flags" in blocker for blocker in assessment.blockers
```
cgsheeh: I ran the CI workflows and it seems this test is failing.

src/lando/api/legacy/transplants.py
```diff
@@ -883,6 +940,43 @@ def blocker_try_task_config(
         return issues[0]
 
 
+def blocker_security_bug_status_flags(
```
cgsheeh: Per comment 2 on bug 2055604, sec-high/sec-critical bugs sit in `core-security` until the fix hits central, so Lando's key can't read exactly the bugs this targets — meaning in the common case this blocker returns `None` and we fall through to the warning. Is bug 2055892 expected to land first, or are we knowingly shipping this as warning-only for now?

src/lando/api/legacy/transplants.py
```diff
@@ -1038,6 +1134,41 @@ def get_parsed_diffs(
     }
 
 
+def fetch_bugs_for_stack(
```
cgsheeh, line 1156: This adds a synchronous, uncached BMO request to the stack-page and landing request paths, and `api_request` never sets a `timeout`, so a hung BMO connection can tie up a web worker indefinitely. Could we pass a timeout through, and cache the result the way we do for raw diffs?

tomrittervg: I added a 60 second cache, but I'm wondering if it should be shorter.  If the user gets this error, they're going to go set the flags and try to reland and if the cache isn't expired they'll get the message again erroneously, won't they?

shtrom: Ah, we already have a caching mechanism in the form of a decorator https://github.com/mozilla-conduit/lando/blob/28a385011f250722eba6ce64455fd53d9686f2b8/src/lando/utils/cache.py#L13-L22

Could you use this one instead?

Here's an example cache_key https://github.com/mozilla-conduit/lando/blob/28a385011f250722eba6ce64455fd53d9686f2b8/src/lando/utils/github.py#L435-L436.

It may make sense to extend this decorator to include a TTL so we can pass the one you set for BMO requests there.

cgsheeh (changes requested):
Will look at this again tomorrow, but in the meantime there are some linting test failures that need to be addressed.

src/lando/api/legacy/bmo.py
```diff
@@ -33,6 +106,10 @@ def api_request(
     if use_api_key:
         common_headers["X-Bugzilla-API-Key"] = settings.BUGZILLA_API_KEY
 
+    # Ensure every BMO request has a timeout so a hung connection can't tie up a
+    # web worker indefinitely. Callers may override by passing `timeout`.
```
cgsheeh: This comment should go in the docstring.

src/lando/api/legacy/bmo.py
```diff
@@ -1,8 +1,81 @@
...
+from lando.utils.cache import cache_method
+
+logger = logging.getLogger(__name__)
+
+SECURITY_KEYWORDS = ("sec-critical", "sec-high")
+UNSET_STATUS_FLAG_VALUES = ("", "---")
+
+BMO_REQUEST_TIMEOUT = 10
+
+BMO_BUGS_CACHE_TIMEOUT = 60
```
zzzeid: Probably should put these two in the settings module.

src/lando/api/legacy/bmo.py
```diff
@@ -82,3 +159,52 @@ def uplift_update_bug(json: dict) -> requests.Response:
...
+                status_code,
+            )
+        else:
+            logger.warning("BMO returned an error fetching bug status flags.")
+    except requests.exceptions.RequestException:
+        logger.warning("Failed to reach BMO for bug status flags.")
+    except KeyError, TypeError, ValueError:
+        logger.exception("BMO returned a malformed bug status-flag payload.")
+
+    return None
```
zzzeid: I think returning `None` here could get messy/confusing. It would be better and more explicit to raise an exception (even if a custom one) to make this more obvious (rather than reusing a "falsey" value and having to differentiate between an empty dict and a null value elsewhere).

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -153,3 +154,32 @@ def migrate_3_bug2054804_enable_PreventSignedCommitsCheck(
...
+        self, ask_confirm: bool = True
+    ):
+        """Set `status_flag_prefix` to `cf_status_firefox` on all Firefox repos.
+
+        This enables the security status-flag landing check (bug 2055604), which
+        blocks sec-high/sec-critical bugs whose Firefox status flags are unset.
+        Applies to every Firefox repo, autoland (`firefox`) included.
+        """
+        fx_repos = Repo.objects.filter(
+            Q(name="firefox") | Q(name__startswith="firefox-"),
```
zzzeid: There is no repo in production with the name `firefox`. This should also be more explicit, ideally by listing out all the repos that we intend to apply this to.

src/lando/utils/tests/test_github.py
```diff
@@ -450,6 +450,39 @@ def test_api_client_get_pull_request_commits(
...
+    """`bugs_by_id` caches, so the two status-flag checks share a single BMO fetch."""
+    with (
+        mock.patch.object(
+            PullRequest, "bug_ids", new_callable=mock.PropertyMock, return_value={123}
+        ),
+        mock.patch(
+            "lando.utils.github.fetch_bugs", return_value={123: {"id": 123}}
+        ) as fetch_bugs,
+    ):
+        pr = PullRequest.__new__(PullRequest)
```
zzzeid: Is there a reason we are using `__new__` here?

src/lando/utils/cache.py
zzzeid, line 1: A bug should be filed for this and linked to in this PR. Also should be called out in the commit message.

src/lando/utils/cache.py
```diff
@@ -13,12 +14,16 @@
 def cache_method(
     key_fn: Callable[..., str],
     cache_alias: str = "default",
+    timeout: int | None = DEFAULT_TIMEOUT,
```
zzzeid: We should simply set `timeout` to `None` here, and pass it to `cache.set` only if it is provided. Importing `DEFAULT_TIMEOUT` here will cause unexpected results if we modify the cache entries in `lando.settings` (which can differ based on which cache we are using anyway).

src/lando/utils/tests/test_github_checks.py
```diff
@@ -0,0 +1,160 @@
...
+    """Build a stand-in PullRequest.
+
+    `.bug_ids` is parsed from the commit messages; `.bugs_by_id` is set directly to
+    `bugs_by_id` (a dict, or `None` to simulate a failed BMO fetch), mirroring the
+    cached property both checks now read.
+    """
+    bug_ids: set[int] = set()
+    for message in commit_messages:
+        bug_ids.update(parse_bugs(message))
+    return SimpleNamespace(bug_ids=bug_ids, bugs_by_id=bugs_by_id)
```
shtrom: What's the motivation for usint `SimpleNamespace` here? Could we use `unittest.mock.Mock`s instead?

src/lando/utils/tests/test_github_checks.py
```diff
@@ -0,0 +1,160 @@
...
+    return PullRequestSecurityBugStatusFlagsBlocker.run(pull_request, target_repo, None)
+
+
+def _run_warning(pull_request, target_repo):
+    return PullRequestSecurityBugStatusFlagsUnverifiedWarning.run(
+        pull_request, target_repo, None
+    )
+
+
+def test_blocker_sec_high_unset_blocks():
```
shtrom: Could we rename those functions to something like `test_sec_flag_blocker...`?

This makes me realise that we don't have any direct test coverage for the GitHub checks, but we do need to add this, so `blocker` is a bit vague on its own.

src/lando/utils/tests/test_github_checks.py
```diff
@@ -0,0 +1,160 @@
+from types import SimpleNamespace
+
+from lando.api.legacy.commit_message import parse_bugs
+from lando.utils.github_checks import (
+    PullRequestSecurityBugStatusFlagsBlocker,
+    PullRequestSecurityBugStatusFlagsUnverifiedWarning,
+)
+
+
+def _pull_request(bugs_by_id, *commit_messages: str) -> SimpleNamespace:
```
shtrom: Could this be made into a `Callable` fixture with the `commit_messages` as the first argument? (also, `commit_messages` looks like it's a `list[str]`)

src/lando/utils/tests/test_github_checks.py
```diff
@@ -0,0 +1,160 @@
...
+    touching the PR — in particular before triggering the `bugs_by_id` BMO fetch.
+    """
+
+    def __getattr__(self, name):
+        raise AssertionError(f"unexpected access to pull_request.{name}")
+
+
+def _repo(status_flag_prefix: str = "cf_status_firefox") -> SimpleNamespace:
+    """Build a stand-in target repo exposing only `.status_flag_prefix`."""
+    return SimpleNamespace(status_flag_prefix=status_flag_prefix)
```
shtrom: We'd be better off adding support for the status flag prefix in the `repo_mc` fixture https://github.com/mozilla-conduit/lando/blob/662ea3d292973f38864e39f12a5a4950a3767ac5/src/lando/conftest.py#L905-L929, and using that.

shtrom (commented):
Some notes on the github_checks tests, to fit them better with our testsuite.

cgsheeh (approved):
LGTM, once comments from the other reviewers are addressed.

src/lando/utils/management/commands/migrate_data.py
```diff
@@ -153,3 +153,43 @@ def migrate_3_bug2054804_enable_PreventSignedCommitsCheck(
                 self.stdout.write(f"{repo.name} ", ending="")
 
         self.stdout.write("done.")
+
+    def migrate_4_bug2055604_set_firefox_status_flag_prefix(
```
shtrom, line 157: We landed another migration in the meantime. This now needs to be 5.
```suggestion
    def migrate_5_bug2055604_set_firefox_status_flag_prefix(
```

src/lando/utils/cache.py
shtrom: -> https://bugzilla.mozilla.org/show_bug.cgi?id=2064533

tomrittervg:
Okay, sorry it took me son long to get back to this, but updated and rebased

src/lando/api/legacy/bmo.py
```diff
@@ -82,3 +165,55 @@ def uplift_update_bug(json: dict) -> requests.Response:
...
+    params = {"id": ",".join(str(bug) for bug in sorted(bug_ids))}
+    try:
+        bugs = uplift_get_bug(params)["bugs"]
+        return {int(bug["id"]): bug for bug in bugs}
+    except requests.exceptions.HTTPError as exc:
+        status_code = exc.response.status_code if exc.response is not None else None
+        if status_code is not None and 400 <= status_code < 500:
+            logger.warning(
+                "BMO rejected the Lando status-flag request with HTTP %s; check "
+                "BUGZILLA_API_KEY. Security status-flag checks will degrade to a "
```
zzzeid: In theory a 400 could result due to various other causes, not just the Bugzilla API key. Might be better to keep the error generic. Other causes could be, for example, rate limiting, malformed request, etc... Also this warning will be quite hard to spot since we're not logging it in Sentry (seems like all these errors are later being caught and skipped).

nit: can just use an f-string here.

src/lando/api/legacy/transplants.py
```diff
@@ -1056,6 +1153,45 @@ def get_parsed_diffs(
...
+    stack_data: RevisionData, supported_repos: dict[str, Repo]
+) -> dict[int, dict] | None:
+    """Fetch BMO bug data for every bug referenced by the stack, keyed by bug id.
+
+    Returns the mapping (empty if the stack references no bugs), or `None` if the
+    fetch failed (`bmo.fetch_bugs` raised `BugFetchError`), in which case the
```
zzzeid: Similar to my comment [here](https://github.com/mozilla-conduit/lando/pull/1352#discussion_r3785376922), I think repurposing a "falsey" value to determine business logic is not ideal / is messy.

src/lando/utils/github.py
```diff
@@ -609,6 +611,21 @@ def commits(self) -> list[dict]:
 
         return commits
 
+    @property
+    def bug_ids(self) -> set[int]:
```
zzzeid: Would be good to have a docstring here.

zzzeid (approved):
Looking pretty good. The only "blocker" is regarding the warnings being logged. If the intention is to expose these to the developers so that they become aware of an underlying issue then we should be logging these to sentry, since our logs are quite noisy and are usually used for digging into issues that are exposed via Sentry.

I also think the logic around setting bugs to `None` to identify that there was an error is not intuitive and can be improved, but won't block on that.

shtrom:
Grah, I introduced conflicts by merging my GitHub refactor. Will fix.

PR #1353, opened 2026-07-28
dependabot: add required `dependency-name` to the npm ignore rule (bug 2058371)

cgsheeh:
Each entry under `ignore` must specify `dependency-name`, so the config has
failed validation since the rule was added. Use `"*"` to keep applying the
major-version ignore to every npm dependency.

---
Lando: [link](https://lando.moz.tools/pulls/lando/1353/)
Bugzilla: [bug 2058371](https://bugzilla.mozilla.org/show_bug.cgi?id=2058371)

:warning: This pull request has 1 warning.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,7 +23,8 @@ updates:
       allow:
           - dependency-type: all
       ignore:
-          - update-types:
-                # Our package.json allows semver-compatible versions (with ^).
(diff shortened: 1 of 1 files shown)


cgsheeh:
I think we'll have to land this to test it, but if there's another way that would be good to know. :)

shtrom (approved):
lgtm

PR #1354, opened 2026-07-28
workers: abort jobs which keep deferring (bug 2021999)

cgsheeh:
A job which repeatedly hits a temporary failure blocks every other job for its
repository, so it now moves to the new `ABORTED` terminal status once it runs
out of attempts, and the requester is emailed the reason. Deferrals expected to
resolve without anyone looking at the job, such as a closed tree, are exempt and
give their attempt back so those jobs are retried indefinitely. The number of
allowed attempts defaults to ten and is configurable via `MAX_JOB_ATTEMPTS`.

---

Lando: [link](https://lando.moz.tools/pulls/lando/1354/)
Bugzilla: [bug 2021999](https://bugzilla.mozilla.org/show_bug.cgi?id=2021999)

:warning: This pull request has 13 warnings.


src/lando/api/tests/test_notifications.py
```diff
@@ -53,6 +53,42 @@ def test_email_content_try():
...
+
+def test_email_content_job_aborted():
+    email = make_job_aborted_email(
+        "sadpanda@failure.test",
+        "Landing",
+        123,
+        "https://lando.test/landings/123/",
+        "Push failed!",
+    )
+    assert email.to == ["sadpanda@failure.test"]
```
shtrom, line 82: 🐼

src/lando/main/models/jobs.py
```diff
@@ -109,9 +139,17 @@ class Meta:
     # To be overridden by subclasses.
     type: str = "undefined"
 
+    # Name of the URL pattern which displays the details of this type of job.
+    # To be overridden by subclasses.
+    url_name: str = ""
```
shtrom: I wonder if we could use this to DRY up the `JobAdmin.view_on_site` https://github.com/mozilla-conduit/lando/blob/main/src/lando/main/admin.py#L141-L143

src/lando/utils/tasks.py
```diff
@@ -115,6 +116,45 @@ def send_bug_update_failure_email(
...
+
+    Args:
+        recipient_email: The email of the user receiving the notification.
+        job_type: Human-friendly name of the type of job, e.g. "Landing".
+        job_id: The ID of the aborted job.
+        job_url: URL of the job details page.
+        error: The error recorded on the job when it was aborted.
+    """
+    if not recipient_email:
+        logger.info(f"Skipping aborted email for job {job_id} without a requester.")
```
shtrom: I think this should be a warning.

src/lando/utils/tasks.py
```diff
@@ -115,6 +116,45 @@ def send_bug_update_failure_email(
...
+                    recipient_email,
+                    job_type,
+                    job_id,
+                    job_url,
+                    error,
+                )
+            ]
+        )
+
+    logger.info(f"Aborted job notification email sent to {recipient_email}")
```
shtrom: Can we include the job type and ID in here?

shtrom (approved):
lgtm with a couple of nits

src/lando/api/tests/test_landings.py
```diff
@@ -708,6 +715,96 @@ def test_integrated_execute_job_with_scm_internal_error(
...
+        "push",
+        mock.MagicMock(side_effect=SCMInternalServerError("push failed", "500")),
+    )
+
+    worker = get_landing_worker(SCMType.GIT)
+    worker.worker_instance.applicable_repos.add(repo)
+    worker.refresh_active_repos()
+
+    for _attempt in range(2):
+        worker.loop()
```
zzzeid: It would be better to use `worker.start(max_loops=2)` here.

Could also check that the first loop did not cause the abort.

src/lando/api/tests/test_landings.py
```diff
@@ -708,6 +715,96 @@ def test_integrated_execute_job_with_scm_internal_error(
...
+        target_repo=repo,
+    )
+
+    worker = get_landing_worker(SCMType.GIT)
+    for _attempt in range(DEFAULT_MAX_JOB_ATTEMPTS + 1):
+        # Stop early on an abort, so the next run cannot defer the job back.
+        if job.status in JobStatus.final():
+            break
+        job.start_attempt()
+        worker.run_job(job)
```
zzzeid, line 800: nit: This does follow the convention (unlike above) though it would also be good to use `worker.start(max_loops=1)` here, since that would go through the full worker flow.

src/lando/api/tests/test_views.py
```diff
@@ -643,3 +645,43 @@ def test__views_landing_job_pull_request_view__warnings(
...
+    job = LandingJob.objects.create(
+        target_repo=repo, status=job_status, is_pull_request_job=True
+    )
+    add_revisions_to_job([Revision.objects.create(pull_number=pull_number)], job)
+
+    response = client.get(f"/api/pulls/{repo.name}/{pull_number}/landing_jobs")
+
+    assert response.json() == {"status": expected_status}, (
+        f"A `{job_status}` job should be reported as `{expected_status}`."
+    )
```
zzzeid, line 704: Would be good to add something in the commit message to describe this addition. It also seems to be quite generic, so not sure we need the reference to the bug on line 669.

src/lando/main/models/jobs.py
zzzeid: I think this is more accurately named `view_name` (or indeed Django uses `viewname` but I think having the underscore is better).

src/lando/main/models/jobs.py
```diff
@@ -202,8 +254,72 @@ def transition_status(
...
+        # A single attempt may defer through more than one code path, so only its
+        # first deferral is taken into account.
+        if action == JobAction.DEFER and not was_deferred:
+            self.handle_deferral(
+                kwargs["message"], abortable=kwargs.get("abortable", True)
+            )
+
         self.save()
 
+    def handle_deferral(self, message: str, abortable: bool):
```
zzzeid: I'm not sure this new method to handle deferrals and aborts in one place is needed. It would be simpler to decouple these things -- i.e., check the max number of attempts before deciding to defer or abort, and then perform the appropriate action. Was there a specific reason/benefit to tying these together in this way? This should be quite simple to trigger in the couple of spots where it's needed.

src/lando/main/models/jobs.py
```diff
@@ -202,8 +254,72 @@ def transition_status(
...
+            max_attempts = 0
+
+        if max_attempts < 1:
+            logger.error(
+                "`MAX_JOB_ATTEMPTS` is not a positive integer, falling back to "
+                f"{DEFAULT_MAX_JOB_ATTEMPTS}."
+            )
+            return DEFAULT_MAX_JOB_ATTEMPTS
+
+        return max_attempts
```
zzzeid: This seems unnecessary / too low level. If the configuration variable is set to be an integer, it should always return an integer (including whatever default we specify in the code). If it doesn't, this is an issue with ConfigurationVariable and not with this specific usage of it.

zzzeid (changes requested):
Looks good, but main question/issue is around intertwining the deferred/abort concepts in `handle_deferral`, which I think could be simplified/decoupled.

src/lando/main/models/jobs.py
cgsheeh: Good call, done. `BaseJob.path` returns the `reverse`d path and `url` joins it onto
`SITE_URL`, so `JobAdmin.view_on_site` is now just `return instance.path()`. That also
fixes a latent bug: `UpliftJobAdmin` and `AutomationJobAdmin` inherit `view_on_site`,
so "view on site" pointed at `jobs-page` for uplift and automation jobs. The override
in `AutomationJobAdmin` is redundant now and is gone too.

src/lando/utils/tasks.py
cgsheeh: Done.

src/lando/utils/tasks.py
cgsheeh: Done, in both messages.

src/lando/api/tests/test_landings.py
cgsheeh: Done.

src/lando/api/tests/test_landings.py
cgsheeh: The worker loop can't reach this one, unfortunately: `refresh_active_repos` drops any
repo whose tree is closed, so the worker never picks the job up and the closed-tree
deferral inside `run_job` never runs. I've left the
direct `run_job` calls with a comment explaining why.

To get the coverage you're after I added `test_non_abortable_failures_do_not_abort_job`,
which exercises the same "attempt is given back, job is never aborted" behaviour through
`worker.start(max_loops=...)`, by raising `TreeApprovalRequired` (the other
non-abortable failure) from `push` with the tree open.

src/lando/api/tests/test_views.py
cgsheeh: Dropped the bug reference from the docstring, and I'll call out
the added pull request status coverage in the commit message.

src/lando/main/models/jobs.py
cgsheeh: Agreed, renamed to `view_name`.

src/lando/main/models/jobs.py
cgsheeh: A single attempt can defer through more than one code path: `Worker.update_repo` and
`LandingWorker.push_changes` both defer and then raise `TemporaryFailureException`, which
`Worker.loop` catches and defers again. `transition_status` is the only place that can
tell those apart, via `was_deferred`, so checking at each call site would count one
attempt twice and abort at half the configured limit. `start_attempt` already owns
`attempts`, so giving one back belongs next to it. Happy to rename this to
`count_deferral` if the name is what's obscuring the intent.

src/lando/main/models/jobs.py
cgsheeh: Fair enough, dropped it. `max_attempts` just reads the variable now and falls back to
`DEFAULT_MAX_JOB_ATTEMPTS` when it is unset.

src/lando/main/models/jobs.py
zzzeid: >transition_status is the only place that can tell those apart, via was_deferred

We should be able to tell whether a job is already deferred just before we call `transition_status` though.

> which Worker.loop catches and defers again

This sounds like a bug, though it sounds like it should not affect the attempt count. Looking at the code, it looks like it's a no-op (minus overriding the message) when the job was already deferred, but then gets deferred again in `loop` when `TemporaryFailureException` is caught. We should be able to detect that the job is already deferred and do nothing in that case, or better yet, get rid of the deferral just before the exception is raised, since that is already handled in `loop` (this could just be a legacy artifact, since the first deferral here is basically being ignored).

> checking at each call site would count one attempt twice and abort at half the configured limit

Determining abortability should be based on attempt count, not on how many times `transition_status` is called. But again if this is happening then this is a bug we should fix (i.e., don't defer twice in a row in one loop).


I'm suggesting adding something like this:

in jobs.py:
```python
actions = {
    ...
    JobAction.DEFER: {
        "required_params": ["message"],
        "status": JobStatus.DEFERRED,
    },
    JobAction.ABORT: {
        "required_params": ["message"],
        "status": JobStatus.ABORTED,
    },
    ...
}
```

In landing_worker.py, you could do something like this:
```python
...
# in run_job (no change)
if not self.treestatus_client.is_open(repo.tree):
    job.transition_status(
        JobAction.DEFER,
        message=f"Tree {repo.tree} is closed - retrying later.",
    )
...

# in apply_and_push
...
message = (
    f"`Temporary error ({e.__class__}) "
    f"encountered while pushing to {repo_push_info}: {e}"
)
logger.exception(message)

# pass both the exception and the job status to determine abortability.
should_abort = self.should_abort(e, job)
if should_abort:
    # abort job (either by passing transition action to `PermanentFailureException` or otherwise
else:
    # defer job (either by passing transition action to `TemporaryFailureException` or otherwise
```

in workers/base.py, you would just adapt the check for abortability to simply abort, or defer, based on the exception.

I think this separation would simplify the logic, and fix some of the bugs along the way rather than implementing this on top of the existing bugs.

shtrom:
Half-baked thought: I'm now wondering if the logic could be simplified by handling deferrals/abortions at the beginning of the loop, rather than in exception handling.

This means we'd only need the abort-decision logic to be called from a single spot, when next picking up the job for processing.

src/lando/main/models/jobs.py
cgsheeh: Makes sense, I gave it a shot and it seems like a nice cleanup. :+1:

src/lando/api/legacy/workers/landing_worker.py
zzzeid, line 327: Could simplify this by doing
```python
except NON_ABORTABLE_FAILURES:
    job.transition_status(
        JobAction.DEFER,
        message=message,
    )
```

i.e., only `defer_or_abort` if an exception is abortable.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -126,6 +126,8 @@ def run_job(self, job: LandingJob) -> bool:
             job.transition_status(
                 JobAction.DEFER,
                 message=f"Tree {repo.tree} is closed - retrying later.",
+                # The tree will reopen without anyone looking at this job.
```
zzzeid: The meaning of this comment is unclear.

src/lando/main/models/jobs.py
```diff
@@ -189,21 +247,53 @@ def transition_status(
...
         if action == JobAction.LAND:
             self.landed_commit_id = kwargs["commit_id"]
 
+        # A deferral which is expected to resolve on its own, such as a closed tree,
+        # gives its attempt back so the job is retried indefinitely.
+        if action == JobAction.DEFER and not kwargs.get("abortable", True):
+            logger.debug(
+                f"Deferral of {self} is not abortable, giving the attempt back."
+            )
+            self.attempts = max(0, self.attempts - 1)
```
zzzeid: This interferes with the existing convention of counting attempts (i.e., this number should always be going up, and reflect the number of iterations that this job has gone through before succeeding). It also adds complexity since `abortable` is only passed and used for this purpose.

It would be simpler to continue counting attempts as was done before. The edge case that this is trying to address is when a number of non-abortable exceptions are encountered, followed by an abortable exception, in which case the job would get aborted after however many attempts are remaining (at least 1). I think this is an acceptable scenario to leave as-is without changing the meaning of `job.attempts`.

src/lando/main/models/jobs.py
```diff
@@ -4,22 +4,48 @@
...
     """Signal an error that should be retried"""
 
+    def __init__(self, *args, abortable: bool = True):
+        """Set `abortable` to `False` to retry the job indefinitely.
+
+        Non-abortable failures are those which are expected to resolve on their own
+        without anyone looking at the job, such as a closed tree.
+        """
+        super().__init__(*args)
+        self.abortable = abortable
```
zzzeid: I don't think this is being used anywhere.

src/lando/api/legacy/workers/base.py
```diff
@@ -267,6 +272,48 @@ def loop(self):
                     extra={"id": job.id},
                 )
 
+            if job.status == JobStatus.ABORTED:
+                # The job deferred too many times and was given up on. It is in a
+                # final state, so the worker can move on to the next job.
+                self.last_job_finished = True
```
zzzeid: This variable is being set twice here. We should set this to the correct value on line 243.

src/lando/api/legacy/workers/base.py
```diff
@@ -267,6 +272,48 @@ def loop(self):
...
+    def should_abort(self, exception: Exception, job: BaseJob) -> bool:
+        """Whether `exception` should abort `job` rather than defer it again.
+
+        A job which keeps deferring blocks every other job for its repository, so it
+        is aborted once it hits an abortable failure with no attempts left.
+        """
+        return self.is_abortable_failure(exception) and not job.has_attempts_remaining()
+
+    def defer_or_abort(self, job: BaseJob, exception: Exception, message: str):
+        """Abort `job` if it has run out of attempts, otherwise defer it."""
```
zzzeid: docstring is not very accurate here. 
```suggestion
        """Abort `job` if it encounterd an abortable exception on the final attempt, otherwise defer it."""
```

src/lando/api/legacy/workers/automation_worker.py
zzzeid, line 159: Could simplify this by doing
```python
except NON_ABORTABLE_FAILURES:
    job.transition_status(
        JobAction.DEFER,
        message=message,
    )
```

And catching remaining exceptions below.

src/lando/api/legacy/workers/base.py
```diff
@@ -267,6 +272,48 @@ def loop(self):
...
+            if job.status == JobStatus.ABORTED:
+                # The job deferred too many times and was given up on. It is in a
+                # final state, so the worker can move on to the next job.
+                self.last_job_finished = True
+                self.notify_user_of_job_abort(job)
+
+    @staticmethod
+    def is_abortable_failure(exception: Exception) -> bool:
+        """Whether a job deferred because of `exception` may eventually be aborted."""
+        return not isinstance(exception, NON_ABORTABLE_FAILURES)
```
zzzeid: Technically this can be determined in the loops/areas where the exception is raised, which I think is a hybrid/similar idea to what @shtrom said [here](https://github.com/mozilla-conduit/lando/pull/1354#issuecomment-5311161098). It would remove the need for having this as part of the `should_abort` since only abortable exceptions would be handled by `defer_or_abort`. Left comments in other areas to illustrate what I mean here.

src/lando/api/legacy/workers/landing_worker.py
```diff
@@ -357,8 +359,10 @@ def rebase_stack(revision: Revision):
                 f"encountered while pushing to {repo_push_info}: {e}"
             )
             logger.exception(message)
-            job.transition_status(JobAction.DEFER, message=message)
-            raise TemporaryFailureException(message)
+            self.defer_or_abort(job, e, message=message)
+            raise TemporaryFailureException(
+                message, abortable=self.is_abortable_failure(e)
+            )
```
zzzeid: This is still deferring twice (once here, and then again in `loop` where `TemporaryFailureException` is caught.

src/lando/main/models/jobs.py
```diff
@@ -128,7 +173,8 @@ def __str__(self) -> str:
     # LDAP email of the user who created the job.
     requester_email = models.CharField(default="", max_length=255)
 
-    # Number of attempts made to complete the job.
+    # Number of attempts made to complete the job. Attempts which end in a deferral
+    # that is expected to resolve on its own, such as a closed tree, are given back.
```
zzzeid: I don't think the meaning of attempts should be changed here (see my other comments). If we run into issues with this (i.e., needing to give back attempt counts) we could think about implementing a different mechanism. This also opens up another avenue for things to go wrong (i.e., unexpected flow that results in the number of attempts never hitting the max.)

src/lando/main/models/jobs.py
```diff
@@ -109,9 +142,21 @@ class Meta:
     # To be overridden by subclasses.
     type: str = "undefined"
 
+    # Name of the view which displays the details of this type of job.
+    # To be overridden by subclasses.
+    view_name: str = ""
```
zzzeid: The variable name here is a too broad. A view related to this model can exist in multiple apps or modules (e.g., one or more API views, or a page in the UI, etc... Maybe there's a better name or way to specify this.

zzzeid (changes requested):
Looking better! A few comments/questions. I will look at the tests in the final pass since some of the questions I left are implementation related.

src/lando/main/tests/test_jobs.py
zzzeid, line 1: We're bifurcating these tests a little here, given that we have adjacent tests in `lando.api.tests.test_landing_job`.

src/lando/main/tests/test_jobs.py
```diff
@@ -0,0 +1,107 @@
...
+        (AutomationJob, "/api/jobs/1/"),
+        (UpliftJob, "/uplift/jobs/1/"),
+    ),
+)
+def test__models__BaseJob__url(job_class: type, expected_path: str):
+    job = job_class(id=1, status=JobStatus.SUBMITTED)
+    assert job.url() == f"https://lando.test{expected_path}", (
+        f"`url` should point at the details page of the {job.type} job."
+    )
+
```
zzzeid, line 36: nit: Would be more readable to move this to the top of the module, this could also be a fixture.

src/lando/main/tests/test_jobs.py
```diff
@@ -0,0 +1,107 @@
...
+    attempt_and_defer(job, DEFAULT_MAX_JOB_ATTEMPTS - 1)
+    assert job.has_attempts_remaining(), (
+        "A job below the maximum number of attempts should have attempts remaining."
+    )
+
+    attempt_and_defer(job, 1)
+    assert not job.has_attempts_remaining(), (
+        "A job at the maximum number of attempts should have no attempts remaining."
+    )
+
```
zzzeid: This test should be a unit test without needing any db access or landing jobs that get attempted `DEFAULT_MAX_JOB_ATTEMPTS` times (would make this much faster). It would achieve the same functionality since we are not testing any integration aspects here (e.g., job status, or what is happening during an attempts.).

src/lando/main/tests/test_jobs.py
```diff
@@ -0,0 +1,107 @@
...
+    job.transition_status(JobAction.ABORT, message="the last failure")
+
+    assert job.status == JobStatus.ABORTED, "`ABORT` should abort the job."
+    assert "Lando gave up on this job" in job.error, (
+        "Error message should explain that the job was aborted."
+    )
+    assert "the last failure" in job.error, (
+        "Error message should include the reason for the last failure."
+    )
```
zzzeid: Could this be more explicit? Would ensure that no unexpected text is in the error message.

src/lando/main/tests/test_jobs.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+@pytest.mark.django_db
+def test__models__BaseJob__transition_status_rejects_unknown_params(
+    make_landing_job: Callable,
+):
+    job = make_landing_job(status=JobStatus.SUBMITTED)
+
+    with pytest.raises(ValueError):
+        job.transition_status(JobAction.DEFER, message="failure", bogus=True)
```
zzzeid: Not clear why this was added in the PR (i.e., unknown parameters).

src/lando/main/tests/test_jobs.py
```diff
@@ -0,0 +1,107 @@
...
+
+    assert job.has_attempts_remaining(), (
+        "The job should have attempts remaining before its last allowed attempt."
+    )
+
+    attempt_and_defer(job, 1)
+
+    assert not job.has_attempts_remaining(), (
+        "The job should have no attempts remaining after its last allowed attempt."
+    )
```
zzzeid: This is re-testing some of the aspects tested in `test__models__BaseJob__has_attempts_remaining`. Ok to do that as long as the other test is made to be a unit test though. Otherwise that one is redundant.

src/lando/main/tests/test_jobs.py
```diff
@@ -0,0 +1,107 @@
...
+    assert job.status == JobStatus.ABORTED, "`ABORT` should abort the job."
+    assert "Lando gave up on this job" in job.error, (
+        "Error message should explain that the job was aborted."
+    )
+    assert "the last failure" in job.error, (
+        "Error message should include the reason for the last failure."
+    )
+
+
+@pytest.mark.parametrize("max_attempts", (1, 2, 5))
```
zzzeid: What's the point of having `2` and `5` in there? Seems like they are testing the same things, with `5` being slower and less efficient.

Though if we wanted to test edge cases, `0` and `-1` could be good candidates, but can optionally add validation. We're unlikely to be modifying this value much though (and we can be sensible about which value we choose here).

src/lando/api/tests/test_landings.py
```diff
@@ -694,6 +701,150 @@ def test_integrated_execute_job_with_scm_internal_error(
...
+    )
+    assert mail.outbox[0].to == ["test@example.com"], (
+        "The abort notification should be sent to the requester."
+    )
+    assert mail.outbox[0].subject == f"Lando: Landing job {job.id} was aborted!", (
+        "The abort notification should identify the aborted job."
+    )
+    assert "push failed" in job.error, (
+        "The job error should include the reason for the last failure."
+    )
```
zzzeid: Can this be more explicit?

src/lando/api/tests/test_landings.py
```diff
@@ -694,6 +701,150 @@ def test_integrated_execute_job_with_scm_internal_error(
...
 
+@pytest.mark.django_db
+def test_aborted_job_notifies_requester(
+    monkeypatch: pytest.MonkeyPatch,
+    repo_mc: Callable,
+    create_patch_revision: Callable,
+    make_landing_job: Callable,
+    get_landing_worker: Callable,
+):
+    """Bug 2021999: the requester is emailed when Lando gives up on their job."""
```
zzzeid: nit: Should the bug number be here? We don't really do that anywhere else.

src/lando/api/tests/test_landings.py
zzzeid: This is basically a symptom of the edge case of when a job is picked up just before a tree is closed I think. It would be harder but not impossible to test that. However, this is acceptable as well.

src/lando/api/tests/test_landings.py
```diff
@@ -694,6 +701,150 @@ def test_integrated_execute_job_with_scm_internal_error(
...
+    repo = repo_mc(SCMType.GIT)
+
+    job = make_landing_job(
+        revisions=[create_patch_revision(1)],
+        status=JobStatus.SUBMITTED,
+        requester_email="test@example.com",
+        target_repo=repo,
+    )
+
+    # The worker loop fetches its own `Repo` instance, so mock the SCM class.
```
zzzeid: nit: Would be more useful here to state what this mock is doing (i.e., "simulate a known temporary failure" or something like that.) Same with the other instance of this comment above.

src/lando/main/models/jobs.py
```diff
@@ -189,21 +236,44 @@ def transition_status(
...
-        if sorted(required_params) != sorted(kwargs.keys()):
-            missing_params = required_params - kwargs.keys()
+        required_params = set(actions[action]["required_params"])
+
+        if missing_params := required_params - kwargs.keys():
             raise ValueError(f"Missing {missing_params} params")
 
+        if unknown_params := kwargs.keys() - required_params:
+            raise ValueError(f"Unknown {unknown_params} params")
+
```
zzzeid: Not clear why this was added in the PR. I assume just to tighten up the usage options?

zzzeid (commented):
Looking good, though there are a couple of comments re: test efficiency that I think should be fixed up (along with any nits you want to fix) before this lands.

src/lando/main/tests/test_jobs.py
cgsheeh: Removed the changes to `job.transition_status` params passing, and this test.

src/lando/main/tests/test_jobs.py
cgsheeh: Switched to keying off the monkeypatched `MAX_ATTEMPTS` instead.

src/lando/main/tests/test_jobs.py
cgsheeh: Updated to checking the exact expected error message.

src/lando/main/tests/test_jobs.py
cgsheeh: Mocked out the `max_attempts` field so we don't need to access the DB's `ConfigurationVariable`.

src/lando/api/tests/test_landings.py
cgsheeh: Removed bug numbers from docstrings.

src/lando/main/models/jobs.py
cgsheeh: Yes it was, reset it back to the previous behaviour and removed the test.

zzzeid (approved):

PR #1404, opened 2026-07-30
views: fix key error with empty warnings

zzzeid:
WIP DO NOT LAND

---

Lando: [link](https://lando.moz.tools/pulls/lando/1404/)

:no_entry_sign: This pull request has 4 blockers.

### DIFF
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -288,8 +288,9 @@ class LandingJobPullRequestAPIView(PullRequestAPIView):
             def clean(self):
 
                 cleaned_data = self.cleaned_data
-                new_warnings = cleaned_data["new_warnings"]
-                old_warnings = cleaned_data["old_warnings"]
+                new_warnings = cleaned_data.get("new_warnings") or []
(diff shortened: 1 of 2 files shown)


src/lando/api/views.py
```diff
@@ -288,8 +288,9 @@ class Form(forms.Form):
             def clean(self):
 
                 cleaned_data = self.cleaned_data
-                new_warnings = cleaned_data["new_warnings"]
-                old_warnings = cleaned_data["old_warnings"]
+                new_warnings = cleaned_data.get("new_warnings") or []
+                old_warnings = cleaned_data.get("old_warnings") or []
```
zzzeid, line 292: nit: These fields should be present in `cleaned_data`, if they are defined in the form. So I think the `.get` here is redundant.

zzzeid:
Closing this PR - @Magnolia-Liu if you want to open a new one with you as the author and we can land that one.

src/lando/api/views.py
zzzeid: Sorry to correct this, if they are defined and valid*.

PR #1405, opened 2026-07-30
views: fix key error with empty warnings  (bug 2059434)

Magnolia-Liu:
---

Lando: [link](https://lando.moz.tools/pulls/lando/1405/)
Bugzilla: [bug 2059434](https://bugzilla.mozilla.org/show_bug.cgi?id=2059434)

### DIFF
--- a/src/lando/api/views.py
+++ b/src/lando/api/views.py
@@ -288,8 +288,9 @@ class LandingJobPullRequestAPIView(PullRequestAPIView):
             def clean(self):
 
                 cleaned_data = self.cleaned_data
-                new_warnings = cleaned_data["new_warnings"]
-                old_warnings = cleaned_data["old_warnings"]
+                new_warnings = cleaned_data["new_warnings"] or []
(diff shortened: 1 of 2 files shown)


src/lando/api/views.py
```diff
@@ -288,8 +288,9 @@ class Form(forms.Form):
             def clean(self):
 
                 cleaned_data = self.cleaned_data
-                new_warnings = cleaned_data["new_warnings"]
-                old_warnings = cleaned_data["old_warnings"]
+                new_warnings = cleaned_data["new_warnings"] or []
+                old_warnings = cleaned_data["old_warnings"] or []
```
zzzeid, line 292: Just to follow up on [this comment](https://github.com/mozilla-conduit/lando/pull/1404#discussion_r3684338889), if new_warnings (or old_warnings) are not validated correctly, these keys would be missing. However, the previous code would have treated them as valid and empty, so at least in this case we get rid of the false positive.

zzzeid (approved):

PR #1406, opened 2026-07-30
revisions: add author override functionality (bug 2055953)

zzzeid:
- add mailbox fields in legacy transplants
- add validation for mailbox in legacy endpoint and form
- add UI elements
- modify hackbot blocker to be a warning when disallowed author is detected
- pre-fill form with name and email of Lando user, if available
- pre-fill form with name from Phabricator revision author if Lando user is not available
- move some fixtures
- add various tests
- fix issue with landing_path cleaning (bug 2060918)

---

Lando: [link](https://lando.moz.tools/pulls/lando/1406/)
Bugzilla: [bug 2055953](https://bugzilla.mozilla.org/show_bug.cgi?id=2055953)

:warning: This pull request has 9 warnings.

### DIFF
--- a/src/lando/api/legacy/api/transplants.py
+++ b/src/lando/api/legacy/api/transplants.py
@@ -228,8 +228,9 @@ def post(phab: PhabricatorClient, user: User, data: dict) -> tuple[dict[str, int
         # TODO: it would be more ideal if this is a validation error, however, we do not have the
         # necessary information at the time of form submission to determine this without performing
         # additional redundant processing outside the form.
-        raise ValueError(
-            "Mailbox values should be present if and only if disallowed authors are persent."
+        raise LegacyAPIException(
+            400,
+            "Mailbox values should be present if and only if disallowed authors are present.",
         )
 
     if assessment.warnings:
@@ -311,13 +312,14 @@ def post(phab: PhabricatorClient, user: User, data: dict) -> tuple[dict[str, int
             flags,
         )[1]
 
-        args = (phab, revision["id"], diff, commit_message)
-        kwargs = {}
-
         if diff_has_disallowed_author(diff):
-            kwargs["mailbox"] = mailbox
-
-        lando_revision = fetch_raw_diff_and_save(*args, **kwargs)
+            lando_revision = fetch_raw_diff_and_save(
+                phab, revision["id"], diff, commit_message, mailbox=mailbox
+            )
+        else:
+            lando_revision = fetch_raw_diff_and_save(
+                phab, revision["id"], diff, commit_message
+            )
 
         revision_reviewers[lando_revision.id] = get_approved_by_ids(
             phab,
--- a/src/lando/api/legacy/api/transplants.py
+++ b/src/lando/api/legacy/api/transplants.py
@@ -53,7 +53,7 @@ from lando.main.models import (
     Repo,
     add_revisions_to_job,
 )
-from lando.main.support import LegacyAPIException
+from lando.main.support import LegacyAPIException, diff_has_disallowed_author
 from lando.utils.phabricator import PhabricatorClient
 from lando.utils.tasks import admin_remove_phab_project
 
@@ -218,6 +218,20 @@ def post(phab: PhabricatorClient, user: User, data: dict) -> tuple[dict[str, int
         )
         raise LegacyAPIException(400, error_message)
 
+    mailbox = data.get("mailbox", None)
+    has_disallowed_author_diffs = any(
+        diff_has_disallowed_author(diff) for revision, diff in to_land
+    )
+    if (has_disallowed_author_diffs and not mailbox) or (
+        not has_disallowed_author_diffs and mailbox
+    ):
+        # TODO: it would be more ideal if this is a validation error, however, we do not have the
+        # necessary information at the time of form submission to determine this without performing
+        # additional redundant processing outside the form.
+        raise ValueError(
+            "Mailbox values should be present if and only if disallowed authors are persent."
+        )
+
     if assessment.warnings:
         # Log any warnings that were acknowledged, for auditing.
         logger.info(
@@ -296,9 +310,14 @@ def post(phab: PhabricatorClient, user: User, data: dict) -> tuple[dict[str, int
             ),
             flags,
         )[1]
-        lando_revision = fetch_raw_diff_and_save(
-            phab, revision["id"], diff, commit_message
-        )
+
+        args = (phab, revision["id"], diff, commit_message)
+        kwargs = {}
+
+        if diff_has_disallowed_author(diff):
+            kwargs["mailbox"] = mailbox
+
+        lando_revision = fetch_raw_diff_and_save(*args, **kwargs)
 
         revision_reviewers[lando_revision.id] = get_approved_by_ids(
             phab,
--- a/src/lando/api/legacy/revisions.py
+++ b/src/lando/api/legacy/revisions.py
@@ -268,7 +268,7 @@ def fetch_raw_diff_and_save(
     revision_id: int,
     diff: dict,
     commit_message: str,
-    mailbox: tuple[str] | None = None,
+    mailbox: tuple[str, str] | None = None,
 ) -> Revision:
     """Fetch the raw diff from Phabricator and save a `Revision` record.
 
--- a/src/lando/api/legacy/revisions.py
+++ b/src/lando/api/legacy/revisions.py
@@ -280,6 +280,7 @@ def fetch_raw_diff_and_save(
         revision_id: The integer revision ID (e.g. 123 for D123).
         diff: A Phabricator diff dict with the `commits` attachment.
         commit_message: The full commit message to embed in the patch.
+        mailbox: A tuple consisting of the author name and email.
     """
     diff_id = PhabricatorClient.expect(diff, "id")
     author_name, author_email = mailbox or select_diff_author(diff)
--- a/src/lando/api/legacy/revisions.py
+++ b/src/lando/api/legacy/revisions.py
@@ -155,20 +155,6 @@ def blocker_diff_author_is_known(*, diff: dict, **kwargs) -> Optional[str]:
     )
 
 
-def blocker_diff_author_is_hackbot(*, diff: dict, **kwargs) -> Optional[str]:
-    """Block revisions that contain commits by Hackbot."""
-    hackbot_email = "hackbot@mozilla.tld"
-    commits = PhabricatorClient.expect(diff, "attachments", "commits", "commits")
-    if not commits:
-        return None
-
-    emails = (c.get("author", {}).get("email", "").lower() for c in commits)
-    if hackbot_email not in emails:
-        return None
-
-    return "Diff contains commit authored by Hackbot."
-
-
 def revision_has_needs_data_classification_tag(
     revision: dict, data_policy_review_phid: str
 ) -> bool:
@@ -282,6 +268,7 @@ def fetch_raw_diff_and_save(
     revision_id: int,
     diff: dict,
     commit_message: str,
+    mailbox: tuple[str] | None = None,
 ) -> Revision:
     """Fetch the raw diff from Phabricator and save a `Revision` record.
 
@@ -295,7 +282,7 @@ def fetch_raw_diff_and_save(
         commit_message: The full commit message to embed in the patch.
     """
     diff_id = PhabricatorClient.expect(diff, "id")
-    author_name, author_email = select_diff_author(diff)
+    author_name, author_email = mailbox or select_diff_author(diff)
     base_revision = get_base_revision_from_diff(diff)
 
     logger.debug("Fetching raw diff for D%s (diff %s).", revision_id, diff_id)
--- a/src/lando/api/legacy/transplants.py
+++ b/src/lando/api/legacy/transplants.py
@@ -24,7 +24,6 @@ from lando.api.legacy.reviews import (
     reviewer_identity,
 )
 from lando.api.legacy.revisions import (
-    blocker_diff_author_is_hackbot,
     blocker_diff_author_is_known,
     gather_involved_phids,
     revision_has_needs_data_classification_tag,
@@ -46,7 +45,7 @@ from lando.main.models import (
     LandingJob,
     Repo,
 )
-from lando.main.support import LegacyAPIException
+from lando.main.support import DISALLOWED_AUTHOR_EMAILS, LegacyAPIException
 from lando.utils.landing_checks import (
     DiffAssessor,
     PreventNSPRNSSCheck,
@@ -331,6 +330,20 @@ class RevisionWarningCheck:
             )
 
         return wrapped
+
+
+@RevisionWarningCheck("Commit contains author that will be modified before landing.")
+def warning_diff_author_is_hackbot(
+    revision: dict, diff: dict, stack_state: StackAssessmentState
+) -> str | None:
+    """Warn when revisions contain commits by disallowed authors (e.g., Hackbot)."""
+    commits = PhabricatorClient.expect(diff, "attachments", "commits", "commits")
+    if not commits:
+        return None
+
+    emails = (c.get("author", {}).get("email", "").lower() for c in commits)
+    if set(emails).intersection(DISALLOWED_AUTHOR_EMAILS):
+        return "Diff contains commit authored by disallowed email."
 
 
 @RevisionWarningCheck("Has a review intended to block landing.")
@@ -878,7 +891,6 @@ REVISION_BLOCKER_CHECKS = [
     blocker_latest_diffs,
     blocker_author_planned_changes,
     blocker_diff_author_is_known,
-    blocker_diff_author_is_hackbot,
     blocker_uplift_approval,
     blocker_revision_data_classification,
     # Diff-based checks.
@@ -901,6 +913,7 @@ WARNING_CHECKS = [
     warning_wip_commit_message,
     warning_unresolved_comments,
     warning_multiple_authors,
+    warning_diff_author_is_hackbot,
 ]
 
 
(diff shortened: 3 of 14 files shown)


zzzeid:
NOTE: there is no front-end validation of the inputs, however, backend validation is exposed once the form is submitted and the page reloads.


image

src/lando/main/support.py
```diff
@@ -1,6 +1,10 @@
 from django.core.files.storage import storages
 from storages.backends.gcloud import GoogleCloudStorage
 
+from lando.api.legacy.revisions import select_diff_author
+
+DISALLOWED_AUTHOR_EMAILS = ("hackbot@mozilla.tld",)
```
shtrom: nit: We should make this a configuration variable, or at least something in `settings`.

src/lando/ui/legacy/forms.py
```diff
@@ -10,15 +13,69 @@
     UpliftTargetSelectionMethod,
     YesNoChoices,
 )
+from lando.main.support import DISALLOWED_AUTHOR_EMAILS
 
 
 class TransplantRequestForm(forms.Form):
+    def validate_author(value):
+        if value in DISALLOWED_AUTHOR_EMAILS:
```
shtrom: Is `value` `strip()`ed on the way?

src/lando/ui/legacy/revisions.py
```diff
@@ -494,6 +495,30 @@ def get(
...
+            # Take the first revision author, try to find an associated user in Lando.
+            # If this is not possible, set the mailbox to the name of the user, which
+            # must be modified before submitting.
+            author_phid = revisions_with_disallowed_authors[0]["author"]["phid"]
+            try:
+                lando_user = Profile.objects.get(phabricator_phid=author_phid).user
+            except Profile.DoesNotExist:
+                form.fields["author_name"].initial = revisions_with_disallowed_authors[
+                    0
+                ]["author"]["real_name"]
```
shtrom, line 510: Is it on purpose that the email is not set here?

src/lando/ui/legacy/revisions.py
```diff
@@ -559,9 +585,6 @@ def post(
             errors.append("You must be logged in to request a landing")
 
         if form.is_valid() and not errors:
-            form.cleaned_data["landing_path"] = json.loads(
-                form.cleaned_data["landing_path"]
-            )
```
shtrom, line 564: Is this removal on purpose?

shtrom (changes requested):
LGTM. Not approving because of the two questions in `revisions.py`.

src/lando/ui/legacy/revisions.py
zzzeid: Yes, there is no email provided by Phabricator via the API.

src/lando/ui/legacy/revisions.py
zzzeid: Yes, this is a bug that I uncovered when I added test coverage for this file. Nice catch though, since I forgot to also remove `json.dumps` on [line 463](https://github.com/mozilla-conduit/lando/pull/1406/changes#diff-ca2ea26a8ae84529a991703e7007daab9994bf99f0bdbd570842fca18e860a43R462) as well. Did that, and filed a [separate bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2060918) for this (but will keep the fix in this file since this is where the test coverage is added). Will add a mention in the commit message.

src/lando/main/support.py
zzzeid: > settings

I'll put this in `settings` for now. A config variable would be better here with some default, but I also dislike overloading the JSON type for this purpose. Probably should add a `CSV` config variable type that converts comma-separated strings to a tuple of strings.

src/lando/ui/legacy/forms.py
zzzeid: Was not, now is.

src/lando/api/legacy/api/transplants.py
```diff
@@ -218,6 +218,20 @@ def post(phab: PhabricatorClient, user: User, data: dict) -> tuple[dict[str, int
...
+    has_disallowed_author_diffs = any(
+        diff_has_disallowed_author(diff) for revision, diff in to_land
+    )
+    if (has_disallowed_author_diffs and not mailbox) or (
+        not has_disallowed_author_diffs and mailbox
+    ):
+        # TODO: it would be more ideal if this is a validation error, however, we do not have the
+        # necessary information at the time of form submission to determine this without performing
+        # additional redundant processing outside the form.
+        raise ValueError(
```
cgsheeh: This is reachable by a normal user (open the modal, clear both author fields, submit), so we'd be handing out a 500 for what is really a bad request. Please raise `LegacyAPIException(400, ...)` like the flags check just above, so it renders as an error message instead of an unhandled exception. Also a typo in the message: "persent" -> "present".

src/lando/api/legacy/api/transplants.py
```diff
@@ -296,9 +310,14 @@ def post(phab: PhabricatorClient, user: User, data: dict) -> tuple[dict[str, int
...
-            phab, revision["id"], diff, commit_message
-        )
+
+        args = (phab, revision["id"], diff, commit_message)
+        kwargs = {}
+
+        if diff_has_disallowed_author(diff):
+            kwargs["mailbox"] = mailbox
+
+        lando_revision = fetch_raw_diff_and_save(*args, **kwargs)
```
cgsheeh: The `args`/`kwargs` indirection is harder to follow than just passing the keyword: `fetch_raw_diff_and_save(phab, revision["id"], diff, commit_message, mailbox=mailbox if diff_has_disallowed_author(diff) else None)`. We're also calling `diff_has_disallowed_author` a second time per diff here; consider collecting the disallowed diffs once above and reusing that.

src/lando/api/legacy/revisions.py
```diff
@@ -282,6 +268,7 @@ def fetch_raw_diff_and_save(
     revision_id: int,
     diff: dict,
     commit_message: str,
+    mailbox: tuple[str] | None = None,
```
cgsheeh: This should be `tuple[str, str] | None` since we always unpack two values from it. Please also document `mailbox` in the `Args` section of the docstring below.

src/lando/main/support.py
```diff
@@ -1,6 +1,9 @@
+from django.conf import settings
 from django.core.files.storage import storages
 from storages.backends.gcloud import GoogleCloudStorage
 
+from lando.api.legacy.revisions import select_diff_author
```
cgsheeh, line 5: Feels like `revisions` should be importing from `support`, not the other way around.

src/lando/main/support.py
```diff
@@ -31,3 +34,20 @@ def __init__(self, status, detail, extra=None):  # noqa: ANN001
...
+    return (
+        revision["diff"]["author"]["email"].strip().lower()
+        in settings.DISALLOWED_AUTHOR_EMAILS
+    )
+
+
+def diff_has_disallowed_author(diff: dict) -> bool:
+    return (
+        select_diff_author(diff)[1].strip().lower() in settings.DISALLOWED_AUTHOR_EMAILS
+    )
```
cgsheeh: `select_diff_author` returns `(None, None)` when a diff has no commits, so `.strip()` raises `AttributeError`. Right now we only get away with it because `blocker_diff_author_is_known` runs first; `(select_diff_author(diff)[1] or "")` would make it safe on its own.

src/lando/main/support.py
```diff
@@ -31,3 +34,20 @@ def __init__(self, status, detail, extra=None):  # noqa: ANN001
         }
         if self.extra:
             self.json_detail.update(self.extra)
+
+
+def get_revisions_with_disallowed_authors(revisions: dict[str, dict]) -> list[dict]:
+    return [r for r in revisions.values() if revision_has_disallowed_author(r)]
```
cgsheeh, line 44: Nit: these methods could use a short description of what makes an author "disallowed".

src/lando/ui/legacy/forms.py
```diff
@@ -19,6 +23,60 @@ class TransplantRequestForm(forms.Form):
...
+        else:
+            cleaned_data["mailbox"] = None
+
+        # This removes the error entries from these fields, if no errors were found.
+        if not self.errors["author_email"]:
+            del self.errors["author_email"]
+        if not self.errors["author_name"]:
+            del self.errors["author_name"]
+
+        return cleaned_data
```
cgsheeh, line 69: Mutating `self.errors` directly (seeding empty lists, then deleting them again) is doing by hand what `self.add_error("author_email", "This field is required.")` already does, and it drops out the pre-seed/cleanup blocks entirely. Also "Peform" -> "Perform", and our docstrings start the summary on the opening line.

src/lando/api/legacy/transplants.py
```diff
@@ -333,6 +333,20 @@ def wrapped(revision: dict, diff: dict, stack_state: StackAssessmentState):
...
+) -> str | None:
+    """Warn when revisions contain commits by disallowed authors (e.g., Hackbot)."""
+    commits = PhabricatorClient.expect(diff, "attachments", "commits", "commits")
+    if not commits:
+        return None
+
+    emails = (c.get("author", {}).get("email", "").strip().lower() for c in commits)
+    if set(emails).intersection(settings.DISALLOWED_AUTHOR_EMAILS):
+        return "Diff contains commit authored by disallowed email."
+
```
cgsheeh, line 348: Should we display the bad email in the error message so the user knows why they're seeing the warning?

src/lando/ui/jinja2/stack/partials/landing-preview.html
```diff
@@ -8,6 +8,21 @@ <h3 class="StackPage-landingPreview-sectionLabel">Landing is Blocked</h3>
...
+        <div>
+            <div class="field">
+                <label class="label">New Commit Author Name</label>
+                <div class="control">{{ form.author_name }}</div>
+            </div>
+            <div class="field">
+                <label class="label">New Commit Author Email</label>
+                <div class="control">{{ form.author_email }}</div>
+            </div>
+        </div>
```
cgsheeh, line 24: With more than one affected revision the links render back-to-back with no separator (ie `D1D2`). We should add a comma or space between them.

src/lando/ui/tests/test_stacks.py
```diff
@@ -103,3 +103,72 @@ def test_draw_stack_graph_complex():
...
+
+    data = {"landing_path": '[{"revision_id": "D1", "diff_id": 1}]'}
+    if trigger:
+        revision_author = phabdouble.api_object_for(r1_author)
+        form = authenticated_client.get(f"/D{r1['id']}/").context_data["form"]
+        assert (
+            form.fields["author_name"].initial == revision_author["fields"]["realName"]
+        )
+        assert form.fields["author_email"].initial is None
+        data["confirmation_token"] = (
```
cgsheeh: Hardcoding the confirmation token hash makes this test break any time a warning's wording changes. Can we read it out of the GET response instead?

src/lando/ui/tests/test_stacks.py
```diff
@@ -103,3 +103,72 @@ def test_draw_stack_graph_complex():
...
+        user.profile.save()
+        form = authenticated_client.get(f"/D{r1['id']}/").context_data["form"]
+        assert form.fields["author_name"].initial == user.profile.userinfo["name"]
+        assert form.fields["author_email"].initial == user.profile.userinfo["email"]
+
+        data["author_name"] = form.fields["author_name"].initial
+        data["author_email"] = form.fields["author_email"].initial
+        response = authenticated_client.post(f"/D{r1['id']}/", data=data)
+        messages = list(response.wsgi_request._messages)
+        assert len(messages) == 0
```
cgsheeh, line 171: This is the case that matters most and we only assert that no error messages were emitted. Can we assert the saved `Revision`'s patch author name/email actually match the submitted override? That's the behaviour the bug is about, and the test name claims it checks the DB.

src/lando/ui/tests/test_stacks.py
```diff
@@ -103,3 +103,72 @@ def test_draw_stack_graph_complex():
...
+@pytest.mark.parametrize(
+    "username,email,trigger",
+    [
+        ("Hackbot", "hackbot@mozilla.tld", True),
+        ("Hackbot", " haCkbOt@moziLla.Tld ", True),
+        ("Someone else", "test@example.org", False),
+    ],
+)
+@pytest.mark.django_db(transaction=True)
+def test_integrated_transplant_simple_partial_stack_saves_data_in_db(
```
cgsheeh: The name is copied from another test and doesn't describe what this one does — something like `test_transplant_disallowed_author_requires_author_override` would be clearer.

src/lando/ui/tests/test_forms.py
```diff
@@ -72,3 +72,69 @@ def test_uplift_assessment_link_form_revision_ids(
...
+                "author_email": [
+                    "The email provided is not allowed",
+                    "This field is required.",
+                ]
+            },
+        ),
+    ),
+)
+def test__TransplantRequestForm(data, mailbox, errors):
+    from lando.ui.legacy.forms import TransplantRequestForm
```
cgsheeh: Please move this import to the top of the file with the other form imports.

cgsheeh (changes requested):

src/lando/api/legacy/api/transplants.py
zzzeid: > This is reachable by a normal user (open the modal, clear both author fields, submit), so we'd be handing out a 500 for what is really a bad request. Please raise LegacyAPIException(400, ...) like the flags check just above, so it renders as an error message instead of an unhandled exception.

Can change that, but see https://bugzilla.mozilla.org/show_bug.cgi?id=1956972 - `LegacyAPIException` is also rendered as a 500 error at this time so there's no functional change here. We also use `ValueError` in other spots.

As mentioned in the comment I think this should be adapted to a validation error but is out of scope for this PR.

src/lando/main/support.py
zzzeid: `select_diff_author` is old code that has existed in that legacy module for a long time, so we're just reusing it here for convenience. I don't think we need to move it to this module.

src/lando/ui/tests/test_stacks.py
zzzeid: Indeed it is copied, along with some of the entry logic. This is basically a broader integration test that does some of the logic in the original test that it copied. Will change the name.

src/lando/api/legacy/revisions.py
zzzeid: This seems unnecessary / implied given that there is one type within the tuple - the [documentation](https://docs.python.org/3/library/typing.html#annotating-tuples) generally references this need if the types are different for each element, not if they are the same. Do you have a reference for this?

src/lando/main/support.py
zzzeid: I am wondering if that code path ever executes, since there should be no case of a diff without a commit.

src/lando/api/legacy/transplants.py
zzzeid: This string is not visible in the UI (that one is the one in the decorator above, which has no access to this set). I guess this is an artifact of these legacy warnings and the way they get parsed (we ignore the "detail" but show the `display` value.)

src/lando/main/support.py
zzzeid: A little too verbose, but wouldn't hurt.

src/lando/ui/legacy/forms.py
zzzeid: `self.add_error` is much nicer indeed.

src/lando/api/legacy/api/transplants.py
zzzeid: Would rather not pass a `None` as a value, though changed it to be explicit if/else instead.

src/lando/api/legacy/revisions.py
cgsheeh: Here's the relevant code snippet from that link:

```
# OK: ``x`` is assigned to a tuple of length 1 where the sole element is an int
x: tuple[int] = (5,)

# OK: ``y`` is assigned to a tuple of length 2;
# element 1 is an int, element 2 is a str
y: tuple[int, str] = (5, "foo")

# Error: the type annotation indicates a tuple of length 1,
# but ``z`` has been assigned to a tuple of length 3
z: tuple[int] = (1, 2, 3)
```

You can see the `z` example is what matches the usage in the PR, and it's called out as an error due to the length mismatch.

zzzeid: I updated this since it's not really worth the time or argument, but the actual text does not specify anything like this and it is not flagged as an error by a backup checker that I have. I think the example contradicts the documentation, or at least specifies additional aspects that are not present there.

>Unlike most other Python containers, however, **it is common in idiomatic Python code for tuples to have elements which are not all of the same type**. **For this reason**, tuples are special-cased in Python’s typing system. [tuple](https://docs.python.org/3/library/stdtypes.html#tuple) accepts any number of type arguments:

src/lando/main/support.py
shtrom: I find a JSON array tolerable (:

Though maybe that would justify a separate type to make sure it's a list.

src/lando/api/legacy/api/transplants.py
```diff
@@ -218,6 +218,21 @@ def post(phab: PhabricatorClient, user: User, data: dict) -> tuple[dict[str, int
         )
         raise LegacyAPIException(400, error_message)
 
+    mailbox = data.get("mailbox", None)
+    has_disallowed_author_diffs = any(
+        diff_has_disallowed_author(diff) for revision, diff in to_land
+    )
+    if (has_disallowed_author_diffs and not mailbox) or (
+        not has_disallowed_author_diffs and mailbox
```
shtrom, line 226: nit: That's an XOR 😆

shtrom (approved):
lgtm, will deploy.

PR #1408, opened 2026-07-31
try_api: enable HeadlessAPIAuthentication for M2M (bug 1994568)

shtrom:


src/lando/try_api/tests/test_api.py
```diff
@@ -390,3 +391,136 @@ def test_try_api_patches_success(
...
+    client_post: Callable,
+):
+    user = scm_user([Permission.objects.get(codename="scm_level_1")], "password")
+    token = ApiToken.create_token(user)
+
+    mock_authenticate = mock_authenticate_builder(None)
+    mock_authenticate.return_value = None
+
+    response = client_post(
+        "/try/patches",
```
cgsheeh: This should hit `/api/try/patches` instead, no?

src/lando/try_api/tests/test_api.py
```diff
@@ -390,3 +391,136 @@ def test_try_api_patches_success(
...
+    rj = response.json()
+    assert rj, "Error response should be a parseable (RFC 7807) JSON payload"
+    assert "title" in rj, f"Missing title in error 400 response: {response.text}"
+    assert rj["title"] == "Forbidden"
+    assert "detail" in rj, f"Missing detail in error 400 response: {response.text}"
+    assert rj["detail"] == "Missing permissions: main.scm_level_1"
+
+
+@pytest.mark.django_db()
+def test_try_api_patches_m2m_auth_success(
```
cgsheeh: This and `test_try_api_patches_m2m_auth_no_scm1` are verbatim copies of the existing `test_try_api_patches_success`/`_no_scm1` apart from how the user is authenticated. Could we parametrize the auth mechanism (a fixture returning either the mocked OIDC user or an `ApiToken` header) rather than duplicating the bodies?

src/lando/try_api/api.py
```diff
@@ -31,7 +32,9 @@
 
 logger = logging.getLogger(__name__)
 
-api = NinjaAPI(auth=AccessTokenAuth(), urls_namespace="try")
+api = NinjaAPI(
+    auth=[HeadlessAPIAuthentication(), AccessTokenAuth()], urls_namespace="try"
```
cgsheeh: Guess we're okay leaving the legacy endpoint as `AccessTokenAuth` only, just wanted to point out and make sure that's intentional.

cgsheeh (changes requested):

src/lando/try_api/api.py
shtrom: Yeah, they get 308d anyway.

src/lando/try_api/tests/test_api.py
shtrom: Ah, yep, good call.

src/lando/try_api/tests/test_api.py
```diff
@@ -144,14 +142,19 @@ def test_try_api_patches_no_scm1(
...
+        token = ApiToken.create_token(user)
 
     response = client_post(
         "/api/try/patches",
-        headers={"AuThOrIzAtIoN": "bEaReR token no_scm1"},
+        headers={"AuThOrIzAtIoN": f"bEaReR {token}"},
     )
 
-    assert mock_authenticate.called, "Authentication backend should be called"
+    if auth_type == "auth0":
```
cgsheeh: `auth_type` is parametrized as `oauth`/`headless`, never `auth0`, so this assertion never runs and the `oauth` case no longer verifies the backend was hit. Same at line 380.

src/lando/try_api/tests/test_api.py
```diff
@@ -363,10 +374,11 @@ def test_try_api_patches_success(
     response = client_post(
         "/api/try/patches",
         data=json.dumps(request_payload),
-        headers={"AuThOrIzAtIoN": "bEaReR token success"},
+        headers={"AuThOrIzAtIoN": f"bEaReR {token}"},
     )
 
-    assert mock_authenticate.called, "Authentication backend should be called"
+    if auth_type == "auth0":
```
cgsheeh: Same dead branch as above: `auth0` is never a parametrized value.

src/lando/try_api/tests/test_api.py
```diff
@@ -115,13 +117,8 @@ def test_try_api_patches_invalid_user(
...
 @pytest.mark.parametrize(
-    "group_scm_1,superuser",
-    (
-        (False, False),
-        (False, True),
-        (True, False),
-        (True, True),
-    ),
+    "group_scm_1,superuser,auth_type",
+    itertools.product((False, True), (False, True), ("oauth", "headless")),
```
cgsheeh: Seeing a new warning from Pytest when running this:

```
../usr/local/lib/python3.12/site-packages/_pytest/python.py:124: 16 warnings
  /usr/local/lib/python3.12/site-packages/_pytest/python.py:124: PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
  Test: src/lando/api/tests/test_landings.py::test_merge_conflict, argvalues type: generator
  Please convert to a list or tuple.
  See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
    metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker
```

I think we need to do `tuple(itertools.product(...` so this isn't a bare iterator.

src/lando/try_api/tests/test_api.py
```diff
@@ -144,14 +142,19 @@ def test_try_api_patches_no_scm1(
     if superuser:
         user = make_superuser(user)
 
-    mock_authenticate = mock_authenticate_builder(user)
+    token = "token no_scm1"
```
cgsheeh: Unlike `test_try_api_patches_success`, there's no `else: raise ValueError` here, so a typo'd `auth_type` would silently fall through to the literal `"token no_scm1"` and still pass with a 403. Worth mirroring that guard.

src/lando/try_api/tests/test_api.py
```diff
@@ -390,3 +402,29 @@ def test_try_api_patches_success(
...
+def test_try_api_patches_m2m_auth_invalid_token(
+    mock_authenticate_builder: Callable,
+    scm_user: Callable,
+    client_post: Callable,
+):
+    user = scm_user([Permission.objects.get(codename="scm_level_1")], "password")
+    token = ApiToken.create_token(user)
+
+    mock_authenticate = mock_authenticate_builder(None)
+    mock_authenticate.return_value = None
```
cgsheeh: `mock_authenticate_builder` sets `side_effect`, which takes precedence over `return_value`, so this line is a no-op. `mock_authenticate_builder(None)` already gives you the behaviour you want.

src/lando/try_api/api.py
cgsheeh: Could we add a comment indicating why we used this order? :)

cgsheeh (changes requested):
LGTM, just found a few issues in the tests. :)

src/lando/try_api/api.py
shtrom: Added a comment explaining why the order is now swapped (:

src/lando/try_api/tests/test_api.py
shtrom: Ah, yeah good catch. I chose more generic naming half-way through, but forgot to change everywhere... Or to catch invalid values...

src/lando/try_api/tests/test_api.py
shtrom: Ah! I was looking for this warning!. I fixed a bunch of them before, but couldn't remember where it was coming from.

I also fixed the other tests raising the same warning, so that it won't hide in the noise next time.

src/lando/try_api/tests/test_api.py
```diff
@@ -144,11 +142,16 @@ def test_try_api_patches_no_scm1(
     if superuser:
         user = make_superuser(user)
 
+    token = f"token no_scm1 {auth_type}"
     mock_authenticate = mock_authenticate_builder(user)
+    if auth_type == "headless":
```
cgsheeh: The `headless` parametrization doesn't exercise anything now. `mock_authenticate_builder(user)` makes `AccessTokenAuth` succeed for any token, and it is first in the chain, so `HeadlessAPIAuthentication` is never reached — I confirmed by making `HeadlessAPIAuthentication.authenticate` raise, and every `[headless]` case still passes. Please build the mock with `None` in the headless case so the request actually falls through to the token authenticator.

src/lando/try_api/tests/test_api.py
```diff
@@ -332,17 +335,24 @@ def test_try_api_patches_failed_checks(
...
     commit_maps: list[CommitMap],
     git_patch: Callable,
     client_post: Callable,
+    auth_type: str,
 ):
     user = scm_user([Permission.objects.get(codename="scm_level_1")], "password")
 
+    token = f"token success {auth_type}"
     mock_authenticate = mock_authenticate_builder(user)
+    if auth_type == "headless":
```
cgsheeh: Same problem here: the OAuth mock succeeds regardless of the token, so `[headless]` is a duplicate of `[oauth]` and none of the job/`requester_email` assertions ever cover the API-token path. This is the test I'd most want to see actually pass through `HeadlessAPIAuthentication`.

src/lando/try_api/api.py
```diff
@@ -31,7 +32,14 @@
 
 logger = logging.getLogger(__name__)
 
-api = NinjaAPI(auth=AccessTokenAuth(), urls_namespace="try")
+api = NinjaAPI(
+    # We expect orders of magnitude more authentication with OAuth access tokens rather
+    # than Headless API token. So we put the former first, even though it requires network
+    # interactions that the latter doesn't. The latter also does log at INFO level when a
```
cgsheeh: Nit; we could make these comment lines conform to the 88-char limit.

src/lando/try_api/tests/test_api.py
```diff
@@ -390,3 +404,28 @@ def test_try_api_patches_success(
...
+
+
+@pytest.mark.django_db()
+def test_try_api_patches_m2m_auth_invalid_token(
+    mock_authenticate_builder: Callable,
+    scm_user: Callable,
+    client_post: Callable,
+):
+    user = scm_user([Permission.objects.get(codename="scm_level_1")], "password")
+    token = ApiToken.create_token(user)
```
cgsheeh, line 423: We could add a test for an `is_valid=False` token as well, since we would use that mechanism to revoke access.

shtrom:
Rebases to work around landing conflict https://lando.moz.tools/landings/78456/

PR #1410, opened 2026-07-31
headless_api: make HeadlessAPIAuthentication chainable with other NinjaAPI authenticators (bug 1994568)

shtrom:
* headless_api: don't require User-Agent in auth
 * headless_api: return None on failed token lookup


src/lando/headless_api/api.py
```diff
@@ -58,27 +56,15 @@ def reject_cli_flags(value: str) -> str:
...
-        if not user_agent.startswith("Lando-User/"):
-            raise APIPermissionDenied("Incorrect `User-Agent` format.")
-
+    def authenticate(self, request: WSGIRequest, token: str) -> ApiToken | None:
         try:
             api_key = ApiToken.verify_token(token)
-        except ValueError as exc:
-            raise APIPermissionDenied(str(exc))
+        except ValueError:
+            # Allow subsequent authentication methods to be tested.
```
cgsheeh, line 71: Swallowing the `ValueError` makes token failures completely silent. Could we add a `logger.debug` here with the token prefix so we know when we've failed to verify a token?

src/lando/headless_api/tests/test_automation_api.py
```diff
@@ -140,7 +97,7 @@ def test_auth_invalid_token(client, headless_user, automation_job):
     assert response.status_code == 401, (
         "Invalid API key should result in 401 status code."
     )
-    assert response.json() == {"details": "Token api-bad-key was not found."}
+    assert response.json() == {"detail": "Unauthorized"}
```
cgsheeh, line 100: This changes the client-side contract for response errors. Do Lando-CLI and the landoscript worker depend on the `details` key?

cgsheeh (dismissed):

src/lando/headless_api/api.py
shtrom: We can. However, this will log every single time that we have an Auth0 token, because the Headless Auth is first (in https://github.com/mozilla-conduit/lando/pull/1408/changes#diff-af456f968caba867ca1e61b5693012347b03c6912c7cc03dd3d86277ee410091R36). We could move it to last, but my reasoning in putting it first was that it doesn't require a network round-trip to a third party, so felt cheaper.

OTOH, the vast majority of requests would be users via Auth0, some maybe it's a case to swap the order of authenticators. 

WDYT?

src/lando/headless_api/tests/test_automation_api.py
shtrom: eeeerr. Yes. I even filed a bug about it 😅 https://bugzilla.mozilla.org/show_bug.cgi?id=2059648

The issue we have here is that it's not the same exception that triggers, and it's now a Django-internal one. So we can't change it wholesale in either direction...

I'd say the best move forwards it so fix lando-cli first.

shtrom: -> https://github.com/mozilla-conduit/lando-cli/pull/20

shtrom:
Rebasing to see if that gets it to run the tests again...

src/lando/headless_api/api.py
```diff
@@ -58,27 +56,20 @@ def reject_cli_flags(value: str) -> str:
...
-
-        if not user_agent.startswith("Lando-User/"):
-            raise APIPermissionDenied("Incorrect `User-Agent` format.")
-
+    def authenticate(self, request: WSGIRequest, token: str) -> ApiToken | None:
         try:
             api_key = ApiToken.verify_token(token)
-        except ValueError as exc:
-            raise APIPermissionDenied(str(exc))
+        except ValueError:
```
cgsheeh, line 65: Do we get debug logs in prod? IIRC there's an open PR to enable that. Maybe we should log at `info` to be safe?

src/lando/headless_api/api.py
```diff
@@ -58,27 +56,20 @@ def reject_cli_flags(value: str) -> str:
...
+            # Shorten the token so we don't show it all.
+            prefix_length = 8
+            if (tl := len(token)) <= prefix_length:
+                prefix_length = tl // 2
+            logger.debug("Invalid headless token f{token[:l]}")
```
cgsheeh: `api_key.token_prefix` should work here, right?

cgsheeh: I think this is meant to be an f-string.

src/lando/headless_api/api.py
cgsheeh: I think your reasoning makes sense with querying the third-party auth system after the local one. However we will likely only have a small number of consumers of the local auth, and the log pollution could get painful.

Let's swap the order for now, and we can stomach the noisy logging only after we run into issues attempting Auth0 first. :)

cgsheeh (changes requested):

src/lando/headless_api/api.py
shtrom: I thought we did have debug logs for lando, but maybe not. 

Ok, there's my old PR here https://github.com/mozilla-conduit/lando/pull/253

This would be the key, I think: https://github.com/mozilla-conduit/lando/pull/253#discussion_r2106475256

src/lando/headless_api/api.py
shtrom: No, because that's handling the case where an `api_key` was not returned.

OTOH, I can reuse `API_TOKEN_PREFIX_LENGTH`.

shtrom: Wow, that's a nice one! `“f{var}”` 🤦🏼

src/lando/headless_api/tests/test_automation_api.py
shtrom: lando-cli 0.4.1 released!

src/lando/headless_api/api.py
shtrom: This'll happen on the next PR of the stack.

src/lando/headless_api/api.py
shtrom: Ok, making it info for now, with the idea that we also swap the order of Authenticators in the nex PR up, so there's not too much noise.

src/lando/headless_api/api.py
shtrom: https://github.com/mozilla-conduit/lando/pull/1408/changes/a3a80ac97c6f2fa31e9145b794664ce845b40c1e

shtrom:
Force-pushing after rebase due to landing failure https://lando.moz.tools/landings/77501/.

shtrom:
A dodgy change tagged along... Reverted.

PR #1432, opened 2026-08-05
generate_requirements: add --no-cooldown option, and configurable make target (bug 2060725)

shtrom:
This allows us to refresh the requirements.txt without upgrading,

    make upgrade-requirements UPGRADE_ARGS=''

or to bump a (known) dependency without cooldown,

    make upgrade-requirements UPGRADE_ARGS='--no-cooldown'

---

Lando: [link](https://lando.moz.tools/pulls/lando/1432/)
Bugzilla: [bug 2060725](https://bugzilla.mozilla.org/show_bug.cgi?id=2060725)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,5 +79,6 @@ upgrade-npm-packages: ## upgrade-npm-packages update package-lock.json
 	$(BASE_COMMAND) npm install --before="$(shell date -I -d '7 days ago')"
 
 .PHONY: upgrade-requirements
-upgrade-requirements: ## upgrade-requirements upgrade packages in requirements.txt
-	$(BASE_COMMAND) lando generate_requirements --upgrade
+upgrade-requirements: UPGRADE_ARGS=--upgrade
+upgrade-requirements: ## upgrade-requirements upgrade packages in requirements.txt. Pass `UPGRADE_ARGS=` to the `make` call to only regenerate the file without upgrades. `--no-cooldown` Is also available.
+	$(BASE_COMMAND) lando generate_requirements $(UPGRADE_ARGS)
--- a/src/lando/utils/management/commands/generate_requirements.py
+++ b/src/lando/utils/management/commands/generate_requirements.py
@@ -19,7 +19,7 @@ class Command(BaseCommand):
             "--cooldown",
             action=argparse.BooleanOptionalAction,
             default=True,
-            help="Whether to apply a cooldown an upgrades",
+            help="Only consider packages uploaded more than 7 days ago (default)",
         )
 
     def handle(self, *args, **options):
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -79,5 +79,6 @@ upgrade-npm-packages: ## upgrade-npm-packages update package-lock.json
 	$(BASE_COMMAND) npm install --before="$(shell date -I -d '7 days ago')"
 
 .PHONY: upgrade-requirements
-upgrade-requirements: ## upgrade-requirements upgrade packages in requirements.txt
-	$(BASE_COMMAND) lando generate_requirements --upgrade
+upgrade-requirements: UPGRADE_ARGS=--upgrade
+upgrade-requirements: ## upgrade-requirements upgrade packages in requirements.txt. Pass `UPGRADE_ARGS=` to the `make` call to only regenerate the file without upgrades. `--no-cooldown` Is also available.
```
cgsheeh, line 83: Because `UPGRADE_ARGS` replaces the default rather than appending to it, `UPGRADE_ARGS='--no-cooldown'` also drops `--upgrade`.

src/lando/utils/management/commands/generate_requirements.py
```diff
@@ -15,6 +15,12 @@ def add_arguments(self, parser: argparse.ArgumentParser):
             action="store_true",
             help="Upgrade all packages to latest versions",
         )
+        parser.add_argument(
+            "--cooldown",
+            action=argparse.BooleanOptionalAction,
+            default=True,
+            help="Whether to apply a cooldown an upgrades",
```
cgsheeh: Typo: "a cooldown an upgrades" -> "a cooldown on upgrades". Though since this applies to any resolution (including `add-requirements`), maybe "Only consider packages uploaded more than 7 days ago" is more accurate.

cgsheeh (dismissed):

Makefile
shtrom: Yes, that's on purpose. I don't think we really want to upgrade everything with no cooldown. This is for topical upgrades of specific dependencies ahead of cooldown. Though we could still do it with `UPGRADE_ARGS='--upgrade --no-cooldown'`. In practice, I think we want one or the other.

Maybe the problem is that the target is `upgrade-requirements`, which is now a bit of a misnomer for the non default cases?

src/lando/utils/management/commands/generate_requirements.py
shtrom:
```suggestion
            help="Only consider packages uploaded more than 7 days ago (default)",
```

Makefile
cgsheeh: Sounds reasonable to me. 👍

cgsheeh (approved):

PR #1433, opened 2026-08-05
github: return full PR description in PullRequestPatchHelper.get_commit_description (bug 2060745)

shtrom:
---

Lando: [link](https://lando.moz.tools/pulls/lando/1433/)
Bugzilla: [bug 2060745](https://bugzilla.mozilla.org/show_bug.cgi?id=2060745)

### DIFF
--- a/src/lando/utils/github.py
+++ b/src/lando/utils/github.py
@@ -755,8 +755,15 @@ class PullRequestPatchHelper(PatchHelper):
         raise NotImplementedError("`from_bytes_io` not implemented.")
 
     def get_commit_description(self) -> str:
-        """Reconstruct the commit description from the PR metadata"""
-        return self._pr.title + "\n\n" + self._pr.commit_body
+        """Returns the full commit description."""
+        # We can't use pr.commit_message here,
+        # as it also appends a trailer with the PR URL.
+        lines = [self._pr.title]
+
+        if self._pr.commit_body:
+            lines += ["", self._pr.commit_body]
+
+        return "\n".join(lines)
 
     @override
     def get_diff(self) -> str:
--- a/src/lando/utils/github.py
+++ b/src/lando/utils/github.py
@@ -755,8 +755,8 @@ class PullRequestPatchHelper(PatchHelper):
         raise NotImplementedError("`from_bytes_io` not implemented.")
 
     def get_commit_description(self) -> str:
-        """Returns the commit description."""
-        return self.get_header("subject")
+        """Reconstruct the commit description from the PR metadata"""
+        return self._pr.title + "\n\n" + self._pr.commit_body
 
     @override
     def get_diff(self) -> str:
--- a/src/lando/utils/tests/test_github.py
+++ b/src/lando/utils/tests/test_github.py
@@ -543,8 +543,12 @@ def test_PullRequestPatchHelper(github_api_client_pr: mock.Mock):
     pr_patch_helper = PullRequestPatchHelper(pr)
 
     assert (
-        pr_patch_helper.get_commit_description()
+        pr_patch_helper.get_commit_title()
         == "WIP: test pull request with multiple commits"
+    )
+    assert (
+        pr_patch_helper.get_commit_description()
+        == "WIP: test pull request with multiple commits\n\ntest description"
     )
     assert pr_patch_helper.get_timestamp() == "1761017419"
     assert pr_patch_helper.parse_author_information() == (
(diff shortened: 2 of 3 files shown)


src/lando/utils/github.py
```diff
@@ -755,8 +755,8 @@ def from_bytes_io(cls, bytes_io: io.BytesIO) -> "PatchHelper":
         raise NotImplementedError("`from_bytes_io` not implemented.")
 
     def get_commit_description(self) -> str:
-        """Returns the commit description."""
-        return self.get_header("subject")
+        """Reconstruct the commit description from the PR metadata"""
+        return self._pr.title + "\n\n" + self._pr.commit_body
```
cgsheeh: When the PR body is empty (or is entirely below the delimiter), `commit_body` is `""` and this returns `"title\n\n"` with trailing blank lines. `PullRequest.commit_message` already assembles title + body and handles that case. Can we delegate to it here, so the message we run checks against can't drift from the one we actually land in `views.py`?

src/lando/utils/github.py
```diff
@@ -755,8 +755,8 @@ def from_bytes_io(cls, bytes_io: io.BytesIO) -> "PatchHelper":
         raise NotImplementedError("`from_bytes_io` not implemented.")
 
     def get_commit_description(self) -> str:
-        """Returns the commit description."""
-        return self.get_header("subject")
+        """Reconstruct the commit description from the PR metadata"""
```
cgsheeh: Nit: missing trailing period. Also worth adding `@override` to match the sibling methods in this class.

src/lando/utils/tests/test_github.py
```diff
@@ -543,9 +543,13 @@ def test_PullRequestPatchHelper(github_api_client_pr: mock.Mock):
     pr_patch_helper = PullRequestPatchHelper(pr)
 
     assert (
-        pr_patch_helper.get_commit_description()
+        pr_patch_helper.get_commit_title()
         == "WIP: test pull request with multiple commits"
     )
+    assert (
+        pr_patch_helper.get_commit_description()
+        == "WIP: test pull request with multiple commits\n\ntest description"
```
cgsheeh: Can we also cover a PR with an empty body and one where the body is only the delimiter? Those are the cases that produce the trailing newlines mentioned below.

cgsheeh (dismissed):

src/lando/utils/github.py
shtrom: Oh 🤦🏼

shtrom: Ah, we can't really use `commit_message`, as it adds trailers.

But the same building logic can be used.

src/lando/utils/tests/test_github.py
shtrom: I added one for the empty body. The PR delimiters is already tested as part of `test__PullRequest___parse_body_segments__no_delimiter`, which feeds the `commit_body` variable we use, so it shouldn't reach that point anyway.

src/lando/utils/github.py
```diff
@@ -755,8 +755,15 @@ def from_bytes_io(cls, bytes_io: io.BytesIO) -> "PatchHelper":
         raise NotImplementedError("`from_bytes_io` not implemented.")
 
     def get_commit_description(self) -> str:
-        """Returns the commit description."""
-        return self.get_header("subject")
+        """Returns the full commit description."""
```
zzzeid: nit
```suggestion
        """Return the full commit description."""
```

src/lando/utils/tests/test_github.py
```diff
@@ -542,15 +542,23 @@ def test_PullRequestPatchHelper(github_api_client_pr: mock.Mock):
...
         "Olivier Mehani",
         "omehani@mozilla.com",
     )
+    assert (
+        pr_patch_helper.get_commit_description()
+        == f"{expected_commit_title}\n\ntest description"
+    ), "Commit description should be the full commit message"
+
+    pr_patch_helper._pr.commit_body = ""
+    assert pr_patch_helper.get_commit_description() == expected_commit_title, (
```
zzzeid: nit: I think it would be cleaner to separate these into explicit variables, as line 559 is a little confusing otherwise.

```python
assert pr_patch_helper.get_commit_description() == expected_description_title_with_body
pr_patch_helper._pr.commit_body = ""
assert pr_patch_helper.get_commit_description() == expected_description_title_only

PR #1434, opened 2026-08-05
landing_checks: return all blocking errors at once (bug 2060110)

shtrom:
Also, only show the first line of a commit message in errors, so as not to overwhelm the report with large repeated blocks of text. The commit title should be sufficient to identify the offending commit.

---

Lando: [link](https://lando.moz.tools/pulls/lando/1434/)
Bugzilla: [bug 2060110](https://bugzilla.mozilla.org/show_bug.cgi?id=2060110)

:warning: This pull request has 10 warnings.

### DIFF
--- a/src/lando/utils/landing_checks.py
+++ b/src/lando/utils/landing_checks.py
@@ -490,17 +490,17 @@ class CommitMessagesCheck(PatchCollectionCheck):
         commit_message = patch_helper.get_commit_description()
         author, _email = patch_helper.parse_author_information()
 
-        current_commit_issues = []
-
         if not commit_message:
             self.commit_message_issues.append("Revision has an empty commit message.")
             return
 
-        firstline = commit_message.splitlines()[0]
+        firstline = commit_message.partition("\n")[0]
 
         if self.ignore_bad_commit_message or "IGNORE BAD COMMIT MESSAGES" in firstline:
             self.ignore_bad_commit_message = True
             return
+
+        current_commit_issues = []
 
         # Ensure backout commit descriptions are well formed.
         if is_backout(firstline):
@@ -508,7 +508,7 @@ class CommitMessagesCheck(PatchCollectionCheck):
             if not backouts or not backouts[0]:
                 current_commit_issues.append(
                     "Revision is a backout but commit message "
-                    + f"does not indicate backed out revisions: {commit_message.partition('\n')[0]}"
+                    + f"does not indicate backed out revisions: {firstline}"
                 )
 
         # Avoid checks for the merge automation users.
@@ -519,33 +519,35 @@ class CommitMessagesCheck(PatchCollectionCheck):
         if "[PATCH" in firstline:
             current_commit_issues.append(
                 "Revision contains git-format-patch '[PATCH]' cruft. "
-                + f"Use git-format-patch -k to avoid this: {commit_message.partition('\n')[0]}"
+                + f"Use git-format-patch -k to avoid this: {firstline}"
             )
 
         if match := REPO_FLAG_RE.findall(firstline):
+            malformed = False
             for repo in match:
                 if "/" in repo:
                     current_commit_issues.append(
-                        f"Push contains commits intended to be locked to {repo} but the repo name is badly formatted. '/' is not allowed: {commit_message.partition('\n')[0]}"
+                        f"Push contains commits intended to be locked to {repo} but the repo name is badly formatted. '/' is not allowed: {firstline}"
                     )
-            if self.repo_name not in match:
+                    malformed = True
+            if not malformed and self.repo_name not in match:
                 current_commit_issues.append(
-                    f"Commit locked to a repo other than {self.repo_name}: {commit_message.partition('\n')[0]}"
+                    f"Commit locked to a repo other than {self.repo_name}: {firstline}"
                 )
 
         if INVALID_REVIEW_FLAG_RE.search(firstline):
             current_commit_issues.append(
-                f"Revision contains 'r?' in the commit message. Please use 'r=' instead: {commit_message.partition('\n')[0]}"
+                f"Revision contains 'r?' in the commit message. Please use 'r=' instead: {firstline}"
             )
 
         if firstline.lower().startswith("wip:"):
             current_commit_issues.append(
-                f"Revision seems to be marked as WIP: {commit_message.partition('\n')[0]}"
+                f"Revision seems to be marked as WIP: {firstline}"
             )
 
         if re.search(RE_DIFF, commit_message):
             current_commit_issues.append(
-                f"Suspected diff found in commit message. Please indent the diff if this is on purpose: {commit_message.partition('\n')[0]}"
+                f"Suspected diff found in commit message. Please indent the diff if this is on purpose: {firstline}"
             )
 
         if current_commit_issues:
@@ -561,11 +563,11 @@ class CommitMessagesCheck(PatchCollectionCheck):
             # Purposely ambiguous: it's ok to say "backed out rev N" or
             # "reverted to rev N-1"
             current_commit_issues.append(
-                f"Backout revision needs a bug number or a rev id: {commit_message.partition('\n')[0]}"
+                f"Backout revision needs a bug number or a rev id: {firstline}"
             )
 
         current_commit_issues.append(
-            f"Revision needs 'Bug N' or 'No bug' in the commit message: {commit_message.partition('\n')[0]}"
+            f"Revision needs 'Bug N' or 'No bug' in the commit message: {firstline}"
         )
 
         self.commit_message_issues.extend(current_commit_issues)
@@ -575,7 +577,7 @@ class CommitMessagesCheck(PatchCollectionCheck):
         if self.ignore_bad_commit_message:
             return []
 
-        return self.commit_message_issues
+        return list(self.commit_message_issues)
 
 
 @dataclass
--- a/src/lando/utils/landing_checks.py
+++ b/src/lando/utils/landing_checks.py
@@ -67,7 +67,7 @@ class PatchCheck(Check, ABC):
         """Pass the next `rs_parsepatch` diff `dict` into the check."""
 
     @abstractmethod
-    def result(self) -> str | None:
+    def result(self) -> list[str]:
         """Calculate and return the result of the check."""
 
 
@@ -139,13 +139,13 @@ class PreventPathCheckMixin(ABC):
             ):
                 self.disallowed_changes.append(filename)
 
-    def result(self) -> str | None:
+    def result(self) -> list[str]:
         """Calculate and return the result of the check."""
         if not self.disallowed_changes:
             # Return early if no disallowed changes were found.
-            return
+            return []
 
-        return self.build_error_message()
+        return [self.build_error_message()]
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


src/lando/utils/landing_checks.py
```diff
@@ -495,53 +506,50 @@ def next_diff(self, patch_helper: PatchHelper):
         if is_backout(firstline):
             backouts = parse_backouts(firstline, strict=True)
             if not backouts or not backouts[0]:
-                self.commit_message_issues.append(
+                current_commit_issues.append(
                     "Revision is a backout but commit message "
-                    f"does not indicate backed out revisions: {commit_message}"
+                    + f"does not indicate backed out revisions: {commit_message.partition('\n')[0]}"
```
cgsheeh: `commit_message.partition('\n')[0]` is the same as the pre-existing `firstline` variable, we should use that instead to avoid the duplicated logic.

src/lando/utils/landing_checks.py
```diff
@@ -495,53 +506,50 @@ def next_diff(self, patch_helper: PatchHelper):
...
-            return
 
         if re.search(RE_DIFF, commit_message):
-            self.commit_message_issues.append(
-                f"Suspected diff found in commit message. Please indent the diff if this is on purpose: {commit_message}"
+            current_commit_issues.append(
+                f"Suspected diff found in commit message. Please indent the diff if this is on purpose: {commit_message.partition('\n')[0]}"
             )
+
+        if current_commit_issues:
```
cgsheeh: Should we continue looking for errors here? The bug and PR title indicate we are aggregating every issue together, but it seems like we stop at this point. A comment indicating why would be helpful.

src/lando/utils/landing_checks.py
```diff
@@ -481,6 +490,8 @@ def next_diff(self, patch_helper: PatchHelper):
         commit_message = patch_helper.get_commit_description()
         author, _email = patch_helper.parse_author_information()
 
+        current_commit_issues = []
```
cgsheeh: We create this list, but still do `self.commit_message_issues.append(...` directly in a few places below. I'd suggest we have a single place to route error messages, or change this variable name so it's clearer when we use each.

src/lando/utils/landing_checks.py
```diff
@@ -552,19 +560,22 @@ def next_diff(self, patch_helper: PatchHelper):
...
+        self.commit_message_issues.extend(current_commit_issues)
+
+    def result(self) -> list[str]:
         """Calculate and return the result of the check."""
-        if not self.ignore_bad_commit_message and self.commit_message_issues:
-            return ", ".join(self.commit_message_issues)
+        if self.ignore_bad_commit_message:
+            return []
+
+        return self.commit_message_issues
```
cgsheeh: Nit: perhaps we should return a fresh `list(self.commit_message_issues)` here to avoid unexpected behaviour from passing a reference.

src/lando/utils/tests/test_landing_checks.py
```diff
@@ -358,16 +358,15 @@ def test_check_commit_message_repolocked(
...
 
-    assert (
-        assessor.run_patch_collection_checks(
-            patch_collection_checks=[CommitMessagesCheck], patch_checks=[]
-        )
-        == expected
-    ), error_message
+    if return_string:
+        expected = return_string + commit_message.partition("\n")[0]
+        assert expected in check_issues, error_message
```
cgsheeh: Can we assert the exact expected check issues? With this setup we could end up having every issue trigger by accident (ie `check_issues` has every possible error message), and the test wouldn't pick it up since it's just checking that one expected message is in the list.

src/lando/utils/tests/test_landing_checks.py
```diff
@@ -847,12 +858,12 @@ def test_landing_checks_run():
...
+                # PreventNSPRNSSCheck and PreventNSPRCheck will get triggered.
                 GIT_PATCH_FILENAME_TEMPLATE.format(filename="nsprpub/testfile.txt"),
             )
         ),
     ]
 
     names_run = landing_checks.run([chk.name() for chk in ALL_CHECKS], patch_helpers)
 
-    assert len(names_run) == 4
+    assert len(names_run) == 5
```
cgsheeh: Can we check which messages were returned instead of just asserting the correct count?

cgsheeh (dismissed):

src/lando/utils/landing_checks.py
shtrom: D'oh!

src/lando/utils/landing_checks.py
shtrom: Made the successful `return` condition more logical.

src/lando/utils/landing_checks.py
shtrom: We (now) have two places where it happens:
1. early sanity check if commit is empty
2. end of the method, which should be the ideal case

The former is an early return, as we can't do any other check there, so I think it's OK. I've now moved the declaration of `current_commit_issues` below the return from that check, so we have only a single path that populates that list, and ultimately extends the global one with it.

src/lando/utils/tests/test_landing_checks.py
shtrom: ... which it did 😅

src/lando/utils/landing_checks.py
shtrom: Hum, I'm hitting a convoluted logic issue. Up to now, if all checks up to that point passed, and the commit was in an acceptable format, we'd return there.

If I flatten this, I have those two checks, which look very similar:
1. before the early return https://github.com/mozilla-conduit/lando/blob/13b7aac1d6ea25d957fc1da65cb06471a65dd703/src/lando/utils/landing_checks.py#L507-L512
2. after the early return https://github.com/mozilla-conduit/lando/blob/13b7aac1d6ea25d957fc1da65cb06471a65dd703/src/lando/utils/landing_checks.py#L560-L565

They look like dupes to me, but perhaps I'm missing a finer point?

shtrom: image

shtrom: I'm tempted to just remove the second check.

Doing so then triggers the No bug check. So maybe that test backout message is not good enough? But the previous logic with the early return bypassed that.

WDYT?

shtrom: (For now, I left the early return).

src/lando/utils/landing_checks.py
```diff
@@ -485,63 +494,64 @@ def next_diff(self, patch_helper: PatchHelper):
             self.commit_message_issues.append("Revision has an empty commit message.")
             return
 
-        firstline = commit_message.splitlines()[0]
+        firstline = commit_message.partition("\n")[0]
```
cgsheeh: `splitlines()[0]` also dropped a trailing `\r`, but `partition` keeps it, so a CRLF commit message now renders as ``Revision seems to be marked as WIP: WIP: my title\r``. Can we keep `splitlines()` here, or `rstrip("\r")`?

src/lando/utils/landing_checks.py
```diff
@@ -485,63 +494,64 @@ def next_diff(self, patch_helper: PatchHelper):
...
-                        f"Push contains commits intended to be locked to {repo} but the repo name is badly formatted. '/' is not allowed: {commit_message}"
+                    current_commit_issues.append(
+                        f"Push contains commits intended to be locked to {repo} but the repo name is badly formatted. '/' is not allowed: {firstline}"
                     )
-                    return
-            if self.repo_name not in match:
-                self.commit_message_issues.append(
-                    f"Commit locked to a repo other than {self.repo_name}: {commit_message}"
+                    malformed = True
+            if not malformed and self.repo_name not in match:
```
cgsheeh: Nit: you could do `malformed = any("/" in repo for repo in match)` and avoid the flag variable.

src/lando/utils/landing_checks.py
```diff
@@ -330,11 +332,14 @@ def next_diff(self, diff: dict):
...
-            return "Patch attempts to modify repository metadata: " + wrap_filenames(
-                self.hg_files
-            )
+    def result(self) -> list[str]:
+        if not self.hg_files:
+            return []
+
+        return [
+            "Patch attempts to modify repository metadata: "
+            + f"{wrap_filenames(self.hg_files)}."
```
cgsheeh, line 342: Nit: the `+` is unnecessary as adjacent string literals will already concatenate.

src/lando/utils/landing_checks.py
```diff
@@ -432,8 +441,8 @@ def run_diff_checks(self, patch_checks: list[type[PatchCheck]]) -> list[str]:
 
         # Collect the results from each check.
         for check in checks:
-            if issue := check.result():
-                issues.append(issue)
+            if check_issues := check.result():
+                issues.extend(check_issues)
```
cgsheeh: I think we can remove the walrus operator here, since `issues.extend([])` is just a no-op. Same thing for the push-wide checks.

cgsheeh (approved):

src/lando/utils/landing_checks.py
shtrom: Yes, however the LSP flags it when the + is missing. I agree with it in the implicit concatenation may happen due to another missing separator, e.g., a `,`. Adding an explicit `+` when a concatenation is intended seems clearer to me.

src/lando/utils/landing_checks.py
shtrom: I think the logic is the opposite though: we only `extend` if `check_issues` is not empty, at which point we need to have the value handy.

src/lando/utils/landing_checks.py
shtrom: Errr. Yep, ok. Splitlines it is. Good catch.

src/lando/utils/landing_checks.py
shtrom: Ah! Nice!

shtrom: I had to make it slightly different so I can retain the repo names for the error message.

src/lando/utils/landing_checks.py
```diff
@@ -491,57 +499,56 @@ def next_diff(self, patch_helper: PatchHelper):
...
+                current_commit_issues.append(
                     "Revision is a backout but commit message "
-                    f"does not indicate backed out revisions: {commit_message}"
+                    + f"does not indicate backed out revisions: {firstline}"
                 )
-                return
 
         # Avoid checks for the merge automation users.
         if author in {"ffxbld", "seabld", "tbirdbld", "cltbld"}:
             return
```
cgsheeh, line 517: This early return now swallows the backout issue collected just above, since it only lands in `current_commit_issues`. On `main` a malformed backout from `ffxbld` was reported; now it returns `[]`. Should the backout check move below this return, or should we flush `current_commit_issues` before bailing?

src/lando/utils/landing_checks.py
```diff
@@ -676,44 +693,50 @@ def next_diff(self, patch_helper: PatchHelper):
 
         self.bug_ids |= set(parse_bugs(commit_message))
 
-    def result(self) -> str | None:
+    def result(self) -> list[str]:
```
cgsheeh, line 688: `UpliftWorker.check_uplift_bug_references` is still annotated `-> str | None` and does `raise ValueError(result)`, so the uplift failure message will now render as a Python list repr. Can we update that caller?

src/lando/utils/landing_checks.py
shtrom: Ah! https://searchfox.org/version-control-tools/source/hghooks/mozhghooks/commit-message.py#123-128

I think that's the origin of that check, i.e, it's a malformed backout, because there's no revision, so we need to flag it as such. In any other case, we complain about the missing bug no instead.

src/lando/utils/landing_checks.py
shtrom: Flushing 👍🏼

PR #1449, opened 2026-08-07
main.admin: add profile model (bug 2042706)

zzzeid:
- add Profile model to admin
- override UserAdmin to show inline Profile
- add human-readable `__str__` to Profile

---

Lando: [link](https://lando.moz.tools/pulls/lando/1449/)
Bugzilla: [bug 2042706](https://bugzilla.mozilla.org/show_bug.cgi?id=2042706)

### DIFF
--- a/src/lando/main/admin.py
+++ b/src/lando/main/admin.py
@@ -3,6 +3,8 @@ from typing import Callable, Self, override
 
 from django import forms
 from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.models import User
 from django.db.models import Field as DbField
 from django.forms import CheckboxSelectMultiple, MultipleChoiceField
 from django.forms import Field as FormField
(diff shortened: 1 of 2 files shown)


src/lando/main/admin.py
```diff
@@ -598,6 +601,26 @@ def job_count(self, instance: UpliftSubmission) -> int:
...
+        "phabricator_api_key_is_set",
+    )
+
+    @staticmethod
+    def phabricator_api_key_is_set(instance: Profile) -> bool:
+        """Return True if phabricator_api_key is set."""
+        return bool(instance.phabricator_api_key)
+
+
+class UserAdmin(UserAdmin):
```
shtrom: nit: Is the shadowing on purpose here?

zzzeid: It was intentional, but fair point, I suppose there's no need to necessarily do it this way.

